PR #44, opened 2026-02-23
chore: update more references to cdktf

gabegorelick:
All of these are in docs or comments, so should be safe to update.

### Related issue

Fixes # 

### Description

In plain English, describe your approach to addressing the issue linked above. For example, if you made a particular design decision, let us know why you chose this path instead of another solution.

### Checklist

### DIFF
--- a/test/run-against-dist
+++ b/test/run-against-dist
@@ -57,7 +57,7 @@ npm init -y > /dev/null
 npm install --global-style cdktn-cli
 
 # stop the local registry now that we installed the CLI
-# Some tests install cdktf in an older version from NPM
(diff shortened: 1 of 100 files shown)


test/run-against-dist
```diff
@@ -57,7 +57,7 @@ npm init -y > /dev/null
 npm install --global-style cdktn-cli
 
 # stop the local registry now that we installed the CLI
-# Some tests install cdktf in an older version from NPM
```
jsteinich: Technically this is still true, though we'll likely update the test in the future to use `cdktn`

jsteinich (approved):

PR #60, opened 2026-03-18
fix(provider-generator): handle inline comments with brackets in variable type parsing

josh-hook:
### Related issue

Fixes https://github.com/open-constructs/cdk-terrain/issues/39

### Description

This PR fixed `cdktn get` command failing to generate code from downstream Terraform that contains a variable with type including a closing bracket in the comments.

Example offending Terraform:
```
variable "iam_policy_statements" {
  description = "A list of IAM policy statements"
  type = list(object({  # TODO - change to map(object({...})) in next major version
    sid       = optional(string)
    actions   = optional(list(string))
    effect    = optional(string)
    resources = optional(list(string))
  }))
  default = null
}
```
(currently present in the [Karpenter Terraform module](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/v21.15.1/modules/karpenter/variables.tf#L121))

The code generation fails due to the `transformVariables` CDKTN function that's parsing the variable type definition. Since the regex does not match newlines, it ends up matching to the commented-out `}` character, causing the resulting type to be parsed as `list(object({ # TODO - change to map(object({...`, which then later can't be parsed to an expression.

Previous approach: match everything on the first line which is wrapped in `${...}`, everything else is an `any`.
Proposed fix: If it's a single-line type then grab everything between the starting `${` and ending `}`, else it's an `any`.

This PR also fixes the `hcl2json`/`hcl-tools` prebuild.sh script as in later versions of Go the wasm directory has changed from `misc` to `lib`.

### Checklist

- [x] I have updated the PR title to match [CDKTN's style guide](https://github.com/open-constructs/cdk-terrain/blob/main/CONTRIBUTING.md#pull-requests-1)
- [x] I have run the linter on my code locally
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the [documentation](https://github.com/open-constructs/cdk-terrain/tree/main/website) if applicable
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works if applicable
- [x] New and existing unit tests pass locally with my changes

PR #61, opened 2026-03-20
fix(provider-generator): handle inline comments with brackets in variable type parsing

so0k:
## Summary
- Includes the fix from #60 by @josh-hook  to handle inline comments with brackets in Terraform variable type constraints during schema parsing
- Adds the missing test snapshot that was not committed in the original PR

Closes #60

## Test plan
- [x] Snapshot generated locally with `yarn test:update` in `packages/@cdktn/provider-schema`
- [x] CI unit tests pass with the new snapshot


sheplu (changes requested):
only a small comment the other part is fine

sheplu (approved):
LGTM

PR #62, opened 2026-03-20
feat(cli): replace HashiCorp telemetry with Sentry analytics

### LINKED ISSUE #48: cdktn-cli: telemetry uses HashiCorp's endpoint
### Expected Behavior

The CLI should use an endpoint not owned by HashiCorp, or not rely on any external calls for telemetry or newest version checks.

### Actual Behavior

This base URL is configured:
https://github.com/open-constructs/cdk-terrain/blob/a5ff6fa3e760979733270fc68011d8e81065f5f6/packages/%40cdktn/commons/src/checkpoint.ts#L13

I haven't checked how it behaves in runtime, though.

### Steps to Reproduce

I haven't tried it in runtime, but I presume a reproducer is a simple as running any CLI command.

### Versions

cdktn-cli: 0.22.0

### Providers

_No response_

### Gist

_No response_

### Possible Solutions

_No response_

### Workarounds

_No response_

### Anything Else?

_No response_

### References

_No response_

### Help Wanted

### Community Note

so0k:
## Summary

Removes all CLI telemetry to HashiCorp's `checkpoint-api.hashicorp.com` and routes usage analytics through the project's own Sentry instead. Resolves #48.

Because Sentry **sunset the 7.x custom-metrics product** the original spec targeted (server-side, 2024-10-07 — a silent no-op on `7.120.4`), this upgrades `@sentry/node` `7.120.4 → ^10.56` and emits analytics via the v10 metrics API. Sentry error/crash reporting is preserved throughout, and a privacy improvement falls out of the migration (machine hostname is now scrubbed from crash reports too).

## 👀 For reviewers — start with `spec.md`

**Please review [`specledger/002-remove-hashicorp-telemetry/spec.md`](https://github.com/open-constructs/cdktn-planning/blob/f4fbc15510ab83d1f2378a04bcb1f9ef7908c89a/specledger/002-remove-hashicorp-telemetry/spec.md) first.** Every other artifact under that folder (`plan.md`, `research.md`, `data-model.md`, `contracts/`, `quickstart.md`) is a **supporting derivation** of the spec and deliberately repeats its concepts — they exist to trace requirements to design and tests, not to introduce new decisions. If you align on the spec's user stories (US1–US5), functional requirements (FR-001…FR-018) and consent model, the rest follows. Skim the supporting artifacts only where you want the "how" behind a specific requirement; the implementation and tests map back to FR/US IDs.

## What changed (4 reviewable phases + remediation)

- **Phase A — SDK upgrade (behavior-unchanged):** `@sentry/node` → `^10.56` (resolves 10.56.0) atomically across `commons`/`cli-core`/`cdktn-cli` (single shared instance); `Sentry.configureScope` → `getCurrentScope()`; `Sentry.init` rebuilt — drop removed `autoSessionTracking`, pin `tracesSampleRate: 0` (metrics deliver independently of trace sampling, confirmed against 10.57) and `serverName: "cdktn-cli"` (scrubs the machine hostname from **error events** — a pre-existing leak — **and** future metrics). [US3, FR-003/FR-018]
- **Phase B — remove HashiCorp transport:** delete `checkpoint.ts` (`ReportRequest`/`ReportParams`/`post`/`BASE_URL`); relocate the preserved `getUserId`/`getProjectId` to `identity.ts`; drop the fire-and-forget `report()` from the `Errors` factories. [US1, US4, FR-001/FR-011/FR-012]
- **Phase C — usage analytics via v10 metrics + consent:** `sendTelemetry` → `cli.command.invoked` / `cli.command.error` / `cli.synth.duration`; new typed `sendUsageTelemetry` flag (`ConfigBase` + validated `CdktfConfig` getter); per-flag interactive consent prompts (TTY, non-CI) persisted to `cdktf.json`; `CHECKPOINT_DISABLE` retained as the usage-telemetry override; bounded `Sentry.flush()` on normal exit so the short-lived CLI actually ships buffered metrics. [US2, US5, FR-002/FR-005…FR-017]
- **Phase D — verification tooling:** `tools/sentry-sink.mjs` + `tools/validate-sentry-e2e.sh` (local envelope sink E2E) and a gated jest bundle-E2E; no-egress oracles (nock runtime canary + static source/bundle scan). [SC-001/SC-007]
- **Review remediation:** addressed the checkpoint review findings (typed `ci-info` import for lint, direct getter tests, the nock runtime no-egress test, dropped dead synth payload), and reconciled the spec artifacts (the spec was re-baselined off the sunset 7.x transport to the v10 decision).

## ⚠️ Migration notes (user-facing)

- **cdktn no longer sends any data to HashiCorp.** Usage analytics now go to the project's own Sentry, only when enabled.
- **New `sendUsageTelemetry` flag** in `cdktf.json`, independent of `sendCrashReports`. Collected: command, language, CI name, synth duration — **no hostname, no stack contents**.
- **One-time interactive prompt on upgrade.** In an interactive terminal (TTY, not CI), a project whose `cdktf.json` lacks `sendUsageTelemetry` is asked once and the choice persisted. You may **also** be prompted for crash reporting if `sendCrashReports` is missing — that prompt existed before but never fired due to a bug (a missing key was silently treated as "no"); it now works as originally intended.
- **Non-interactive (CI/piped) unchanged:** no prompts; usage telemetry stays on by default (now to Sentry instead of HashiCorp), crash reporting stays off by default.
- **`CHECKPOINT_DISABLE` still works** (disables usage telemetry everywhere, never affects crash reporting), and telemetry is fully inert without a Sentry DSN.

## Test plan

- [x] Unit tests green across `@cdktn/commons`, `@cdktn/cli-core`, `cdktn-cli`
- [x] Delivery oracle: real v10 client + capturing transport asserts a `trace_metric` envelope (`cli.command.invoked`) + `flush()===true`
- [x] Consent-gating matrix (per-flag prompt, FR-016/FR-017 precedence, `CHECKPOINT_DISABLE`, no-project, CI)
- [x] No-egress: nock runtime canary (net disabled) + static source/bundle scan — zero `checkpoint-api.hashicorp.com` references
- [x] Bundle E2E (`tools/validate-sentry-e2e.sh`): convert (success-path flush) + failing synth (error metric) delivered to a local sink
- [x] Maintainer (one-time): enable the Sentry "Logs & Metrics" product (org-side; SDK ≥10 gate); release sourcemap symbolication sanity check (sentry-cli stays 2.58.4)

Resolves #48


sakul-learning:
### Automated Review - `002-remove-hashicorp-telemetry`

Verdict:  No Issues found

### Validations Ran - Summary

**1. UUID Persistence — 7/7 ✓**  
`getId()` generates a random UUIDv4, persists to `~/.cdktf/config.json`, and is stable across runs:
- ✓ First call generates + persists a v4 UUID
- ✓ Second call returns same UUID (cross-run constancy)
- ✓ `forceCreation=true` does not clobber stored UUID
- ✓ Deleting `config.json` triggers a new UUID (no leakage)
- ✓ `getUserId()` returns stable ID across 3 simulated "runs"
- ✓ 20 regenerated UUIDs are all unique v4 — no collisions
- ✓ Config comment explains UUID is random, not PII, and deletable

**2. Consent Logic Simulation — 58/58 ✓**  
All 15 consent flag combinations tested against a pure-Node.js replica of `error-reporting.ts`:
- ✓ Crash & usage consent are **independent** — setting one never affects the other's prompt
- ✓ `CHECKPOINT_DISABLE` only gates usage telemetry, never crash reporting
- ✓ `beforeSend` drops errors when crash consent is absent
- ✓ `beforeSend` drops "Usage Error" exceptions even with crash consent
- ✓ Non-interactive environments (no TTY, CI) skip prompts, default usage ON, crash OFF
- ✓ Outside a project (no `cdktf.json`) → no prompts, usage default-ON
- ✓ No `SENTRY_DSN` → Sentry not initialized regardless of consent flags

**3. Interactive PTY Simulation — 6/6 ✓**  
Real `@inquirer/prompts confirm()` prompts driven over a PTY:

| # | Scenario | Result |
|---|----------|:---:|
| 1 | Both UNSET, answer Y Y → both prompts fire | ✓ |
| 2 | Both UNSET, answer N N → both prompts fire, both declined | ✓ |
| 3 | Crash TRUE, usage UNSET → crash prompt SKIPPED, usage fires | ✓ |
| 4 | Crash UNSET, usage FALSE → only crash prompts | ✓ |
| 5 | CHECKPOINT_DISABLE → crash prompts, usage SKIPPED | ✓ |
| 6 | Both EXPLICITLY SET → NO prompts at all | ✓ |

### Analysis
No PII is collected. The userId is cryptographically random (UUIDv4). The consent prompts fire independently for crash reporting and usage telemetry, and are correctly gate-kept by the existing consent flags so users are never re-prompted after making a choice. Defaults in non-interactive environments are conservative (crash OFF). No `SENTRY_DSN` means no outbound traffic regardless of consent state.

so0k:
**CI triage for the failed checks on 30b0c1f** (re-triggered by pushing 6ee638bf0):

- **Quality / prettier — real failure, fixed.** `CLAUDE.md`'s generated "Active Technologies" block contained a bullet garbled by a mid-`**bold**` line wrap, which `prettier --check` rejected. Repaired in 6ee638bf0 (also refreshed the stale testing note in that block).
- **Examples java-aws + typescript-aws-multiple-stacks @ tf1.6.5 — infra flake.** Job annotations show `Docker pull failed with exit code 1` with multiple backoff retries (registry rate-limiting on the runners); unrelated to this branch. The new head re-runs them.

so0k:
**Second CI triage (head 6ee638bf0):** Quality/prettier now passes (the real failure, fixed in this head). The remaining 3 failures are all the same Docker Hub flake — `docker pull terraconstructs/jsii-terraform@sha256:…` timing out at container init (`Client.Timeout exceeded while awaiting headers`, 3 retries) before any project code ran: `Build / build-and-package`, `@examples/java-gradle-shared-module @ tf1.6.5` (note: a *different* example than last round — the two previous flaky ones now pass), and the `PR - Result` aggregate. Re-ran failed jobs.

sakul-learning:
### 🔬 Local E2E CDKTN_SENTRY validation — PASS

Ran `tools/validate-sentry-e2e.sh` against the real esbuild bundle:

| Trigger | Metric | Status |
|---------|--------|:---:|
| `cdktn convert` (success path) | `cli.command.invoked` | ✓ |
| `cdktn synth --app "node -e process.exit(1)"` (error path) | `cli.command.error` | ✓ |
| Bundle grep for `checkpoint-api.hashicorp.com` | 0 hits | ✓ |

**Details:**
- Bundle built with `SENTRY_DSN=http://cdktn@localhost:9999/1` baked in
- Local Sentry sink recorded both `trace_metric` envelopes
- SDK sent `sentry.javascript.node/10.56.0` client info
- Both envelopes gzip-none, raw POST to the local sink
- HashiCorp checkpoint references fully removed from the bundle

so0k:
**Please anchor the review on [open-constructs/cdktn-planning - spec.md](https://github.com/open-constructs/cdktn-planning/blob/f4fbc15510ab83d1f2378a04bcb1f9ef7908c89a/specledger/002-remove-hashicorp-telemetry/spec.md).** Everything else under `specledger/002-remove-hashicorp-telemetry/` — `plan.md`, `research.md`, `data-model.md`, `contracts/`, `quickstart.md` — is a **supporting artifact derived from the spec** and intentionally restates the same user stories / requirements from a different angle (design, data, contracts, test journeys). They don't introduce independent decisions. **If you're aligned on the spec (US1–US5, FR-001…FR-018, the consent model), the rest follows**; dip into the supporting docs only for the "how" behind a specific requirement. The code and tests trace back to FR/US IDs.

~CI status: the red checks are a Docker Hub rate-limit flake on the example/build job containers (`terraconstructs/jsii-terraform` pull timing out at container init, a different leaf each run) — not this PR; all code checks (unit, prettier, lint, CodeQL) pass. The fix is #261 (pull the CI image from GHCR), **blocked on org admin making the `jsii-terraform` GHCR package public**; tracked under #206.~

sakul-learning:
### Minor observations

**1. Dual config readers — intentional, worth keeping in sync**

There are two surfaces that read `sendUsageTelemetry` from `cdktf.json`:

| Reader | Location | Behavior |
|--------|----------|----------|
| `getUsageTelemetryConsent()` | `commons/src/telemetry.ts` | **Forgiving**: raw `JSON.parse`, silent `undefined` on errors, handles `true`/`false` booleans and string `"true"` |
| `CdktfConfig.sendUsageTelemetry` | `cli-core/src/lib/cdktf-config.ts` | **Typed/validated**: throws `Errors.External` for invalid values, handles booleans and both `"true"`/`"false"` strings |

The forgiving reader exists because `@cdktn/commons` can't import from `@cdktn/cli-core` (circular dependency), and consent gating must work before/without a valid project. The typed getter remains the validated surface for project-level consumers.

Both are tested (`commons/src/telemetry.test.ts`, `cli-core/src/test/cdktf-config.test.ts`) which catches drift. The code comment at `telemetry.ts:18-24` documents the rationale.

**2. Upgrade path: crash set, usage unset → prompts once (verified)**

At `error-reporting.ts:84-108`: when `sendCrashReports: true` is already in `cdktf.json` but `sendUsageTelemetry` is absent, `shouldReportCrash()` returns `true` (not `undefined`) so the crash prompt is skipped, while `getUsageTelemetryConsent()` returns `undefined` so the usage prompt fires exactly once and the decision is persisted. ✓

so0k:
Related PR might be https://github.com/open-constructs/cdk-terrain/pull/226

packages/cdktn-cli/src/test/sentry-bundle-e2e.test.ts
```diff
@@ -0,0 +1,96 @@
...
+// trace_metric envelope reaches a local Sentry sink — proving the DSN
+// bake-in, consent gating and the bounded success-path flush end to end.
+//
+// Requires the bundle to be PRE-BUILT with the matching local-sink DSN:
+//   (cd packages/cdktn-cli && SENTRY_DSN=http://cdktn@localhost:9999/1 \
+//      node build-config/build.js)
+// then run with CDKTN_SENTRY_E2E=1 (port override: CDKTN_SENTRY_E2E_PORT).
+// tools/validate-sentry-e2e.sh wraps the same flow as a standalone script.
+
+const gated = process.env.CDKTN_SENTRY_E2E ? describe : describe.skip;
```
jsteinich, line 20: This generally means that we'll never run this test again.
I think we should either come up with a way that the test is actually ran on a regular basis, or just not add it at all (and the supporting files).

packages/cdktn-cli/src/test/no-hashicorp-egress.test.ts
jsteinich, line 1: This test doesn't really add any value.

packages/@cdktn/commons/src/telemetry.ts
```diff
@@ -0,0 +1,133 @@
...
+  return getUsageTelemetryConsent(projectPath) !== false;
+}
+
+/**
+ * Bounded flush of buffered telemetry. Required before any explicit
+ * `process.exit()` on a path that may have emitted metrics — the legacy
+ * checkpoint transport awaited its POST inline, but Sentry buffers
+ * asynchronously and a hard exit drops the buffer.
+ */
+export async function flushTelemetry(timeoutMs = 4000): Promise<void> {
```
jsteinich, line 84: I noticed at least 2 other spots doing something similar to this. Can we update them to use this or is this actually used?

packages/@cdktn/cli-core/src/test/no-hashicorp-runtime-egress.test.ts
jsteinich, line 1: This test doesn't seem to add value.

packages/@cdktn/cli-core/src/lib/synth-stack.ts
```diff
@@ -271,22 +281,18 @@ Command output on stdout:
...
     totalTime: number,
-    stacks: SynthesizedStack[],
+    _stacks: SynthesizedStack[],
     synthOrigin?: SynthOrigin,
   ): Promise<void> {
     const config = readConfigSync();
 
+    // Only command/language/ci and the duration are collected (FR-014);
+    // the legacy stackMetadata/requiredProviders payload is no longer
+    // sent anywhere, so it is not computed.
```
jsteinich, line 291: I think we lose out on a bunch of useful information without these.
Required providers gives insight into which providers should have pre-built support.
The stack metadata includes things like when overrides are used (built-in functionality is lacking) and other pieces.

packages/@cdktn/commons/src/errors.ts
```diff
@@ -1,34 +1,18 @@
...
 }
 
-function reportPrefixedError(type: ErrorType, command: string) {
+function reportPrefixedError(type: ErrorType, _command: string) {
   return (
     message: string,
     originalError: Error = new Error(),
     context?: Record<string, any>,
   ) => {
-    report(command, { ...context, message, type });
```
jsteinich, line 31: It seems like removing this means we are just dropping the reporting. I'm not sure why we'd want to get rid of it.

packages/@cdktn/cli-core/src/lib/error-reporting.ts
```diff
@@ -54,38 +78,75 @@ function isPromise(p: any): p is Promise<any> {
...
-  if (shouldReport === undefined && runConsentPrompt) {
-    // But only if it's a user
-    if (ci) {
-      return;
-    }
+  // Prompting requires a real user at a terminal (TTY and not CI) and a
+  // cdktf.json to persist the decision into; otherwise fall through to
+  // the per-flag non-interactive defaults below.
+  const canPrompt =
+    Boolean(process.stdout.isTTY) &&
```
jsteinich, line 91: This isn't accurate in every terminal, but it matches the behavior used elsewhere in the codebase and a false negative isn't that consequential for this use.

jsteinich (changes requested):

packages/@cdktn/cli-core/src/lib/synth-stack.ts
so0k: agree, will look into Sentry to find a way to track stuff- also the new targetVersion info would be really useful to know

so0k:
will rebase this and address the request to keep usage information (we know should be able to track actual binary (tofu vs terraform) and its version as well as the project `targetVersion`). This should allow us to prioritize upcoming features better (as well as get rid of Hashicorp checkpoint endpoint).

so0k:
Superseded by a six-slice stack, and closing this one in its favour. Everything here lands in the stack; nothing is dropped except the two egress tests and the gated jest bundle E2E, which the review asked for and which the stack never reintroduces.

Review order is S1 → S6, each based on the previous slice:

1. #409 `chore(deps): upgrade @sentry/node to 10.x with unchanged reporting behaviour`
2. #410 `fix(cli): stop the top-level error handler racing yargs` (fixes #360, fixes #361)
3. #411 `feat(cli): replace HashiCorp checkpoint telemetry with Sentry usage metrics and consent` (resolves #48)
4. #412 `feat(cli): report the installed binary, target versions and platform in usage metrics`
5. #413 `feat(cli): report per-stack, per-provider and per-command usage metrics`
6. #414 `chore(gha): run the telemetry delivery e2e on every build`

@jsteinich, your threads map to slices as follows. Each is answered in that slice's body, and the code that answers it is in that slice's diff:

| Thread | Gist | Answered in |
|---|---|---|
| [r3408167759](https://github.com/open-constructs/cdk-terrain/pull/62#discussion_r3408167759) | "we'll never run this test again" (gated jest bundle E2E) | #414: the wrapper is gone; `tools/validate-sentry-e2e.sh` is a build-workflow step running on every build |
| [r3408168497](https://github.com/open-constructs/cdk-terrain/pull/62#discussion_r3408168497) | "doesn't really add any value" (static no-egress test) | #411: never introduced; the bundle grep in #414 carries the claim instead |
| [r3408179555](https://github.com/open-constructs/cdk-terrain/pull/62#discussion_r3408179555) | "doesn't seem to add value" (nock runtime no-egress test) | #411: never introduced, and `nock` leaves `cli-core`, where it was the only user |
| [r3408173331](https://github.com/open-constructs/cdk-terrain/pull/62#discussion_r3408173331) | "at least 2 other spots doing something similar" (duplicated flush) | #411: one bounded `flushTelemetry()` in commons; nothing else calls `Sentry.flush` or `Sentry.close` |
| [r3408186638](https://github.com/open-constructs/cdk-terrain/pull/62#discussion_r3408186638) | dropped `Errors` reporting | #411: back as `cli.error{type, command}`, a count with no message and no context, behind the usage gate |
| [r3408182518](https://github.com/open-constructs/cdk-terrain/pull/62#discussion_r3408182518) | dropped stack payload loses useful information, plus so0k's target-versions follow-up | #413 for the per-stack metrics; #412 for the declared target versions on every command metric |
| [r3408200466](https://github.com/open-constructs/cdk-terrain/pull/62#discussion_r3408200466) | `isTTY` is not accurate in every terminal | #411: taken as informational, the check stays as it is |

If it helps to see the result whole rather than in six parts, the branch `002-remove-hashicorp-telemetry-monolith` is the unsplit tree the stack was cut from. The top of the stack is byte-identical to it, so it is usable for cross-validation: read it for the shape of the finished thing, and review the slices for the change itself.

The planning ledger for this work is https://github.com/open-constructs/cdktn-planning/pull/6.

PR #93, opened 2026-04-09
fix(lib): fail on constructs outside of stacks (under feature flag)

### LINKED ISSUE #57: cdktn: synthesis succeeds if resource is added with app's scope
### Expected Behavior

For resources attached to `scope=App()`, I'd expect CDKTF to either deploy it within some root/unnamed stack, or fail fast since this resource isn't created under any stack.

### Actual Behavior

CDKTF succeeds during synthesis, and doesn't deploy the resource anywhere.

### Steps to Reproduce

See this piece of code:

```python
from cdktn import App
from cdktn_provider_google.project_iam_member import ProjectIamMember

ProjectIamMember(
    scope=app,
    id_="some-iam-member",
    member=...,
    project=...
    role=...
)
```

### Versions

cdktn: 0.22.0
cdktn-cli: 0.22.0

### Providers

cdktn-provider-google==17.0.0

### Gist

_No response_

### Possible Solutions

I don't know what's the proper solution yet, given that ideally we'd fail fast for all supported languages.

### Workarounds

In my case, I use Python. I added extra validation as a subclass of `App`, like:

```python
class ValidatedApp(App):
       def synth(self) -> None:
        orphans = []
        for node in self.node.find_all():
            if isinstance(node, (App, TerraformStack)):
                continue
            try:
                find_terraform_stack(node)
            except ValueError:
                orphans.append(Node.of(node).path)
        if orphans:
            bullet_list = "\n".join(f"  - {path}" for path in orphans)
            raise ValueError(
                "Found constructs outside of a TerraformStack "
                "(did you pass scope=app instead of scope=stack?):\n" + bullet_list
            )
        super().synth()
```

### Anything Else?

_No response_

### References

_No response_

### Help Wanted

### Community Note

krzema12:
### Related issue

Fixes #57

### Description

Adds extra validation (enabled with a feature flag) that catches constructs outside of stacks. Such constructs aren't rendered in output of any stack, which is misleading. One can assume that there's some app's implicit stack or that the resource will somehow be rendered and deployed, but it's not true. This change is a step towards stricter validation of this case.

⚠️ BREAKING CHANGE
If the feature flag is enabled, apps having constructs with the app itself as their scope will fail to synthesize. Only constructs with stacks or their descendants as their scope are legal.

### Checklist

- [X] I have updated the PR title to match [CDKTN's style guide](https://github.com/open-constructs/cdk-terrain/blob/main/CONTRIBUTING.md#pull-requests-1)
- [X] I have run the linter on my code locally
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the [documentation](https://github.com/open-constructs/cdk-terrain/tree/main/website) if applicable
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works if applicable
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/packages/cdktn/lib/features.ts
+++ b/packages/cdktn/lib/features.ts
@@ -29,4 +29,6 @@
 export const FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS =
   "cdktn:failOnConstructsOutsideOfStacks";
 
-export const FUTURE_FLAGS = {};
+export const FUTURE_FLAGS = {
+  FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS: true,
+};
--- a/packages/cdktn/lib/features.ts
+++ b/packages/cdktn/lib/features.ts
@@ -22,4 +22,11 @@
  *
  * Tests must cover the default (disabled) case and the future (enabled) case.
  */
+/**
+ * When enabled, synthesis will throw an error if any constructs are scoped
+ * directly to the App instead of to a TerraformStack.
+ */
+export const FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS =
+  "cdktn:failOnConstructsOutsideOfStacks";
+
 export const FUTURE_FLAGS = {};
--- a/packages/cdktn/lib/features.ts
+++ b/packages/cdktn/lib/features.ts
@@ -27,8 +27,8 @@
  * directly to the App instead of to a TerraformStack.
  */
 export const FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS =
-  "cdktn:failOnConstructsOutsideOfStacks";
+  "failOnConstructsOutsideOfStacks";
 
 export const FUTURE_FLAGS = {
-  FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS: true,
+  [FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS]: "true",
 };
--- a/packages/cdktn/lib/app.ts
+++ b/packages/cdktn/lib/app.ts
@@ -7,7 +7,12 @@ import { DISABLE_STACK_TRACE_IN_METADATA } from "./annotations";
 import { Manifest } from "./manifest";
 import { ISynthesisSession } from "./synthesize";
 import { TerraformStack } from "./terraform-stack";
-import { appValidationFailure, noAppFound } from "./errors";
+import {
+  appValidationFailure,
+  constructsOutsideOfStacks,
+  noAppFound,
+} from "./errors";
+import { FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS } from "./features";
 
 const APP_SYMBOL = Symbol.for("cdktf/App");
 export const CONTEXT_ENV = "CDKTF_CONTEXT_JSON";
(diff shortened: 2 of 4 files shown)


packages/cdktn/lib/features.ts
```diff
@@ -22,4 +22,11 @@
  *
  * Tests must cover the default (disabled) case and the future (enabled) case.
  */
+/**
+ * When enabled, synthesis will throw an error if any constructs are scoped
+ * directly to the App instead of to a TerraformStack.
+ */
+export const FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS =
+  "cdktn:failOnConstructsOutsideOfStacks";
```
jsteinich: There's a little bit more that we should do to follow the feature flag pattern: https://github.com/open-constructs/cdk-terrain/blob/main/CONTRIBUTING.md#feature-flags

jsteinich (changes requested):

packages/cdktn/lib/features.ts
krzema12: Ah, I forgot to add the BREAKING CHANGE section - it's now in the PR description. Did I miss anything else?

packages/cdktn/lib/features.ts
```diff
@@ -22,4 +22,11 @@
...
+
 export const FUTURE_FLAGS = {};
```
jsteinich:
```suggestion
export const FUTURE_FLAGS = {
  [FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS]: "true",
};
```

packages/cdktn/lib/features.ts
jsteinich: I added a suggestion with the other part (might not have the correct indent)

krzema12: Sorry, I forgot to document it.

When getting a "yes" answer to my question about the behavior being opt-in (https://github.com/open-constructs/cdk-terrain/issues/57#issuecomment-4095623706), I assumed that it should be opt-in in all contexts. If we add it in `FUTURE_FLAGS`, for new projects it will be the default behavior.

However, you're right that extending `FUTURE_FLAGS` is good according to the currently documented feature flag pattern, so let me add it: https://github.com/open-constructs/cdk-terrain/pull/93/commits/f561be234bba773dca27db8a9a9a02cca910daf2

packages/cdktn/lib/features.ts
krzema12: Hmm, why is the key a list? I inspected the code that consumes `FUTURE_FLAGS` and it looks like everywhere it's just a key (a primitive value). I see some previous feature flags here: https://github.com/hashicorp/terraform-cdk/commit/3c323e76df868ffef8f6b7990941b2f3ce170a74#diff-d279f458811fa5a97343f4b89bda2490c61600c81ace411663b32912cedb2244, but not clear to me why the key type was this way.

krzema12: Whatever we establish here, it should probably land in a form of a type hint, for future disambiguation.

krzema12: TODO on me: test it with `cdktn init`.

jsteinich: The key isn't a list. It's just a different syntax used to initialize when the key comes from a variable. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#computed_property_names.

Adding type information so that it's more explicit does some like a good idea.

krzema12: Oh, ok - TIL!

krzema12: @jsteinich fixed - however, I failed to add a reasonable type hint. The value is a string, but the key looks like just some property which name is taken from a string. That's why I skipped adding the type hint, to focus on converging in this PR. We can always follow up separately.

jsteinich (approved):

PR #110, opened 2026-04-18
feat(lib): add StringMapMap class

### LINKED ISSUE #10: Provider bindings reference non-existent `StringMapMap`
### Expected Behavior

Can build provider bindings for the latest `oci` provider

### Actual Behavior

```.gen/providers/oci/data-oci-ai-language-job/index.ts(67,38): error TS2551: Property 'StringMapMap' does not exist on type```

### Steps to Reproduce

1. Add `oci` provider
2. Run `cdktn get`

### Versions

Unreleased

### Providers

_No response_

### Gist

_No response_

### Possible Solutions

- Add the support (potentially rebasing the existing PR from cdktf)
- Improve code generation to either skip incompatible types or generate as `any`

### Workarounds

_No response_

### Anything Else?

_No response_

### References

- https://github.com/hashicorp/terraform-cdk/issues/3915
- https://github.com/hashicorp/terraform-cdk/pull/3917

### Help Wanted

### Community Note

### PR DESCRIPTION
**Pull request by OrfeasZ**
### Related issue

Fixes #10 

### Description

This was originally sent [as a PR against the CDKTF repo](https://github.com/hashicorp/terraform-cdk/pull/3917) by @nikolaymatrosov. I've simply adjusted it for CDKTN and re-submitted here with their permission. Original description follows below:

This PR adds support for `StringMapMap` type to handle provider schemas that define attributes with `map[map[string]]` type structure. The yandex-cloud provider (and potentially others) define certain attributes like `airflow_config` with a nested map structure that wasn't previously supported by the CDKTF code generation.

The implementation includes:

1. **New `StringMapMap` class**: Added to `packages/cdktf/lib/complex-computed-list.ts` that extends `ComplexResolvable` and provides a `lookup(key: string): StringMap` method for accessing nested maps
2. **Provider generator updates**: Modified `attribute-type-model.ts` to detect when the element type is `StringMap` and generate the appropriate `StringMapMap` type instead of the generic pattern
3. **Comprehensive tests**: Added test fixtures and snapshots to ensure the new functionality works correctly

The fix enables users to successfully generate providers that use nested map structures in their schemas, resolving the JSII compilation errors that were preventing provider generation.

### Checklist

- [x] I have made corresponding changes to the [documentation](https://github.com/open-constructs/cdk-terrain/tree/main/website) if applicable

### DIFF
--- a/packages/@cdktn/provider-generator/lib/get/generator/models/attribute-type-model.ts
+++ b/packages/@cdktn/provider-generator/lib/get/generator/models/attribute-type-model.ts
@@ -393,6 +393,10 @@ export class MapAttributeTypeModel implements CollectionAttributeTypeModel {
   }
 
   get storedClassType() {
+    // Special case for map of map of string
+    if (this.elementType.storedClassType === "StringMap") {
+      return "StringMapMap";
+    }
     return `${this.elementType.storedClassType}Map`;
   }
 
--- a/packages/@cdktn/provider-generator/lib/get/generator/models/attribute-type-model.ts
+++ b/packages/@cdktn/provider-generator/lib/get/generator/models/attribute-type-model.ts
@@ -393,10 +393,6 @@ export class MapAttributeTypeModel implements CollectionAttributeTypeModel {
   }
 
   get storedClassType() {
-    // Special case for map of map of string
-    if (this.elementType.storedClassType === "StringMap") {
-      return "StringMapMap";
-    }
     return `${this.elementType.storedClassType}Map`;
   }
 
--- a/packages/@cdktn/provider-generator/lib/get/__tests__/generator/__snapshots__/types.test.ts.snap
+++ b/packages/@cdktn/provider-generator/lib/get/__tests__/generator/__snapshots__/types.test.ts.snap
@@ -3967,6 +3967,97 @@ export class ListOfStringMap extends cdktn.TerraformResource {
 "
 `;
 
+exports[`map of map of string attribute 1`] = `
+"// https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/map_of_map_of_string
+// generated from terraform resource schema
+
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: packages/@cdktn/provider-generator/lib/get/generator/models/attribute-type-model.ts [hunk 1]
```diff
@@ -393,6 +393,10 @@ export class MapAttributeTypeModel implements CollectionAttributeTypeModel {
   }
 
   get storedClassType() {
+    // Special case for map of map of string
+    if (this.elementType.storedClassType === "StringMap") {
```
jsteinich: This shouldn't be needed as the current generation doesn't have any defensive checks that the code generated is actually valid.

Ideally we do eventually add such validation, but hopefully we can make that a bit more general purpose so that a special case like this isn't needed.

jsteinich (changes requested):
It would be nice if we added a couple tests showing that referencing / looking up a property from such a type results in the correct Terraform being synthesized, but existing test coverage for these sort of complex types isn't great.

OrfeasZ:
I've removed the redundant check and added a couple of test cases. The generated terraform code seems to match what I'd expect. 

One more thing: Should I also add classes for other types of nested maps like `BooleanMapMap`, `NumberMapMap` and `AnyMapMap`? I'm not sure if any providers use these and how to check, but might be good to have for future proofing.

jsteinich (approved):

jsteinich:
> I've removed the redundant check and added a couple of test cases. The generated terraform code seems to match what I'd expect.

Thanks

> One more thing: Should I also add classes for other types of nested maps like `BooleanMapMap`, `NumberMapMap` and `AnyMapMap`? I'm not sure if any providers use these and how to check, but might be good to have for future proofing.

We could, but I'm not sure that anything is actually using them, so could be a follow-up PR.

PR #112, opened 2026-04-21
fix(lib): ForExpression missing iteratorContext and OperatorExpression drops falsy right operands

### LINKED ISSUE #111: CDKTN: `ForExpression` missing `iteratorContext` and `OperatorExpression` drops falsy right operands
### Expected Behavior

1. `TerraformIterator.fromComplexList(list, "key").forExpressionForList(expr)` should iterate the raw list directly (as it does when used via `TerraformDynamicExpression`), producing `[ for key, val in <list>: <expr> ]`.

2. `Op.gt(someExpr, 0)` (and other operators with falsy right-hand operands like `0`, `false`, `""`) should render correctly, e.g. `${(length(...) > 0)}`.

### Actual Behavior

1. `forExpressionForList` on a `fromComplexList` iterator always produces the map conversion `{ for key, val in <list>: val.<keyAttr> => val }` instead of using the raw list. This causes a `Duplicate object key` error when multiple items share the same key attribute value.

2. `Op.gt(someExpr, 0)` renders as `${(length(...) > undefined)}` because `0` is falsy, causing a Terraform `Unknown variable "undefined"` error.

### Bug 1: `ForExpression` missing `iteratorContext`

1. Create a data source that returns a list of objects where the key attribute is not unique:
   ```typescript
   const data = new DataSource(this, "data", {});
   const it = TerraformIterator.fromComplexList(data.items, "name");
   ```
2. Use `forExpressionForList` to filter the list:
   ```typescript
   new Resource(this, "res", {
     value: Fn.element(it.forExpressionForList(`val.id if val.name == "foo"`), 0),
   });
   ```
3. Run `cdktn synth` and observe the generated JSON uses the map conversion:
   ```
   element([ for key, val in { for key, val in data.source.data.items: val.name => val }: val.id if key == "foo"], 0)
   ```
4. Run `terraform plan` and observe `Duplicate object key` error.

### Bug 2: `OperatorExpression` drops falsy operands

1. Use any operator with `0`, `false`, or `""` as the right operand:
   ```typescript
   Op.gt(Fn.lengthOf(someList), 0)
   ```
2. Run `cdktn synth` and observe the output contains `> undefined` instead of `> 0`.

### Versions

language: typescript
cdktn-cli: 0.22.0
node: v22.16.0
cdktn: 0.22.0
constructs: 10.3.0
jsii: 5.8.9
terraform: 1.5.5
arch: arm64
os: darwin 26.3.1

### Migration Context

### Upstream cdktf Issue

_No response_

### Providers

N/A

### Gist

_No response_

### Possible Solutions

Both fixes are in `packages/cdktn/lib/tfExpression.ts`:

### Bug 1: `ForExpression.resolve()` does not set `iteratorContext`

`DynamicListTerraformIterator._getForEachExpression()` returns a `Lazy` that checks `context.iteratorContext` — returning the raw list for `"FOR_EXPRESSION"` and the map conversion for the default case. `TerraformDynamicExpression.resolve()` correctly sets `context.iteratorContext = "FOR_EXPRESSION"` before resolving, but `ForExpression.resolve()` does not.

**Fix:** Add `context.iteratorContext = "FOR_EXPRESSION"` in `ForExpression.resolve()`:

```diff
  public resolve(context: IResolveContext): string {
    const suppressBraces = context.suppressBraces;
    context.suppressBraces = true;
+   context.iteratorContext = "FOR_EXPRESSION";
    const key = this.resolveArg(context, FOR_EXPRESSION_KEY);
    const value = this.resolveArg(context, FOR_EXPRESSION_VALUE);
    const input = this.resolveArg(context, this.input);
```

### Bug 2: `OperatorExpression.resolve()` uses truthiness check instead of undefined check

Line 325 uses `this.right ? ... : undefined` which treats all falsy values (`0`, `false`, `""`) as missing. The `-` operator branch (line 337) has the same bug.

**Fix:**

```diff
- const right = this.right ? this.resolveArg(context, this.right) : undefined;
+ const right = this.right !== undefined ? this.resolveArg(context, this.right) : undefined;
```

```diff
  case "-": {
-   if (right) {
+   if (right !== undefined) {
      // subtraction
      expr = `(${left} - ${right})`;
    } else {
```

### Workarounds

- **Bug 1:** Use raw HCL expression strings instead of `forExpressionForList`:
  ```typescript
  containerRepoId: `\${[for c in ${data.fqn}.items : c.id if c.name == "foo"][0]}`,
  ```
- **Bug 2:** Avoid using `Op` with falsy literal values. Use raw HCL strings for conditions instead.

### Anything Else?

_No response_

### References

- [CDKTF Iterators documentation](https://developer.hashicorp.com/terraform/cdktf/concepts/iterators)
- [Terraform for Expressions](https://developer.hashicorp.com/terraform/language/expressions/for)

### Help Wanted

### Community Note

X-Guardian:
### Related issue

Fixes #111 

### Bug 1: `ForExpression.resolve()` did not set `context.iteratorContext = "FOR_EXPRESSION"`

Inside a for-expression body two things broke:

1. `TerraformIterator._getKey()` / `_getValue()` fell through to `"DYNAMIC_BLOCK"` and rendered `each.key` / `each.value` instead of `key` / `val`.
2. `DynamicListTerraformIterator._getForEachExpression()` used a context-dependent `Lazy` that returned the map-wrapped form `{ for k,v in list : v.<name> => v }`, producing a doubly-nested `[ for key, val in { for k,v in list : v.name => v } : ...]` instead of `[ for key, val in list : ...]`.

#### `DynamicListTerraformIterator` - remove context dependency

The root cause of issue 2 was that` _getForEachExpression()` served two purposes — map-wrapped for `for_each`, raw list for for-expression bodies — and switched on `context.iteratorContext`.

Instead, split the concerns:

- `_getForEachExpression()` - always the map-wrapped form (single purpose: produce a `for_each` value).
- `_getForExpressionInput()` - new `@internal` hook on `ITerraformIterator`, defaults to `_getForEachExpression()`. `DynamicListTerraformIterator` overrides it to return `this.list` raw.
- `keys()` / `values()` / `pluckProperty()` / `forExpressionForList()` / `forExpressionForMap()` and `TerraformDynamicExpression.resolve()` now use `_getForExpressionInput()`.

`ListTerraformIterator` / `MapTerraformIterator` / `ResourceTerraformIterator` are unaffected (they inherit the default).

#### `ForExpression.resolve()` - retain the `iteratorContext` set

`context.iteratorContext = "FOR_EXPRESSION"` is still set, but with this design it has only one job: telling `_getKey() `/ `_getValue()` to render `key` / `val`. It no longer influences `DynamicListTerraformIterator`, so there is no longer a risk that the broadcast flag silently changes a nested iterator's shape depending on what resolves it. 

### Bug 2: `OperatorExpression.resolve()` uses truthiness check instead of undefined check

Line 325 uses `this.right ? ... : undefined` which treats all falsy values (`0`, `false`, `""`) as missing. The `-` operator branch (line 337) has the same bug.

**Fix:**

```diff
- const right = this.right ? this.resolveArg(context, this.right) : undefined;
+ const right = this.right !== undefined ? this.resolveArg(context, this.right) : undefined;
```

```diff
  case "-": {
-   if (right) {
+   if (right !== undefined) {
      // subtraction
      expr = `(${left} - ${right})`;
    } else {
```

### Checklist

**Note: This solution is heavily LLM generated.**


packages/cdktn/lib/tfExpression.ts
```diff
@@ -402,6 +405,7 @@ class ForExpression extends TFExpression {
   public resolve(context: IResolveContext): string {
     const suppressBraces = context.suppressBraces;
     context.suppressBraces = true;
+    context.iteratorContext = "FOR_EXPRESSION";
```
jsteinich, line 408: I don't believe this is the correct solution. It does work for your example and test case, but I believe it will fail if the input expression actually references a map.

I wonder if `DynamicListTerraformIterator` should resolve the for each expression with a value set at creation time rather than relying on the current context. Or perhaps the `ForExpression` needs more information at creation time about the `input` expression.

X-Guardian: Thanks for the review @jsteinich, I have updated the PR if you would like to review again.

jsteinich: I still think this explicit context setting needs to be removed, though I'm not certain what case will actually hit it in a way that matters.

X-Guardian: This is following the same pattern used at https://github.com/open-constructs/cdk-terrain/blob/0d5c3be4f83fda946a347b3c754437da9d22c3f7/packages/cdktn/lib/terraform-dynamic-expression.ts#L47

I've added a regression test that fails without this setting:

```typescript
test("forExpression sets iteratorContext so iterator.getString in body renders as val.<attr>", () => {
  const app = Testing.app();
  const stack = new TerraformStack(app, "test");
  const variable = new TerraformVariable(stack, "list", {});
  const it = TerraformIterator.fromList(variable.listValue as any);

  // Hand-rolled use of the advanced forExpression() API. iterator.getString
  // routes through _getValue(), which is a Lazy that reads
  // context.iteratorContext. ForExpression.resolve() must set it to
  // "FOR_EXPRESSION" so the body renders `val.id` rather than `each.value.id`.
  new TestResource(stack, "test", {
    name: Token.asString(
      forExpression(Fn.toset(variable.listValue as any), it.getString("id")),
    ),
  });

  const synth = JSON.parse(Testing.synth(stack));
  expect(synth).toHaveProperty(
    "resource.test_resource.test.name",
    "${[ for key, val in toset(var.list): val.id]}",
  );
});
```

jsteinich (approved):

PR #116, opened 2026-04-23
chore: emit compiled TS to build/ instead of next to sources

### LINKED ISSUE #115: @cdktn packages: Refine compilations to use dedicated build folders
### Description

Currently, all the `@cdktn` packages are configured so that their typescript compilations emit the built `.js` and `.d.ts` files at the same file location as the typescript files. This causes a lot of file noise when developing and causes confusion when searching across files etc.

A better pattern would be to use a dedicated `build` directory for the compiled output.

The packages are also inconsistent with their source folder naming, with some using `src` and others using `lib`. It would be better if these were consistent.

### References

_No response_

### Help Wanted

### Community Note

X-Guardian:
### Related issue

Fixes #115 

### Description

- Redirect TypeScript compilation to a dedicated `build/` folder in the seven `@cdktn/*` packages so that `.js`/`.d.ts` no longer land next to `.ts` sources. Each package now sets `rootDir: ./src` and `outDir: ./build`, with `main`/`types` pointing at `build/`.
- Rename `lib/` → `src/` in `@cdktn/provider-generator`, `@cdktn/hcl2cdk`, and `@cdktn/hcl2json` so every package consistently uses `src/` for sources and `build/` for output.
- Fix stale subpath imports exposed by the rename: `@cdktn/hcl2cdk` no longer imports `@cdktn/hcl2json/lib/syntax-tree` (collapsed onto the existing `TFExpressionSyntaxTree` namespace import), and three `hcl2json` test files move from `../lib` to `../src`.

####  Changes per package

| Package | Notes |
|---|---|
| `@cdktn/cli-core` | `main`: `src/lib/index.js` → `build/lib/index.js` |
| `@cdktn/commons` | — |
| `@cdktn/provider-schema` | — |
| `@cdktn/hcl-tools` | Go WASM glue in `wasm/` untouched; `build-go.sh` unchanged |
| `@cdktn/provider-generator` | `lib`→`src` rename; `postbuild` path updated; dropped vestigial cross-package include for `../commons/src/module-schema.ts` (commons re-exports it via `index.ts`) |
| `@cdktn/hcl2cdk` | `lib`→`src` rename; narrowed tsconfig `include` from `**/*.ts` to `src/**/*.ts` so `test/` is no longer compiled to disk (ts-jest runs tests directly) |
| `@cdktn/hcl2json` | `lib`→`src` rename; Go WASM glue untouched |

### Checklist


so0k:
I agree splitting emit target is a good practice, but not sure if modifying workspaced packages to use the build output is the best approach across all packages?

X-Guardian:
Hi @so0k, thanks for the comment. Can you clarify what your concern is? The workspace packages were already using the build output already, this PR doesn't change that.

jsteinich:
I've been going a bit back and forth on this.

In regards to the `src` changes:

- The consistency is nice
- Looking into historical changes is slightly harder (some tools handle the git rename nicely while others get lost)
- Small concern of possibly breaking changes. I wouldn't expect so, but I haven't validated each package's exports
- Is there some other benefit?

In regards to the `build` changes:

- It's definitely common in other languages. I'm less certain about TypeScript.
- Is there likely to be any impact to performance (builds / tests) either positively or negatively)?
- I want to checkout the branch and see how my IDE handles it (just a note for myself as the current structure is handled very nicely)

X-Guardian:
Thanks for the feedback @jsteinich. In response to your comments:

#### Changes from `lib` to `src`

- Purely for consistency.  As a new comer to the repo, It was confusing for me to have inconsistent source file directories for the different packages.

#### `build` changes

- All major Typescript repos, especially monorepos, that I have worked on always output their build files to a separate directory, keeping them away from the source files. I thought that was pretty much standard practice now. Take a look at Nx's [typescript-template](https://github.com/nrwl/typescript-template) repo for an example.
- There is no performance impact either way.
- I had several issues whilst I was creating my first changes to the repo with stale javascript files alongside the typescript files confusing VSCode's tooling.

I'm happy to split the two changes into separate pull requests if you would like.

jsteinich (approved):

jsteinich:
I'm fine with the changes.
@so0k any remaining concerns?

so0k:
it’s ok, my main concern was the requirement that you’d have to build before cross pkg TS resolver works? but maybe the tsconfig already takes care of this

i’m more familiar with TurboRepo recommended set up where internal pkgs are JIT compiled (package.json points to src/*ts files instead of build 

but i think all of the CDKTF pkgs are published so they all need to build if you want to avoid heavy lifting in tsconfig / eslint set up 

i haven’t tried the changes locally but if you have no concerns then I withdraw my concern

X-Guardian:
For info, you had to build the packages already before this change, as their `package.json` `types` reference `index.d.ts` which only exists once the package is built.

PR #142, opened 2026-05-01
chore: Refine provider-integration GitHub workflows

X-Guardian:
### Description

Refine the `provider-integration` GitHub workflows:

### 1. Terraform cache path → `${RUNNER_TEMP}/terraform-cache`

Was previously `/usr/local/share/.cache/terraform`, hard-coded as a Linux path. On Windows runners under Git Bash this resolves into a Git Bash–virtual path that doesn't actually exist on the Windows filesystem, producing this warning every run:

```
There are no cache folders on the disk
```

Replaced with `${RUNNER_TEMP}/...` which is set by the runner on both Linux and Windows. Both `linux_provider` and `windows_provider` updated for consistency.

Side effect: Windows now actually gets a working terraform plugin cache (it never did before — the path was non-existent).

### 2. Fix windows_provider terraform cache key

Was using `${{ matrix.terraform }}` while linux_provider used `${{ inputs.terraform_version }}`. The matrix doesn't expose a `terraform` field, so the windows cache key was silently interpolating to an empty string — every windows matrix entry was sharing a single cache regardless of Terraform version. Aligned to use `inputs.terraform_version` like linux.

### 3. Remove unused steps

Remove 'Install pipenv' and 'Install Go' steps from the `windows_provider` workflow as they not used.

### Checklist


.github/workflows/provider-integration.yml
```diff
@@ -159,6 +159,7 @@ jobs:
         uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version: 1.16.x
+          cache: false # This is disabled because we don't have a go.sum file and setup-go expects it to use caching. Thus, caching is always broken anyways
```
jsteinich: We can potentially remove this step. We now require go 1.25.x, but none of the tests broke. I suspect because a newer version is already installed, though maybe no tests run

X-Guardian: Removed the 'install go' and 'install pipenv' steps, as neither are used in the Windows provider tests

jsteinich (approved):

PR #145, opened 2026-05-01
chore(lib): Move cdktn package to src/ and lib/ layout

X-Guardian:
## Summary

Migrates the `cdktn` package to the standard `src/` (sources) + `lib/` (build output) layout used by every other workspace package. Sources and compile output were previously co-located in `lib/`, which caused stale `.js`/`.d.ts` artifacts to confuse knip after every local `yarn build` (false-positive "unused dependency" reports for `json-stable-stringify`, `yazl`, `@types/json-stable-stringify`, `@types/semver` because TS module resolution was preferring the stale `.d.ts` typings over the actual `.ts` sources).

## Why

- After [#116](https://github.com/open-constructs/cdk-terrain/pull/116) migrated other packages to `src/` → `build/`, `cdktn` was the only holdout still emitting compiled output next to sources.
- Mixed layout meant `knip` would silently fail locally after a build, even though it was clean in CI (CI runs knip on a fresh checkout with no compiled artifacts).

## What changed

**Source layout:**
- 85 `.ts` source files moved: `packages/cdktn/lib/**/*.ts` → `packages/cdktn/src/**/*.ts` (subdir layout preserved: `backends/`, `functions/`, `hcl/`, `private/`, `synthesize/`, `testing/`, `tokens/`, `validations/`)
- `package.json` `"main"` and `"types"` continue to point at `lib/index.js` / `lib/index.d.ts` — distribution surface is unchanged.

**Build configuration ([packages/cdktn/package.json](packages/cdktn/package.json)):**
- Added `"jsii"."tsc": { "outDir": "lib", "rootDir": "src" }` — JSII regenerates `tsconfig.json` to compile from `src/` and emit to `lib/`.
- Added `"jsii"."excludeTypescript": ["src/**/__tests__/**"]` — three colocated `__tests__` files inside `src/testing/` import from `test/helper/` (outside `rootDir`); excluding them keeps JSII's `rootDir` constraint satisfied. Jest still picks them up via `testMatch`.

**Test wiring:**
- 33 test files + 5 helpers updated: `from "../lib"` → `from "../src"` (and `../../lib` → `../../src` for nested `test/helper/` and `test/tokens/` files).
[...]

PR #95, opened 2026-04-09
fix(deps): replace archiver with yazl

dnwlf:
### Related issue

Fixes #94

### Description

This PR replaces CDKTN's dependency on `archiver` which includes a dependency on a version of lodash vulnerable to [CVE-2026-4800](https://nvd.nist.gov/vuln/detail/CVE-2026-4800), with a dependency on [yazl](https://github.com/thejoshwolfe/yazl), which looks to be a small stable package with minimal dependencies. Not sure if this will be considered a pro or a con, but VS Code has a dependency on it.

### Checklist

- [x] I have updated the PR title to match [CDKTN's style guide](https://github.com/open-constructs/cdk-terrain/blob/main/CONTRIBUTING.md#pull-requests-1)
- [x] I have run the linter on my code locally
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the [documentation](https://github.com/open-constructs/cdk-terrain/tree/main/website) if applicable
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works if applicable
- [x] New and existing unit tests pass locally with my changes

PR #148, opened 2026-05-02
chore: replace yazl with fflate

X-Guardian:
### Description

The `archiveSync` function in `packages/cdktn/lib/private/fs.ts` runs from `TerraformAsset._onSynthesize`, which executes inside the synchronous constructs synth lifecycle, so it has to block until the zip is on disk. yazl only exposes a streaming API, so the previous implementation worked around this by re-invoking the current module file as a child Node process via `execSync(`node ${__filename} ...`)`. That hack relies on `__filename` always resolving to a CommonJS-loadable JS file — an assumption that breaks the moment ts-jest (or anything else) causes `__filename` to point at a `.ts` source. Node's strip-types loader applies ESM resolution to such files and rejects the extensionless relative import `from "../errors"`.

fflate ships a real synchronous zip API (`zipSync`) and, for the test rig's HTTP-streaming use case, a streaming `Zip` / `ZipDeflate` pair. Swapping yazl for fflate lets us delete the child-process bridge entirely and the whole class of layout-fragility bugs that come with it.

- `archiveSync` rewritten on `fflate.zipSync` — ~15 lines, no subprocess, no `__filename` dependency, no inline runner script.
- `test/template-server.ts` migrated to `fflate.Zip` + `ZipDeflate`; chunk-by-chunk streaming to the HTTP response is preserved.
- `yazl` + `@types/yazl` removed from `packages/cdktn` and `test`; `fflate` added to both. Root `nohoist` updated so `fflate` lands inside `packages/cdktn/node_modules/` for `bundledDependencies`.

### Checklist


jsteinich (commented):
I do like the simplifications made, but I am a bit concerned with using `fflate`. It hasn't been updated in 2 years which isn't necessarily a deal breaker; however, there are some reported potential compatibility issues with Node 22 and TypeScript 5.9.

so0k:
> It hasn't been updated in 2 years

Owner is responding to discussion - https://github.com/101arrowz/fflate/discussions/248#discussioncomment-16510352

The Node v22 discussion - https://github.com/101arrowz/fflate/issues/227

> This issue does not affect any synchronous functions because no data needs to be transferred between threads. It also doesn't affect most of the asynchronous functions because they do not assume consume the input by default (i.e. they copy memory to the other thread instead of transferring). However, it does affect the streaming asynchronous APIs, including `AsyncZipDeflate` and cousins, because those do consume the input (transfer memory instead of copying) by default.
>
> For now, you can work around the issue by copying all buffers before passing them into a streaming asynchronous API: ... <workaround for `AsyncZipDeflate`>

The metamask project added the workaround, but then replaced `fflate` with `yazl` in this commit

https://github.com/MetaMask/metamask-extension/commit/f9e59045bf5a47fa03c57d1cb2f3250c5349b977#diff-161cb29682beb647abebd2d85da170103ef4db84989977124f1327c9a9b7153e

But they used `fflate` in a webpack plugin, so moving to `yazl` makes sense for them and doesn't apply to us?
>  ... the asset-moving and zip-writing pipeline harder to reason about and harder to optimize. ...

I think we can conclude that Nodev22 should work (they had the workaround in place Dec 2024 -> Apr 2026)

jsteinich:
There is also https://github.com/101arrowz/fflate/issues/242 which could be a concern

so0k:
perhaps @dnwlf has some insights - I didn't realise `yazl` was just introduced in https://github.com/open-constructs/cdk-terrain/pull/95

dnwlf:
> perhaps @dnwlf has some insights - I didn't realise `yazl` was just introduced in #95

I don't have any concerns that weren't already raised in this discussion. I viewed #95 as a short-term solution to the lodash CVE (a lateral change), with an understanding there was already a desire for longer-term improvements. It's ok with me if my short-term solution is **very** short-term.

X-Guardian:
In response to the concerns raised, https://github.com/101arrowz/fflate/issues/227 only affects fflate's async APIs, https://github.com/101arrowz/fflate/issues/242 is a TypeScript-types-only concern on unzip return types. Neither applies to our zipSync-only use case

I've added some Fflate regression tests to prove the compression behaviour.

The `packages/cdktn/lib/private/fs.ts` changes are needed for https://github.com/open-constructs/cdk-terrain/pull/145

jsteinich (approved):

PR #151, opened 2026-05-04
feat(provider-generator): configurable import extensions

gabegorelick:
Currently, code generated by cdktn uses CommonJS for import semantics. No file extensions are emitted, and the `/index` path is omitted. This is problematic as the JS/TS community has largely moved to ESM.

For consumers that want ESM, we add a setting that allows customizing the extension for emitted imports.

See #100

### Related issue

#100

### Description

Add a `typescriptImportExtension` setting that can be specified in cdktf.json. You can set it to `.js` or `.ts`. Omitting it keeps the current behavior (no extension emitted for `import` statements).

### Checklist


packages/@cdktn/cli-core/src/lib/cdktf-config.ts
```diff
@@ -73,6 +75,21 @@ export class CdktfConfig {
     return this.getProperty("codeMakerOutput") as string;
   }
 
+  public get typescriptImportExtension(): ImportExtension | undefined {
+    const raw = this.getProperty("typescriptImportExtension");
```
gabegorelick: Is this the first language-specific setting that customizes our output? Do we need to think more about how this setting is exposed?

packages/@cdktn/commons/src/config.ts
```diff
@@ -251,6 +255,16 @@ export interface Config {
   readonly language?: Language;
   readonly output: string;
   readonly codeMakerOutput: string;
+  /**
+   * File extension to write on emitted TypeScript imports. When set,
+   * folder targets emit `./foo/index<ext>` and file targets emit
+   * `./bar<ext>` — compatible with both CommonJS and ESM. When unset,
+   * folder targets emit `./foo` and file targets emit `./bar` — the
+   * legacy CommonJS-only style relying on Node's resolver to fill in
+   * `/index` and `.js`. Applied only when `language === "typescript"`.
```
gabegorelick: I could be persuaded that we should always emit `/index`. It would simplify the logic here somewhat, and I don't *think* it would break any consumers.

Similarly, we could decide to default to `.js`. But that feels like a riskier change.

jsteinich:
I'm not sure if this matches what we would want long term. It might still be valuable as an intermediate step, but we should probably settle on a long term goal first.

Possible options:

1. Support only CJS
2. Support only ESM
3. Support both CJS and ESM
  a. Always support for both at the same time
  b. Conditional for one is built (this PR)

Option (1) seems like it would be a mistake.
Option (2) could make sense, but we should probably get a wider survey to check for community needs.
For Option (3) with pre-built providers we would want (a). Hopefully that doesn't require generating the code twice. For locally generated providers, perhaps the story is different.

gabegorelick:
Pre-built providers are a little different because they are loaded out of `node_modules` and have their own package.json that specifies they're CommonJS. They're also consumed as JS + .d.ts, and not as raw .ts files. So there's less of a need for them to be ESM. It's only an issue in as much as it's an issue if any of your dependencies are CJS instead of ESM.

Self-built providers are different because they are Typescript designed to be integrated into your own project. So it has to somehow support all the different ways Typescript can be written: not just CJS vs ESM, but also whether the TS will be transpiled to JS or interpreted directly. 

> Hopefully that doesn't require generating the code twice.

If we need, we can have some kind of pre-publish workflow separate from the provider generator code. We already do that to publish JS.

> Support only ESM

Given how much of the JS ecosystem has moved to ESM-only and dropped support for CJS, I think this is reasonable. It's certainly annoying as a consumer, but it's what the community has largely coalesced around.

gabegorelick:
> Support both CJS and ESM

I think this is doable if we emit imports as `/index.js`. Including the extension and the `index` filename works fine in CJS, it's just not typically done since it's unnecessary. 

I didn't do that in this PR since my preference was to not disrupt existing consumers at all, but if we think it's OK to change imports in this way, then I can update this PR so that `typescriptImportExtension` defaults to `.js` and we always include `/index` when importing directories. 

Would that work?

packages/@cdktn/commons/src/config.ts
jsteinich: Given that all of the generated code it TypeScript, couldn't we just always emit as `.ts`?

It also seems like it should be fine to always emit `/index`.

gabegorelick: > Given that all of the generated code it TypeScript, couldn't we just always emit as .ts?

That's actually the opposite of what most TS setups do. It only works if you have `allowImportingTsExtensions` [1] enabled, which is a newer setting, and it's disabled by default. For any setup that transpiles to JS, importing `.js` files is what you typically want.

[1] https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions

so0k:
I do sometimes consider the git diff this would generate for some upgrades (but consumers could rightfully mark generated files as such and have them hidden by default in PR reviews)

if it simplifies the logic to always emit .js and support both CJS and ESM then I think it’s a good trade-off

packages/@cdktn/commons/src/config.ts
jsteinich: Good to know.

Is the value in allowing someone to specify a different file extension mostly for the case where they have `allowImportingTsExtensions` enabled? If so, I'm curious if we'd be better off not having the configuration and just always using `.js`?

gabegorelick: > Is the value in allowing someone to specify a different file extension mostly for the case where they have allowImportingTsExtensions enabled?

It's for https://nodejs.org/api/typescript.html and similar mechanisms that allow running Typescript without transpiling to JS, which is what `allowImportingTsExtensions` enables. As https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-7.html#path-rewriting-for-relative-paths says,

> To be maximally compatible with all these tools, a TypeScript file that’s imported “in-place” must be imported with the appropriate TypeScript extension at runtime.

So while we could emit .js imports to support ESM and CJS, it wouldn't fully enable running with type stripping.

packages/@cdktn/cli-core/src/lib/cdktf-config.ts
jsteinich: We could have a `languageOptions` key that is effectively a discriminated union based on the `language` key. `importExtension` would be the key for the TS/JS language options.

I'm also fine if we just keep with what you have though.

packages/@cdktn/commons/src/config.ts
jsteinich: > I could be persuaded that we should always emit `/index`. It would simplify the logic here somewhat, and I don't _think_ it would break any consumers.
> 
> Similarly, we could decide to default to `.js`. But that feels like a riskier change.

@gabegorelick are you able to make these adjustments?

gabegorelick: I think we should still allow emitting imports with no extension, since there could be setups using custom resolvers that break when extensions are present. Should that be enabled by specifying `languageOptions.importExtension: ""`? 

And then we can just always emit `/index` with no option to disable. I can't imagine that would break anything. If that's wrong, then we can reevaluate.

jsteinich: I think that makes sense.

packages/@cdktn/cli-core/src/lib/cdktf-config.ts
gabegorelick: Updated to use `languageOptions`.

packages/@cdktn/commons/src/config.ts
```diff
@@ -29,6 +29,29 @@ export const LANGUAGES = [
...
 
 const CONFIG_FILE = "cdktf.json";
+
+export interface TypeScriptLanguageOptions {
+  /**
+   * Suffix to append to relative imports.
+   * Defaults to ".js".
+   * Use "" to omit the extension (CommonJS).
+   */
+  readonly importExtension?: string;
```
gabegorelick, line 39: Import extension can be any string. The original version of this PR restricted it to `.ts` or `.js`, but you could use `.cjs`, `.mjs`, `.cts`, `.mts`, or anything really.

gabegorelick:
Looks like switching to .js broke something. Let me look into fixing that.

gabegorelick:
The test failures are highlighting a real issue: the ts-node setup used by the integration tests does *not* work with `.js` imports. It will resolve extension-less imports to .ts files, but not `.js` files to `.ts`. 

I think we should go back to no extension being the default. If we want to make ESM the default in the future, we can, but that's clearly going to be a breaking change for many consumers.

gabegorelick:
Updated to go back to no extension by default. The current state of the PR is:
- Added a `languageOptions` setting to cdktf.json. The only option is `importExtension` for typescript.
- Emitted directory imports now always include `/index`. This is somewhat pointless if you don't use an extension (you're still using CJS), but it allows us to treat `importExtension` as an opaque string and not a CJS or ESM signifier. 

If/when we decide to support ESM by default, then we can change the default value for `importExtension`.

gabegorelick:
@jsteinich @so0k all tests are passing. This is ready for another review.

jsteinich (approved):

gabegorelick:
In testing this locally I realized we neglected to pass this setting all the way through the CLI. See https://github.com/open-constructs/cdk-terrain/pull/179 for the fix.

PR #154, opened 2026-05-04
chore(tests): migrate python integration tests to cdktn-native package matrix

X-Guardian:
### Description

The Python integration tests pinned `cdktf~=0.10.4`, a 4-year-old release whose JSII-generated wheels occasionally trip pip's `assert os.path.exists(pyc_path)` during install - causing flaky CI tests. The pin also asserted against `cdktf-provider-random` package versions that exist only on the legacy `cdktf` peer line, so the tests weren't exercising the package format we actually ship.

Example test failure: https://github.com/open-constructs/cdk-terrain/actions/runs/25256981576/job/74059561406

### Fix

This migrates the three Python integration tests to `cdktn~=0.22.1` and swaps the provider under test to `cdktn-provider-local` (12.0.0 / 12.1.0, peer `cdktn ^0.22.0`). The new fixture exercises the upgrade resolver against a cdktn-native two-version matrix and removes the legacy-wheel flakiness as a side effect.

Each test file now defines its version constants at the top with JSDoc explaining why each version was chosen.

`test-helper.ts` adds `pip install --upgrade pip` after venv activation as a small hardening measure against the same pip wheel-installer bug.

### Checklist


jsteinich:
We can do it in stages, but we should plan to migrate these tests in all languages over to cdktn.

test/python/provider-add-command/test.ts
```diff
@@ -1,48 +1,73 @@
...
         const res = await driver.exec("cdktn", [
           "provider",
           "add",
-          "random@=3.1.3", // this is not the latest version, but theres v0.2.55 of the pre-built provider resulting in exactly this package
+          PROVIDER_TO_ADD,
         ]);
 
         expect(res.stdout).toContain(
-          `Installing package cdktf-cdktf-provider-random @ 0.2.55 using pipenv.`,
+          `Installing package cdktn-provider-local @ ${EXPECTED_PKG_VERSION} using pipenv.`,
```
so0k, line 52: we do have https://github.com/cdktn-io/cdktn-provider-random which is published to pypi
https://github.com/cdktn-io/cdktn-provider-random

but I'm fine with using `provider-local`

EDIT: `provider-random` pkg is 44% larger than `provider-local`  (31 Kb more?)

X-Guardian: The `cdktn-provider-random` provider only has one 'cdtn native' release (v13.0.0). v12.1.0 and v12.1.1 both still depend on cdtf. We need two versions of the provider that are cdktn native to test the upgrade, which `cdktn-provider-local` has: v12.0.0 and v12.1.0.
I kept the provider the same across the add, list and upgrade tests for consistency.

jsteinich (approved):

PR #160, opened 2026-05-04
chore: migrate jest to @nx/jest plugin with swc-jest

X-Guardian:
### Description

The previous setup had the jest configuration duplicated across every package, with `ts-jest` doing TypeScript transforms on every test run and `lerna run` orchestrating package-level scripts that didn't understand the dependency graph. That meant slow tests, repeated boilerplate when adding a new package, and CI workflows that ran every package on every PR even when only one was affected.

This PR adopts the standard Nx jest setup. The motivations:

- **Speed.** swc-jest replaces ts-jest, cutting per-package test time roughly 2-3x in local benchmarks. `nx affected` skips packages whose source (and upstream dependencies) didn't change, so PRs touching one package don't pay the full matrix cost.
- **Less boilerplate.** A shared `jest.preset.js` plus a tiny per-package `jest.config.js` plus a `.spec.swcrc` replaces the bespoke ts-jest config each package carried.
- **Correct build sequencing.** The `@nx/jest` plugin honours `dependsOn: ["^build"]` from `nx.json`, so cross-package tests no longer fail with "module not found" when an upstream hasn't been built. The per-package `pretest: yarn build` workarounds disappear.
- **Coverage visibility.** Coverage is now collected on every CI run and rendered as a markdown table on each job's run page, so reviewers can see which packages were exercised and at what coverage.
- **CI partitioning.** Only three packages actually exercise the terraform binary in their tests; the others were paying the cost of a 2x terraform-version matrix for no reason. Tagging those three (`unit-test:terraform`) lets the rest run in a single, faster job.

### Summary

Migrates the jest test pipeline from per-package `ts-jest` configs driven by `lerna run` to the `@nx/jest` plugin with shared swc-jest transformer, and rewires CI to use `nx affected` with coverage surfacing in the GitHub job summary.

#### Test runner

- 9 packages converted: `cdktn`, `cdktn-cli`, `@cdktn/commons`, `@cdktn/cli-core`, `@cdktn/hcl-tools`, `@cdktn/hcl2cdk`, `@cdktn/hcl2json`, `@cdktn/provider-generator`, `@cdktn/provider-schema`.
- Each package gets a `.spec.swcrc` and a thin `jest.config.js` that extends the new root `jest.preset.js` (which itself extends `@nx/jest/preset`). `ts-jest`, `jest`, and `@types/jest` removed from per-package devDependencies; hoisted to root.
- `@nx/jest/plugin` (registered in `nx.json`) provides the `test` target, replacing per-package `test`, `test:ci`, and `jest-watch` scripts. `test:update` retained for snapshot updates.
- Two Nx tags partition the test surface: `unit-test` (5 packages) and `unit-test:terraform` (4 packages whose tests exercise the terraform binary or need a packaged `dist/` directory: `cdktn`, `@cdktn/provider-schema`, `@cdktn/provider-generator`, `@cdktn/cli-core`). The tags are exact-match-disjoint, so CI selectors become trivial allowlists.
- `@cdktn/hcl2cdk`'s `test` target declares an Nx target-level dependency on `cdktn-cli:build`. Its `globalSetup.ts` spawns `packages/cdktn-cli/bundle/bin/cdktn get`, so the CLI bundle must exist before tests run. Wiring this in nx (rather than as a CI step) makes it work uniformly for `nx affected`, `nx test <pkg>`, and local runs.

#### Root scripts

- Migrated `test`, `test:update` from `lerna run` to `nx run-many`. Removed the now-dead `test:ci` root script and `targetDefaults.test:ci` block.
- Removed per-package `pretest: yarn build` (nx now handles `^build`) and the root `pretest` lint hook (lint runs separately via `quality.yml`).

#### CI

- New `pr-unit.yml` (reusable workflow called from `pr.yml`): two `nx affected` jobs replace what would have been an 8×2 matrix (16 jobs). The `affected_no_terraform` job uses `--exclude='*,!tag:unit-test'`; the `affected_terraform` job runs a 1.6.5/1.5.5 matrix using `--exclude='*,!tag:unit-test:terraform'`.
- `pr-unit.yml` runs `yarn package` conditionally before tests when `@cdktn/hcl2cdk` (no-terraform job) or `@cdktn/cli-core` (terraform job) is in the affected set. Both packages have tests that spawn `cdktn init --dist=...` and read jsii tarballs from the workspace-root `dist/` directory — that's produced by `yarn package` + `tools/collect-dist.sh`, not by an Nx target chain. PRs that don't touch either package don't pay the cost.
- `TF_PLUGIN_CACHE_DIR` (set by the container image) is created in each unit-test job before tests run, so terraform doesn't print misleading "plugin cache dir cannot be opened" errors in the build logs.
- `CHECKPOINT_DISABLE` is now set in `jest.preset.js` (via `process.env` at preset-load time) rather than at the workflow env level. Tests inherit it automatically in both CI and local runs, so local runs no longer hang ~30s on HashiCorp Checkpoint timeouts.
- `nrwl/nx-set-shas` action pinned to v5.0.1; explicit `--base="$NX_BASE" --head="$NX_HEAD"` passed.
- Failures surface as inline annotations via `--reporters=github-actions`.
- `unit.yml` (still called from `release.yml` and `release_next.yml` for single-package runs) updated to `yarn nx test <pkg>` with the same hcl2cdk-conditional `yarn package` step and coverage summary.
- All `npx nx` invocations in workflows replaced with `yarn nx` so they use the workspace-pinned versions.
- Coverage collected on every CI run (`--coverage` plus `coverageReporters` in the preset) and rendered to `$GITHUB_STEP_SUMMARY` via a new `tools/test-coverage-summary.mjs` ESM script that aggregates per-package `coverage-summary.json` into a markdown table (jest has no built-in cross-project aggregator). The script filters out rows where any metric is non-numeric, so packages with no tests (`passWithNoTests: true`, e.g. `@cdktn/hcl-tools`) don't appear as `Unknown%` noise.

#### Local dev workflow

- `yarn nx test <pkg> --watch` is the new iterative test loop (Nx builds the `^build` chain once, jest's watch UI reruns affected tests on each save). Documented in `CLAUDE.md`.

#### knip

- Added `test/` (the integration test workspace) to `knip.jsonc` workspaces with TODO `ignoreDependencies` matching the convention used elsewhere.

#### Test fixes

- `@cdktn/cli-core/src/test/lib/terraform-parallelism.test.ts`: removed the `get()` call from `beforeAll` and the deferred `execMockActive` toggle. The tests assert on `cross-spawn` arguments (terraform apply/plan/destroy parallelism flags) and never read from `.gen/`, so the real provider download in `beforeAll` was unnecessary boilerplate that made the test brittle in environments where it couldn't reach the terraform registry.

#### Workarounds

- `@cdktn/cli-core` — extra `transformIgnorePatterns` for `templates/` (sscaff template hooks rely on sloppy-mode globals) and `moduleFileExtensions: ['js','ts','tsx']` override (sscaff in node_modules ships both `.ts` source and compiled `.js`).
- `cdktn-cli` — same `moduleFileExtensions` override (sscaff again).

### Checklist


X-Guardian:
Blocked by https://github.com/open-constructs/cdk-terrain/pull/166

jsteinich (approved):

PR #161, opened 2026-05-05
chore: cdktn tests: isolate test outdirs to fix mkdirSync race

X-Guardian:
## Why

`packages/cdktn/test/*.test.ts` has multiple test files that construct `new App({ stackTraces: false })` without an explicit `outdir`. `App` defaults `outdir` to `cdktf.out` in the working directory, so every jest worker for the cdktn package targets the same `cdktf.out/` directory. The `App` constructor guards directory creation with:

```ts
if (!fs.existsSync(this.outdir)) {
  fs.mkdirSync(this.outdir);
}
```

There is a narrow TOCTOU window between `existsSync` returning `false` and `mkdirSync` being called: if a sibling jest worker calls `mkdirSync` for the same path during that window, the second worker throws:

```
EEXIST: file already exists, mkdir 'cdktf.out'
  at new mkdirSync (lib/app.ts:123:10)
```

The shared cwd-relative outdir is also a latent correctness issue: tests can interfere with each other's synthesised output without anyone noticing.

Example Job failure: https://github.com/open-constructs/cdk-terrain/actions/runs/25370001238/job/74390757224?pr=159#step:13:25

Testing.app() (in `packages/cdktn/lib/testing/index.ts` already exists and already does the right thing — it allocates a fresh `mkdtempSync(...)` outdir when none is passed. The fix is purely to migrate callers off `new App(...)` and onto `Testing.app(...)`.

### Checklist

### DIFF
--- a/packages/cdktn/test/app.test.ts
+++ b/packages/cdktn/test/app.test.ts
@@ -31,7 +31,7 @@ test("context can be passed through CDKTF_CONTEXT", () => {
     key1: "val1",
     key2: "val2",
   });
-  const prog = new App();
+  const prog = Testing.app({ stubVersion: false, enableFutureFlags: false });
   const node = prog.node;
   expect(node.tryGetContext("key1")).toEqual("val1");
   expect(node.tryGetContext("key2")).toEqual("val2");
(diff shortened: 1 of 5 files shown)


so0k (approved):
looks like good clean up using the `Testing.app` helper in tests

PR #181, opened 2026-05-17
chore: point to correct docs repo

gabegorelick:
https://github.com/open-constructs/cdk-terrain-docs is where the documentation lives now.

### Description

Update link to docs in PR template. Also, delete `website` directory that points to the wrong docs site. 

### Checklist


website/docs/cdktn/docs.json
```diff
@@ -1,1281 +0,0 @@
-{
```
gabegorelick, line 1: Is this file needed for anything? It seems to duplicate what's in https://github.com/open-constructs/cdk-terrain-docs/blob/main/content/docs.json.

website/README.md
```diff
@@ -1,2 +0,0 @@
-> [!IMPORTANT]
```
gabegorelick, line 1: If we want, we could keep this README and update it to point to https://github.com/open-constructs/cdk-terrain-docs/tree/main/content. But I'm not sure what would even reference this `website` dir to need to be redirected.

jsteinich (approved):

PR #186, opened 2026-05-17
chore: remove some usages of cdktf from tests

gabegorelick:
We have some tests that use cdktf providers instead of cdktn.

### Description

Remove dependency on `@cdktf/provider-*` packages from tests. Also, install `cdktn` instead of `cdktf` in tests.

### Checklist

- [x] I have made corresponding changes to the [documentation](https://github.com/open-constructs/cdk-terrain/tree/main/website) if applicable


so0k:
should we pick this back up? We are not really supporting dual dependencies anymore and have been publishing pre-built providers for a while - do we still need to support really old cdktf providers?

gabegorelick:
Yes, we should finish this. The test failures should be fixable if we can get rid of cdktf support.

I can try to find the time to finish this. Otherwise, feel free to pick this up if you want.

test/typescript/provider-list-command/test.ts
```diff
@@ -11,16 +11,12 @@ describe("provider list command", () => {
     }); // reset CDKTF_DIST set by run-against-dist script & disable version check as we have to use an older version of cdktf-cli
```
jsteinich:
```suggestion
    }); // reset CDKTF_DIST set by run-against-dist script & disable version check as we have to use an older version of cdktn-cli
```

jsteinich (approved):

jsteinich:
Superseded by #277

PR #192, opened 2026-05-22
chore: consolidate release workflows for npm OIDC

X-Guardian:
### Description

We can't use npm OIDC trusted publishing today because the same packages publish from two different workflow files, and npm's trusted-publisher config matches on a single workflow filename. Consolidating into one file unblocks OIDC and removes `NPM_TOKEN` from the loop.

## Summary

- Merge `release_next.yml` into `release.yml` so the same workflow file publishes both the release and `@next` channels - a precondition for switching to npm OIDC trusted publishing (npm binds trusted publishers to `repository + workflow filename`).
- Extract the publish phase into a reusable `release-publish.yml` workflow called twice (once per channel). Replaces 10 nearly-identical publish jobs (npm / pypi / maven / nuget / golang × release/next) with 2 callers.
- Add `publib` to root `devDependencies`; replace `npx -p publib publib-X` with `yarn publib-X` so publib is version-pinned and not re-resolved on every run. Required `publib >= 0.2.1009` for OIDC support (we pin to `^0.2.1047`).
- Scaffold the OIDC switch (`permissions: id-token: write`, `NPM_TRUSTED_PUBLISHER: "true"`) as commented-out lines, ready to uncomment once trusted publishers are configured on npmjs.com.

### Checklist


.github/workflows/provider-integration.yml
```diff
@@ -67,6 +67,13 @@ jobs:
           restore-keys: |
             go-${{ runner.os }}-${{ hashFiles('**/go.sum') }}-
             go-${{ runner.os }}-
+      - name: Create TF_PLUGIN_CACHE_DIR if set
```
jsteinich: Seems ok, but this is pretty unrelated.

.github/workflows/release.yml
```diff
@@ -10,10 +10,11 @@ env:
   SENTRY_ORG: cdktn
   SENTRY_PROJECT: cdktn
 
-concurrency:
```
jsteinich, line 13: I think we still want this. Mostly for the case of merging in multiple PRs in quick succession.

.github/workflows/release.yml
```diff
@@ -159,45 +311,35 @@ jobs:
...
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+    uses: ./.github/workflows/release-publish.yml
+    with:
+      channel: release
+      dist_artifact_name: dist-release
+    secrets: inherit
 
-  release_pypi:
-    name: Release to PyPi
+  release_sentry:
```
jsteinich, line 344: This should probably depend on the `publish_release` job as well

jsteinich (changes requested):
There's definitely room for additional consolidation (release and next have some differences, but wouldn't be too much effort to conditionalize), but we could do that in a follow-up PR.

I also think we could just skip the whole `next` flow when we are doing an actual release. It's a bunch of CI time for something that is unlikely to actually be used (it's the same code). That could also be done in a follow-up.

.github/workflows/provider-integration.yml
X-Guardian: Removed. Unintentionally committed

.github/workflows/release.yml
X-Guardian: Restored. This could do with further refinement later as now both release and release-next are now in the same group.

.github/workflows/release.yml
X-Guardian: Added

X-Guardian (commented):
Agreed that there is room for further refinement. In this PR I intentionally tried to not change the behaviour of the current workflows to keep the diff to a minimum.

jsteinich (approved):

PR #199, opened 2026-05-24
chore(tests): Ensure we terminate Verdaccio

gabegorelick:
Verdaccio was introduced in 548aa63 so that tests could install the CLI from an npm registry, similar to how real users install it. Unfortunately, we didn't always kill it after we were done.

### Related issue

Fixes # 

### Description

Integration tests often leave orphaned verdaccio processes around. This can cause subsequent test runs to fail.

### Checklist

### DIFF
--- a/test/local-registry.sh
+++ b/test/local-registry.sh
@@ -16,8 +16,11 @@ function startLocalRegistry {
   echo "Cleaning storage dir ($storage_dir).."
   rm -rf $storage_dir
 
-  nohup npx ${VERDACCIO_PACKAGE:-$default_verdaccio_package} -c $1 &>$tmp_registry_log &
+  # Enable job control so the background job becomes its own process group leader.
+  set -m
+  npx ${VERDACCIO_PACKAGE:-$default_verdaccio_package} -c $1 &>$tmp_registry_log &
   verdaccio_pid="$!"
+  set +m
   
   # Wait for Verdaccio to boot
   echo "Waiting for local Registry to start"
@@ -32,8 +35,16 @@ function startLocalRegistry {
 }
 
 function stopLocalRegistry {
+  # Noop if we never started Verdaccio
+  [ -z "${verdaccio_pid:-}" ] && return 0
+
   # Restore the original NPM and Yarn registry URLs and stop Verdaccio
   npm set registry "$original_npm_registry_url"
   yarn config set registry "$original_yarn_registry_url"
-  kill $verdaccio_pid
+
+  # Kill Verdaccio's process group.
+  # We ignore errors since Verdaccio may have died or never started successfully.
+  kill -TERM -- "-$verdaccio_pid" 2>/dev/null || true
+
+  unset verdaccio_pid
 }
--- a/test/run-against-dist
+++ b/test/run-against-dist
@@ -7,23 +7,12 @@ scriptdir=$(cd $(dirname $0) && pwd)
 source "$scriptdir"/local-registry.sh
 
 function cleanup {
+  set +x
   # Restore the original NPM and Yarn registry URLs and stop Verdaccio
   stopLocalRegistry
 }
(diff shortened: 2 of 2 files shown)


test/local-registry.sh
```diff
@@ -32,8 +35,16 @@ function startLocalRegistry {
 }
 
 function stopLocalRegistry {
+  # Noop if we never started Verdaccio
+  [ -z "${verdaccio_pid:-}" ] && return 0
+
   # Restore the original NPM and Yarn registry URLs and stop Verdaccio
   npm set registry "$original_npm_registry_url"
   yarn config set registry "$original_yarn_registry_url"
-  kill $verdaccio_pid
```
gabegorelick, line 38: This was killing the npx process, but it's not clear that npx was always forwarding it to Verdaccio.

test/local-registry.sh
```diff
@@ -16,8 +16,11 @@ function startLocalRegistry {
   echo "Cleaning storage dir ($storage_dir).."
   rm -rf $storage_dir
 
-  nohup npx ${VERDACCIO_PACKAGE:-$default_verdaccio_package} -c $1 &>$tmp_registry_log &
```
gabegorelick, line 19: The `nohup` here is copied from https://github.com/react/create-react-app/blob/6254386531d263688ccfa542d0e628fbc0de0b28/tasks/local-registry.sh#L12, but it seems counterproductive. With how we call `local-registry.sh`, I don't think we ever want verdaccio to survive if we exit.

so0k:
lgtm, but would like @X-Guardian  comments

test/run-against-dist
```diff
@@ -7,23 +7,12 @@ scriptdir=$(cd $(dirname $0) && pwd)
...
 
 function cleanup {
+  set +x
   # Restore the original NPM and Yarn registry URLs and stop Verdaccio
   stopLocalRegistry
 }
 
-function handle_error {
-  cleanup
-  exit 1
```
gabegorelick, line 16: `exit 1` here will actually mask the real error code. Claude is adamant that bash will preserve the exit code from a trapped ERR. So we don't need a separate error handler.

test/local-registry.sh
```diff
@@ -32,8 +35,16 @@ function startLocalRegistry {
...
+  [ -z "${verdaccio_pid:-}" ] && return 0
+
   # Restore the original NPM and Yarn registry URLs and stop Verdaccio
   npm set registry "$original_npm_registry_url"
   yarn config set registry "$original_yarn_registry_url"
-  kill $verdaccio_pid
+
+  # Kill Verdaccio's process group.
+  # We ignore errors since Verdaccio may have died or never started successfully.
+  kill -TERM -- "-$verdaccio_pid" 2>/dev/null || true
```
gabegorelick, line 47: > since Verdaccio may have died

This was happening a lot to me locally. We weren't always killing verdaccio before we exited, which means that when we `nohup npx verdaccio &` our verdaccio instance would tend to die since the port it wanted was already in use. But everything still works fine since the old verdaccio instance is perfectly capable of serving requests. The only issue is that when we go to kill our verdaccio instance, it's no longer there, which caused us to error out.

X-Guardian:
Please see this comment: https://github.com/open-constructs/cdk-terrain/pull/202#pullrequestreview-4352119378

jsteinich (approved):

gabegorelick:
> Please see this comment: https://github.com/open-constructs/cdk-terrain/pull/202#pullrequestreview-4352119378

@X-Guardian since there seems to be active discussion around the rewrite of the local-registry script, I think it may be best if we merge this PR first. It's easy for #202 to resolve the conflicts introduced by this PR, since it rewrites all of the local-registry script anyway.

X-Guardian (approved):

PR #202, opened 2026-05-24
chore(tests): add verdaccio to package.json

gabegorelick:
We were installing it via npx, which means its dependencies weren't pinned. In the age of supply chain attacks, that's not great.

### Related issue

Fixes # 

### Description

Add `verdaccio` to `test/package.json`

### Checklist


test/local-registry.sh
```diff
@@ -3,20 +3,19 @@
 # Source: https://github.com/facebook/create-react-app/blob/0a827f69ab0d2ee3871ba9b71350031d8a81b7ae/tasks/local-registry.sh
 
 custom_registry_url=localhost:4873
-original_npm_registry_url=`npm get registry`
-original_yarn_registry_url=`yarn config get registry`
-default_verdaccio_package=verdaccio@^6.5.2
+original_npm_registry_url="$(npm get registry)"
+original_yarn_registry_url="$(yarn config get registry)"
```
gabegorelick: Some unrelated cleanup while I'm in here.

test/local-registry.sh
```diff
@@ -3,20 +3,19 @@
...
   # Start local registry
-  tmp_registry_log=`mktemp`
+  tmp_registry_log="$(mktemp)"
   echo "Verdaccio Registry log file: $tmp_registry_log"
   
-  storage_dir="$(dirname $1)/storage"
+  storage_dir="$(dirname "$1")/storage"
   echo "Cleaning storage dir ($storage_dir).."
-  rm -rf $storage_dir
+  rm -rf "$storage_dir"
```
gabegorelick: Since we didn't quote, if `storage_dir` had spaces, we could `rm -rf` the wrong dir!

test/local-registry.sh
```diff
@@ -3,20 +3,19 @@
...
 
-  nohup npx ${VERDACCIO_PACKAGE:-$default_verdaccio_package} -c $1 &>$tmp_registry_log &
+  npx -c "verdaccio -c $1" &> "$tmp_registry_log" &
```
gabegorelick: See #199 for discussion on `nohup`.

X-Guardian (commented):
I've actually already included this change in #170, but took it further and re-wrote the `local-registry.sh` script in javascript and enhanced the verdaccio process handling. Happy to split this out, as #170 is getting very large.

How about we include the local-registry change from #170 in this PR? I have refined it to also resolve #199.

test/local-registry.mjs
```diff
@@ -0,0 +1,250 @@
+#!/usr/bin/env node
```
gabegorelick, line 1: This is from #170, per https://github.com/open-constructs/cdk-terrain/pull/202#pullrequestreview-4352119378.

test/local-registry.mjs
```diff
@@ -0,0 +1,250 @@
...
+/**
```
gabegorelick, line 2: It would be nice to convert this file to Typescript, which we should be able to run directly in Node.

I looked into that, but it causes headaches since the tsconfig for `test/` targets CommonJS. Porting that to ESM is a whole separate issue. There are workarounds we could do, like have a separate tsconfig, but I didn't think it was worth it.

gabegorelick:
> How about we include the local-registry change from https://github.com/open-constructs/cdk-terrain/pull/170 in this PR?

@X-Guardian Done.

so0k:
seems tests are still failing after rebase

```console
   SYNTH_HCL_OUTPUT: false
...
Cleaning storage dir (/__w/cdk-terrain/cdk-terrain/test/storage)..
Verdaccio Registry log file: /tmp/verdaccio-OH0GZw/verdaccio.log
Waiting for local Registry to start
Verdaccio exited before becoming ready (code=127, signal=null).
--- verdaccio log (tail) ---
sh: 1: verdaccio: not found

----------------------------
Error: Process completed with exit code 1.
```

gabegorelick:
> sh: 1: verdaccio: not found

@X-Guardian any ideas why we'd see this only for some tests?

X-Guardian:
We need to install the test dependencies as part of the examples integration test workflow. See https://github.com/open-constructs/cdk-terrain/pull/170/changes#diff-8906835152921ef903f55779586f3a092362f65fca98df94d71801cf974ec95bR141-R142

```
      - name: Install test dependencies
        run: cd test && yarn install --frozen-lockfile --prefer-offline
```

test/local-registry.mjs
X-Guardian: ESM plus JSDoc is pretty close to typescript anyway.

gabegorelick: It's close, but definitely not as ergonomic.

gabegorelick:
> We need to install the test dependencies as part of the examples integration test workflow.

Thanks, updated. Let's see if that's enough to fix all the errors. 🤞

test/local-registry.mjs
```diff
@@ -0,0 +1,250 @@
...
+  rmSync(storageDir, { recursive: true, force: true });
+
+  const logDir = mkdtempSync(join(tmpdir(), "verdaccio-"));
+  const logPath = join(logDir, "verdaccio.log");
+  const logFd = openSync(logPath, "w");
+  console.log(`Verdaccio Registry log file: ${logPath}`);
+
+  const originalRegistry = npmGetRegistry();
+
+  const child = spawn("npm", ["exec", "-c", `verdaccio -c ${configPath}`], {
```
gabegorelick, line 76: @X-Guardian should we use https://www.verdaccio.org/docs/verdaccio-programmatically#using-the-module-api to run verdaccio in-process? Seems like we want it to be tied to the lifetime of the calling process anyway.

jsteinich: That sort of seems ideal, though we'd probably need to restructure things as the registry is started and stopped at separate points.

If we are going through all this effort, it would be nice to unify the windows side. As is, they are drifting a part quite a bit which will be harder to maintain.

X-Guardian: One of the benefits of using javascript is that it works in Linux, Mac and Windows. No need to maintain separate scripts. I would suggest making these refinements in a follow up PR, or the scope of this PR is going to grow and grow...

X-Guardian:
@gabegorelick, it would be great to get this merged if you could resolve the conflicts.

gabegorelick:
> it would be great to get this merged if you could resolve the conflicts.

Done. Conflicts are resolved.

X-Guardian (approved):

jsteinich (approved):

test/local-registry.mjs
gabegorelick: > I would suggest making these refinements in a follow up PR

https://github.com/open-constructs/cdk-terrain/pull/248

PR #212, opened 2026-05-25
chore: Pin node version to 22.22.2 via .nvmrc

X-Guardian:
### Description

Pin the CI/dev Node.js version to a single source of truth (`.nvmrc`), so we stop relying on whatever Node ships in the `terraconstructs/jsii-terraform` container or the GitHub-hosted runner image of the week.

**What changed:**

- Added `.nvmrc` containing `22.22.2` (matches the container image's `node22`.
- Updated `mise.toml` to drop the `node` pin and enable `idiomatic_version_file_enable_tools = ["node"]`, so `mise install` reads `.nvmrc` locally — the same file CI uses.
- Updated all 11 workflows under `.github/workflows/` to add (or modify) `actions/setup-node` with `node-version-file: .nvmrc`. This covers all 30 spots that need Node:
  - Container-based jobs (`build`, `unit`, `pr-unit`, `examples`, `integration`, `provider-integration`, `quality`, `release`, `release-publish`, `yarn-upgrade`, `registry-docs-pr-based`) now explicitly install `.nvmrc` Node on top of the image, instead of inheriting an unpinned `node22-nightly`.
  - Windows jobs (`integration`, `provider-integration`) likewise pin via `.nvmrc` instead of inheriting the `windows-2022` runner's bundled Node.
- Added `.nvmrc` and `mise.toml` to Nx's `sharedGlobals` namedInput in `nx.json`, so Nx cache hashes invalidate when the Node version moves. This prevents cache hits from a previous Node version restoring artifacts that may have been built differently.
- Documentation (`CLAUDE.md`, `CONTRIBUTING.md`) now point at `.nvmrc` rather than referencing a hardcoded Node version.

**Why `NODE_NO_WARNINGS=1` in tests:**

Node 22.22 emits `DEP0040` (`punycode` builtin deprecation) for some transitive deps (`jsii-rosetta → @jsii/spec → ajv → uri-js`). Several tests do `expect(result.stderr).toEqual("")` and were failing on the warning text. Setting `NODE_NO_WARNINGS=1` in two places:

- `test/test-helper.ts` (`TestDriver.env`) — applies to all child processes spawned by integration tests.
- `jest.preset.js` — applies to every package's jest run, covering `cdktn-cli`'s `convert.test.ts` / `init.test.ts` that use `execa` directly.

This is forward-compatible: future Node releases adding new deprecation warnings won't suddenly break test stderr assertions.

### Checklist


jsteinich:
I have some reservations to doing this as we are generally using a container with the appropriate dependencies.
This adds an extra install that could fail to tests and potentially causes a mismatch with our tooling.

X-Guardian:
This change has big advantages.

- Removes two 'mystery' dependencies from the CI workflows (node and yarn) and version pins them precisely from metadata in this repo.
- Aligns the CI versions of node and yarn with the local developer versions pinned in the repo. Currently misaligned.
- Deals with corepack configuration automatically, which is required for the pending move to pnpm.
- Follows the standard GitHub workflow node configuration so that this repo isn't 'special'.
- Gives a single point of update for the node version in the future.

jsteinich:
> This change has big advantages.
> 
> * Removes two 'mystery' dependencies from the CI workflows (node and yarn) and version pins them precisely from metadata in this repo.
> * Aligns the CI versions of node and yarn with the local developer versions pinned in the repo. Currently misaligned.
> * Deals with corepack configuration automatically, which is required for the pending move to pnpm.
> * Follows the standard GitHub workflow node configuration so that this repo isn't 'special'.
> * Gives a single point of update for the node version in the future.

There's two main ways to deal with toolchain dependencies in CI.

1. Build an environment with everything installed / setup and use that
2. Reinstall / setup everything each time in a consistent way

There are tradeoffs with each approach. Currently we are using more of the first approach while this change pushes more in the direction of the second.

It is true that (1) somewhat obscures the dependency; however, I do think that tradeoff is worthwhile.
I see the toolchain dependencies as more of an inherited piece from JSII. Using their image as the starting point ensures we are aligned with requirements.
We also have a long CI time, so requiring less setup steps saves time there.

I do support making changes that make the local development experience easier. Adding `.nvmrc` does seem like it helps there. Yes, that does mean 2 spots need updating; however, the node version is changed very infrequently.

Most projects aren't building for 6 languages out of the same repository, so ultimately our project is a bit `special`. Having our process reflect that doesn't seem like a bad thing.

Updating node version still has other considerations. We are using node type definitions which should also be updated to ensure we stay aligned with what is actually present. Going to node 22 in particular causes more work as we need to strip out remaining `cdktf` usages.


Regarding `NODE_NO_WARNINGS=1`:
I understand the sentiment, but I also think it obscures changes that will need to be made. I think being warned about something ahead of it breaking is preferable to not finding out until it breaks. As our node version is effectively the minimum supported version, we aren't really subject to cutting edge things causing a warning, so delaying is amplified.

X-Guardian:
Node version is pinned by major only upstream. The jsii superchain [Dockerfile](https://github.com/aws/jsii-superchain/blob/main/superchain/Dockerfile) pins node to a major version (22) — the patch is whatever was current when the image was last rebuilt. Our digest pin freezes a specific patch (today, 22.22.0), but bumping the image SHA can silently shift the patch with no record in the repo. `.nvmrc` makes the actual version explicit and reviewable, decoupled from when we last bumped the image.

Local dev and CI diverge. `mise.toml` pinned node 20.20; the container ships `22.22.0`. Anyone running tests locally are hitting a different node than CI. Single .nvmrc sourced by both fixes that.

Bumping is one file edit. Today: edit Dockerfile, merge, wait for docker.yml, hand-copy the new digest into ~12 workflow files. With `.nvmrc` the node bump itself is one line.

Setup cost is small. setup-node against GitHub's pre-warmed toolcache is typically less than 5 seconds. Considering the work I have done recently has reduced the CI time from 7 hours plus to less than 1 hour, this is an irrelevance.

Container still does the heavy lifting for jdk/.NET/python/go/gradle/terraform. This PR only takes node-version ownership — the one toolchain that needs to align with package.json's packageManager field and the test runner.

I've removed `NODE_NO_WARNINGS` but having tests that will fail on node deprecation warnings that may occur in transient dependencies that you have no control over is not a good pattern.

jsteinich:
> I've removed `NODE_NO_WARNINGS` but having tests that will fail on node deprecation warnings that may occur in transient dependencies that you have no control over is not a good pattern.

I don't think silently ignoring the deprecations with no other detection until it completely breaks is what we want either. If we can come up with some other mechanism, I'd be happy to add it back in.

so0k:
I'm biased because:
1. I like mise, I introduced mise and use it across several repos with different toolchains on MacOS / Omarchy
2. I don't like devcontainers - I do like the jsii container with all language frameworks (mise has issues installing arm/amd64 for some of the languages - for that use case I think the docker image is good)

I like the PR as it is now, I think letting mise use nvmrc is the right call and having all gh actions point to the same nvmrc

I realise that the docker image currently used not being built in this repo is a problem, that's on me to move it back over

but I don't think that's the main issue raised here about the docker image? for a lot of people they like to build on their local and the other languages are an afterthought (hence the PR to allow only some packages to be build for faster iterations), I think this PR helps them a lot while it keeps the CI/CD on the devcontainer for a consistent environment across all other languages.

as for:

> I've removed `NODE_NO_WARNINGS` but having tests that will fail on node deprecation warnings that may occur in transient dependencies that you have no control over is not a good pattern

I can see the concern that these deprecation warnings (which are out of our control) cause flaky tests, and it takes time to address these deprecations

shouldn't we have another channel to raise them (CI/CD GH Checks -> annotations) so they are flagged on PRs but don't fail the tests?

jsteinich (approved):

so0k:
> I realise that the docker image currently used not being built in this repo is a problem, that's on me to move it back over

I forgot that I already moved the docker build over to this repo!

https://github.com/open-constructs/cdk-terrain/actions/runs/27010747371/job/79713445834#step:9:1482

```console
#14 [auth] terraconstructs/jsii-terraform:pull,push token for registry-1.docker.io
#14 DONE 0.0s

#13 exporting to image
#13 pushing layers 1.6s done
#13 pushing manifest for docker.io/terraconstructs/jsii-terraform:latest@sha256:447c0bca28608f6c5a4298d27464cb2d88b1ea7c27bfc911a103c5e61453a321
#13 pushing manifest for docker.io/terraconstructs/jsii-terraform:latest@sha256:447c0bca28608f6c5a4298d27464cb2d88b1ea7c27bfc911a103c5e61453a321 3.0s done
#13 pushing layers 0.2s done
```

PR #215, opened 2026-05-26
feat(lib): allow disabling creation stacks

### LINKED ISSUE #209: cdktn: excessive cost of captureStackTrace
### Expected Behavior

I'm using simple TypeScript code to generate my resources and CDKTF constructs. I expected it to be a fast process.

### Actual Behavior

When I profiled my script (without even running cdktf synth, just the script directly), I see huge overhead in `captureStackTrace` which is seemingly invoked all the time, whenever I call into a cdktn routine.

### Steps to Reproduce

In a sizable JS/TS cdktn project, capture a performance trace with Chrome devtools, you'll see `captureStackTrace` all over the place in leaf nodes.

After that, test my workaround/hypothesis by executing the same command, attaching the chrome debugger, but then executing `require('cdktf/lib/tokens/private/stack-trace').captureStackTrace = () => []` in the console before recording the performance profile.

Stubbing out `captureStackTrace` gives me a 10x speedup (3 seconds vs 300ms) in my cdktn stack generation. The synth function of cdktn after that runs in about the same time, it's unaffected.

See https://github.com/hashicorp/terraform-cdk/issues/3892

### Versions

language: typescript
cdktn-cli: 0.22.1
node: v24.9.0
terraform: 1.9.2
arch: x64
os: linux 6.14.5

### Migration Context

### Upstream cdktf Issue

https://github.com/hashicorp/terraform-cdk/issues/3892

### Providers

_No response_

### Gist

_No response_

### Possible Solutions

Add a setting to disable these stack traces.

### Workarounds

```js
require('cdktf/lib/tokens/private/stack-trace').captureStackTrace = () => [];
```

### Anything Else?

_No response_

### References

_No response_

### Help Wanted

### Community Note

gabegorelick:
Allow setting `CDKTN_DISABLE_CREATION_STACK` environment variable to disable capturing stack traces when constructing IResolvables.

Capturing stack traces is fairly slow in Node, so disabling them saves a lot of time when synthing large stacks.

### Related issue

Fixes #209

### Description

Don't capture stack traces if the `CDKTN_DISABLE_CREATION_STACK` env var is set.

### Checklist

### DIFF
--- a/packages/cdktn/src/errors.ts
+++ b/packages/cdktn/src/errors.ts
@@ -446,16 +446,15 @@ export const indexTooLargeToEncode = (index: number) =>
 
 export const argToIntrinsicMustBePlainValue = (
   value: any,
-  creationStack: string[],
-) =>
-  new Error(
-    `You can only use a plain value (not a function) while creating an Intrinsic token. We received the value '${value}' created at:\n${creationStack.join(
-      "\n",
-    )}. If you want to use a function, please use the Lazy class. For example:  Lazy.anyValue({ produce: () => "Hello World" }).`,
+  creationStack: string[] | undefined,
+) => {
+  const createdAt = creationStack?.length
+    ? ` created at:\n${creationStack.join("\n")}`
+    : "";
+  return new Error(
+    `You can only use a plain value (not a function) while creating an Intrinsic token. We received the value '${value}'${createdAt}. If you want to use a function, please use the Lazy class. For example:  Lazy.anyValue({ produce: () => "Hello World" }).`,
   );
-
-export const intrinsicNewError = (message: string, createdAt: string) =>
-  new Error(`${message}\nToken created:\n    at ${createdAt}\nError thrown:`);
+};
 
 export const unableToResolveCircularReference = (pathName: string) =>
   new Error(`Unable to resolve object tree with circular reference at '${pathName}'.
--- a/packages/cdktn/src/tokens/private/intrinsic.ts
+++ b/packages/cdktn/src/tokens/private/intrinsic.ts
@@ -1,13 +1,10 @@
 // Copyright (c) HashiCorp, Inc
 // SPDX-License-Identifier: MPL-2.0
 // copied from https://github.com/aws/constructs/blob/e01e47f78ef1e9b600efcd23ff7705aa8d384017/lib/private/intrinsic.ts
-import {
-  argToIntrinsicMustBePlainValue,
-  intrinsicNewError,
-} from "../../errors";
+import { argToIntrinsicMustBePlainValue } from "../../errors";
 import { IResolvable, IResolveContext } from "../resolvable";
 import { Token } from "../token";
-import { captureStackTrace } from "./stack-trace";
+import { captureCreationStack } from "./stack-trace";
 
 /**
  * Token subclass that represents values intrinsic to the target document language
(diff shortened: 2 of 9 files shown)


packages/cdktn/src/tokens/private/intrinsic.ts
```diff
@@ -70,14 +67,6 @@ export class Intrinsic implements IResolvable {
     // and needs resolving.
     return `<unresolved-token>`;
   }
-
-  /**
-   * Creates a throwable Error object that contains the token creation stack trace.
-   * @param message Error message
-   */
-  protected newError(message: string): any {
```
gabegorelick, line 78: This seems to be unused. There are no references to `newError` in the codebase.

Plus, `Intrinsic` isn't supposed to be used externally anyway, right?

I can add it back if it is needed. But it would be nice to get rid of code that uses `creationStack` unnecessarily.

packages/cdktn/src/errors.ts
```diff
@@ -446,16 +446,15 @@ export const indexTooLargeToEncode = (index: number) =>
...
+  creationStack: string[] | undefined,
+) => {
+  const createdAt = creationStack?.length
+    ? ` created at:\n${creationStack.join("\n")}`
+    : "";
+  return new Error(
+    `You can only use a plain value (not a function) while creating an Intrinsic token. We received the value '${value}'${createdAt}. If you want to use a function, please use the Lazy class. For example:  Lazy.anyValue({ produce: () => "Hello World" }).`,
   );
-
-export const intrinsicNewError = (message: string, createdAt: string) =>
```
gabegorelick, line 457: `intrinsicNewError` was only used by `Intrinsic.newError`.

jsteinich:
Found a couple pieces from the AWS CDK that make me wonder if we should just take this further:

- https://github.com/aws/aws-cdk/pull/11170
- https://github.com/aws/aws-cdk/pull/37642

Basically, they flat out removed them from tokens. That would be a breaking change, though I wonder if anyone would actually miss the behavior.

gabegorelick:
For https://github.com/aws/aws-cdk/pull/11170, it looks like they added a `CDK_DISABLE_STACK_TRACE` env var and made stack traces dependent on `CDK_DEBUG` being enabled. That's pretty much what this PR does (minus the `CDK_DEBUG` gate).

But then in https://github.com/aws/aws-cdk/pull/37642, they removed the stack traces completely. I am fine removing them from cdktn as well, if that's what the consensus decides. But like you say, it's more of a breaking change. The change in this PR is non-breaking.

jsteinich:
> For [aws/aws-cdk#11170](https://github.com/aws/aws-cdk/pull/11170), it looks like they added a `CDK_DISABLE_STACK_TRACE` env var and made stack traces dependent on `CDK_DEBUG` being enabled. That's pretty much what this PR does (minus the `CDK_DEBUG` gate).
> 
> But then in [aws/aws-cdk#37642](https://github.com/aws/aws-cdk/pull/37642), they removed the stack traces completely. I am fine removing them from cdktn as well, if that's what the consensus decides. But like you say, it's more of a breaking change. The change in this PR is non-breaking.

I expect we'll have a release with breaking changes in the next month (potentially the next release), so I don't mind a breaking change.

It is probably worth writing up a bit of detail on what we'd lose and post to slack to see if there's any pushback before going forward with that path though.

gabegorelick:
Should we merge this PR and then discuss whether it makes sense to remove the stack traces completely? 

I'd hate for this to get blocked indefinitely while we figure out if we can delete the stack traces. But if you think we can come to a quick consensus, that's fine.

gabegorelick:
In any case, here's a PR to delete creations stacks completely: https://github.com/open-constructs/cdk-terrain/pull/219

I will start a discussion on Slack.

so0k (approved):
LGTM, only tagged this PR in the config surface tracking issue - but it's a new capability in CDKTN so feels fine to use the CDKNT_ prefix

I agree that merging this (potentially for a performance improving patch release with the fast stringify?)

PR #217, opened 2026-05-27
feat(lib): add useLockfile to S3BackendConfig

parthweprom:
### Description

Terraform 1.10 (Dec 2024) deprecated the S3 backend's `dynamodb_table` argument
in favor of `use_lockfile = true`, which enables S3-native state locking via a
`<key>.tflock` object in the state bucket. `S3BackendConfig` was forked from
CDKTF before this feature existed, so the typed prop was missing — users had to
fall back to escape hatches to set it.

This PR adds an optional `useLockfile?: boolean` field to `S3BackendConfig`.
It's a pure pass-through to the rendered backend JSON (no validation,
no interaction with `dynamodbTable`), so users can combine or migrate at their
own pace. Defaults to `false` to preserve existing behavior.

Requires Terraform CLI ≥ 1.10 at apply time; documented in the JSDoc.

### Checklist

- [x] I have updated the PR title to match CDKTN's style guide
- [ ] I have made corresponding changes to the documentation if applicable
- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/packages/cdktn/src/backends/s3-backend.ts
+++ b/packages/cdktn/src/backends/s3-backend.ts
@@ -319,6 +319,20 @@ export interface S3BackendConfig {
    * This behavior does not align with the authentication flow of the AWS CLI or SDK's, and will be removed in the future.
    */
   readonly useLegacyWorkflow?: boolean;
+
+  /**
+   * Whether the backend should use S3-native state locking (Terraform 1.10+).
+   *
+   * When `true`, Terraform acquires a `<key>.tflock` object in the state bucket
+   * for the duration of state-modifying operations. This is the modern
+   * replacement for `dynamodbTable`, which is deprecated in Terraform 1.10+.
+   *
+   * When using this option, ensure your Terraform CLI is at least version 1.10.
+   *
+   * @default false
+   */
+  readonly useLockfile?: boolean;
+
   /**
    * (Optional) The endpoint configuration block.
    */
--- a/packages/cdktn/test/backend.test.ts
+++ b/packages/cdktn/test/backend.test.ts
@@ -265,3 +265,17 @@ test("s3 backend", () => {
 
   expect(Testing.synth(stack)).toMatchSnapshot();
 });
+
+test("s3 backend with use_lockfile", () => {
+  const app = Testing.app();
+  const stack = new TerraformStack(app, "test");
+
+  new b.S3Backend(stack, {
+    bucket: "mybucket",
+    key: "path/to/my/key",
+    region: "us-east-1",
+    useLockfile: true,
+  });
+
+  expect(Testing.synth(stack)).toMatchSnapshot();
+});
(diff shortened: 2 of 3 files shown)


packages/cdktn/src/backends/s3-backend.ts
```diff
@@ -319,6 +319,20 @@ export interface S3BackendConfig {
...
+   *
+   * When `true`, Terraform acquires a `<key>.tflock` object in the state bucket
+   * for the duration of state-modifying operations. This is the modern
+   * replacement for `dynamodbTable`, which is deprecated in Terraform 1.10+.
+   *
+   * When using this option, ensure your Terraform CLI is at least version 1.10.
+   *
+   * @default false
+   */
+  readonly useLockfile?: boolean;
```
so0k, line 334: Given this is moving into past fork territory, we should consider using version validators such as

https://github.com/open-constructs/cdk-terrain/blob/178ea7fce602f66bdc94fb90221b7ff9a8ec55da/packages/cdktn/src/importable-resource.ts#L26-L32

would require a conditional validation on the config object having this set in the constructor

https://github.com/parthweprom/cdk-terrain/blob/9c320b409b81eb71b54a7b264985e8dd5bc40ede/packages/cdktn/src/backends/s3-backend.ts#L17

so0k: confirmed support is in both OpenTofu and Terraform, both on version 1.10+

https://cani.tf/

we should probably maintain our own version of cani.tf in the code (work for another PR)

jsteinich: I agree that having a validation is good to add. I could definitely seeing increasing the minimum supported version in the near future, but not sure we would go all the way to 1.10 in the first jump.

packages/cdktn/test/backend.test.ts
```diff
@@ -265,3 +265,17 @@ test("s3 backend", () => {
...
+
+test("s3 backend with use_lockfile", () => {
+  const app = Testing.app();
+  const stack = new TerraformStack(app, "test");
+
+  new b.S3Backend(stack, {
+    bucket: "mybucket",
+    key: "path/to/my/key",
+    region: "us-east-1",
+    useLockfile: true,
```
jsteinich, line 277: Probably worth a test with `false` as well.

jsteinich (changes requested):

sakul-learning:
Created #237 as a base PR for Terraform/OpenTofu feature-version validation.

It adds:
- deterministic CLI product detection from the first line of plain `version` output (`Terraform v...` vs `OpenTofu v...`)
- `ValidateTerraformFeatureVersion`, which models `feature -> terraform|opentofu -> semver constraint`

This should let this PR depend on/reuse the validator instead of encoding the version/product detection in the S3 backend change itself. For `useLockfile`, the downstream validation would be:

```ts
new ValidateTerraformFeatureVersion("S3 native state locking", {
  terraform: ">=1.10.0",
  opentofu: ">=1.10.0",
});
```

If future features land in Terraform/OpenTofu at different versions, the same validator supports different constraints per CLI product.

jsteinich:
Superceeded by #234

PR #228, opened 2026-05-31
feat(cli)!: rename `--cdktf-version` in `init`

gabegorelick:
The `init` CLI command currently accepts a `--cdktf-version` option. This change renames it to `--cdktn-version`.

### Related issue

Fixes # 

### Description

`cdktn init --cdktf-version` -> `cdktn init --cdktn-version`

This is a breaking change, but should hopefully not be too disruptive.

### Checklist

### DIFF
--- a/packages/@cdktn/cli-core/src/lib/init.ts
+++ b/packages/@cdktn/cli-core/src/lib/init.ts
@@ -34,7 +34,7 @@ export interface RemoteProject extends LocalProject {
 export type Project = LocalProject | RemoteProject;
 
 export type InitArgs = {
-  cdktfVersion?: string;
+  cdktnVersion?: string;
   destination: string;
   dist?: string;
   providers?: string[];
@@ -57,7 +57,7 @@ for (const template of availableTemplates) {
 }
 
 export async function init({
-  cdktfVersion,
+  cdktnVersion,
   destination,
   dist,
   projectId,
@@ -68,8 +68,8 @@ export async function init({
   providersForceLocal,
   silent,
 }: InitArgs) {
-  const deps: any = await determineDeps(
-    cdktfVersion,
+  const deps = await determineDeps(
+    cdktnVersion,
     dist,
     path.basename(templatePath),
   );
@@ -83,8 +83,8 @@ export async function init({
     ...projectInfo,
     futureFlags,
     projectId,
-    sendCrashReports,
-    silent,
+    sendCrashReports: String(!!sendCrashReports),
+    silent: String(!!silent),
   });
   const cdktfConfig = CdktfConfig.read(destination);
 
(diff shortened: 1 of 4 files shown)


packages/@cdktn/cli-core/src/lib/init.ts
```diff
@@ -83,8 +83,8 @@ export async function init({
     ...projectInfo,
     futureFlags,
     projectId,
-    sendCrashReports,
-    silent,
+    sendCrashReports: String(!!sendCrashReports),
+    silent: String(!!silent),
```
gabegorelick, line 87: `deps` was typed as `any`, which masked that these are supposed to be strings.

packages/cdktn-cli/src/bin/cmds/handlers.ts
```diff
@@ -359,12 +363,15 @@ export async function get(argv: {
...
+export async function init(argv: RunInitOptions) {
   await terraformCheck();
   await displayVersionMessage();
   await checkEnvironment();
 
-  if (["", ".", process.cwd()].includes(argv.fromTerraformProject)) {
+  if (
+    argv.fromTerraformProject &&
+    [".", process.cwd()].includes(argv.fromTerraformProject)
+  ) {
```
jsteinich, line 374: Won't this now ignore having an empty string set for `fromTerraformProject` while previously it would have errored?

jsteinich (dismissed):
Can you also prep a PR on https://github.com/open-constructs/cdk-terrain-docs about the change?

Likely going to hold this a bit since I think we'll have at least one more release before introducing breaking changes.

packages/cdktn-cli/src/bin/cmds/handlers.ts
gabegorelick: The `""` case was included in the `.includes` before, but it's now covered in the `argv.fromTerraformProject` conditional. The latter is needed for Typescript to discern that that field is defined.

gabegorelick:
> Can you also prep a PR on https://github.com/open-constructs/cdk-terrain-docs about the change?

https://github.com/open-constructs/cdk-terrain-docs/pull/13

Side note, it would be nice if the docs were merged into this repo so we didn't have split commits.

so0k:
oops, I merged the docs PR thinking this was already merged and part of the 0.23.4 release - I'll revert and re-open the PR on docs.

I think we are comfortable on CI/CD in this repo and can merge docs repo back in (will take some time for me to verify the mintlify configs for it and all..)

sakul-learning:
Thanks for the update. For this flag rename, I’d prefer we make it a two-release transition rather than removing the old spelling immediately:

1. This release: introduce `--cdktn-version`, keep `--cdktf-version` as a deprecated alias, and emit a deprecation warning telling users to migrate.
2. A future release: remove `--cdktf-version` once users have had a release window to update scripts/docs/automation.

Because `init` is a CLI entry point and the command builder is strict, removing `--cdktf-version` outright will make existing automation fail with an unknown option. A yargs-compatible shape would be something like:

```ts
.option("cdktn-version", {
  alias: "cdktf-version",
  type: "string",
  desc: "The cdktn version to use while creating a new project.",
  default: pkg.version,
})
.deprecateOption(
  "cdktf-version",
  "Use --cdktn-version instead. --cdktf-version will be removed in a future release.",
)
```

That keeps the new flag canonical while preserving backward compatibility for one release cycle.

PR #238, opened 2026-06-04
chore(gha): run integration tests on Depot runners

sakul-learning:
## Summary
- Move Linux integration test jobs to Depot's `depot-ubuntu-24.04-16` runner.
- Move provider integration preparation and Linux provider matrix jobs to Depot runners.
- Keep Windows integration/provider jobs on their existing Windows runners.

## Test Plan
- `python3 -c 'import yaml; [yaml.safe_load(open(f)) for f in [".github/workflows/integration.yml", ".github/workflows/provider-integration.yml"]]'`
- `/home/ubuntu/.cache/go-tools/actionlint -ignore 'label "depot-ubuntu-24.04-16" is unknown' -ignore 'constant expression "false" in condition' .github/workflows/integration.yml .github/workflows/provider-integration.yml`
- `git diff --check`

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -28,7 +28,7 @@ concurrency:
 
 jobs:
   prepare-integration-tests:
-    runs-on: depot-ubuntu-24.04-16
+    runs-on: depot-ubuntu-24.04-4
     outputs:
       tests: ${{ steps.build-test-matrix.outputs.tests }}
     container:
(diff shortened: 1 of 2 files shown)


sakul-learning:
Good point. I looked up the Depot runner sizes and `depot-ubuntu-24.04-16` is 16 vCPU / 64 GB, which is larger than these individual integration jobs should need.

GitHub's standard Linux runner for public repos is 4 vCPU / 16 GB, so I switched the Linux integration and provider-integration jobs to `depot-ubuntu-24.04-4` instead. That keeps the job size comparable to the public GitHub-hosted runner capacity while still moving the jobs off the GitHub-hosted runner queue. The main win here is avoiding the long free-runner queue time rather than increasing per-job compute.

Updated in 22fdb7f1d.

so0k (approved):

PR #248, opened 2026-06-07
chore(tests): run verdaccio in-process

gabegorelick:
We currently run verdaccio by spawning an `npm run` process. But since `local-registry` is a JS file now (#202), we can just call verdaccio's JS API directly.

See https://github.com/open-constructs/cdk-terrain/pull/202#discussion_r3300645026

### Related issue

Fixes # 

### Description

In plain English, describe your approach to addressing the issue linked above. For example, if you made a particular design decision, let us know why you chose this path instead of another solution.

### Checklist


test/local-registry.mjs
```diff
@@ -40,18 +26,14 @@ if (command === "start") {
     process.exit(2);
   }
   await start(resolve(configArg));
-} else if (command === "stop") {
-  stop();
```
gabegorelick, line 44: There's no `stop` command anymore. Instead, callers can just kill this process to stop the registry.

test/local-registry.mjs
```diff
@@ -66,126 +48,41 @@ async function start(configPath) {
...
-  child.on("exit", (code, signal) => {
-    exited = true;
-    exitInfo = { code, signal };
-  });
+  const app = await runServer(configPath);
 
   console.log("Waiting for local Registry to start");
-  const ready = await waitForReady(() => exited);
-  if (!ready) {
-    const tail = safeReadTail(logPath, 4000);
```
gabegorelick, line 93: One change is that now verdaccio's output gets written directly to this process's stdout/stderr. I think that's cleaner than selectively tailing, but if it gets too noisy we can try to filter the logs.

test/local-registry.mjs
```diff
@@ -66,126 +48,41 @@ async function start(configPath) {
...
+    );
     process.exit(1);
   }
 
   npmSetRegistry(REGISTRY_URL);
   appendAuthToken();
-
-  writeFileSync(
-    STATE_FILE,
-    JSON.stringify({ pid: child.pid, originalRegistry, logPath }, null, 2),
```
gabegorelick, line 120: The state file was used to track a detached verdaccio process. But we don't need it if we never detach verdaccio.

.copywrite.hcl
```diff
@@ -9,7 +9,6 @@ project {
     ".github/ISSUE_TEMPLATE/*.yml",
     "packages/@cdktn/cli-core/templates/**",
     "test/verdaccio.yaml",
-    "test/local-registry.sh",
```
gabegorelick, line 12: This was missed in #202 when we ported local-registry to JS.

knip.jsonc
```diff
@@ -66,6 +66,11 @@
       "project": ["scripts/**/*.ts"]
     },
     "test": {
+      "entry": [
+        "local-registry.mjs",
+        "translate-edge-tests.js",
+        "provider-tests/generate-provider-test.js"
```
gabegorelick: Adding some other entrypoints that were missed when we added knip.

X-Guardian:
I don't think we should be moving code back into bash scripts. The whole point of the original change was to move away from from these. The node script can manage the whole service, something like:

- `local-registry.mjs start` calls `runServer`, then internally waits for `/-/ping` then daemonises itself and records the child PID to the state file.
- `local-registry.mjs stop` reads the state file, SIGTERMs the child, restores the npm registry

The `run-against-dist` script then goes back to:

```
trap 'node "$scriptdir"/local-registry.mjs stop || true' EXIT
node "$scriptdir"/local-registry.mjs start "$scriptdir"/verdaccio.yaml
...
node "$scriptdir"/local-registry.mjs stop
```

gabegorelick:
> I don't think we should be moving code back into bash scripts. The whole point of the original change was to move away from from these.

If that's the goal, then why not move `run-against-dist` to JS? This PR gets us closer to that, by removing the funky process state file management.

Should I go ahead and do that? Or is there a reason we need `run-against-dist` to remain bash?

jsteinich:
> If that's the goal, then why not move `run-against-dist` to JS? This PR gets us closer to that, by removing the funky process state file management.
> 
> Should I go ahead and do that? Or is there a reason we need `run-against-dist` to remain bash?

I am definitely supportive of moving away from bash entirely.

X-Guardian:
Yep. No reason to have bash scripts in a Node project.

test/run-against-dist.mjs
```diff
@@ -0,0 +1,260 @@
...
+//   ./run-against-dist.mjs <command> [args...]
+
+import { access, glob, mkdtemp, readFile, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { delimiter, join, resolve } from "node:path";
+import { setTimeout as sleep } from "node:timers/promises";
+
+// execa (over child_process.spawn) resolves `.cmd`/`.bat` shims and quotes args
+// correctly on Windows.
+import execa from "execa";
```
gabegorelick, line 21: I went with `execa` since we already had it in package.json. 

We are stuck on an old version because newer versions are ESM only. But eventually we will migrate to ESM.

test/run-against-dist.bat
```diff
@@ -1,39 +0,0 @@
-@echo off
```
gabegorelick, line 1: We don't need a dedicated script for Windows if we write the JS right.

test/run-against-dist.mjs
```diff
@@ -0,0 +1,260 @@
...
+
+/**
+ * Async existence check.
+ *
+ * @param {string} path Filesystem path to test.
+ * @returns {Promise<boolean>} Whether the path exists.
+ */
+async function pathExists(path) {
+  try {
+    await access(path);
```
gabegorelick, line 84: While we tend to do a lot of sync IO in cdktn, I intentionally did not do that in this file so that we don't block verdaccio. While some sync IO would be fine here, other instances would deadlock verdaccio. 

The easiest thing is to just use async IO everywhere so we don't have to worry about it. Yes, we take an ever so slight perf hit over using sync, but it's not going to be perceivable.

test/test-helper.ts
```diff
@@ -75,14 +75,14 @@ export class TestDriver {
...
-    // which shadows the cdktn CLI we want to test against and set in run-against-dist.sh
+    // which shadows the cdktn CLI we want to test against and set in run-against-dist
     // While the code is the same, we want the CLI to be installed in a temporary directory
     // (handled by run-against-dist) to ensure we don't have broken relative paths in
     // that don't work then the CLI is installed via NPM. One example for such a bug we want
     // to catch with this is: https://github.com/hashicorp/terraform-cdk/issues/3025
     // To achieve this, we prepend the path exported by the run-against-dist script that
     // it installed the cdktn CLI in
-    const newPath = `${process.env.TEST_PATH_CDKTF_CLI}:${process.env.PATH}`;
+    const newPath = `${process.env.TEST_PATH_CDKTN_CLI}:${process.env.PATH}`;
```
gabegorelick, line 85: Taking the opportunity to update this env var. This is the only consumer, and the only writer is in `run-against-dist.mjs`.

package.json
```diff
@@ -22,20 +22,20 @@
...
-        "integration:windows:go": "cd test && run-against-dist.bat \"npx jest --runInBand go\"",
+        "integration": "cd test && yarn edge:install && ./run-against-dist.mjs npx jest --runInBand",
+        "integration:update": "cd test && yarn && yarn edge:install && NODE_OPTIONS='--max-old-space-size=7168' ./run-against-dist.mjs npx jest -u --runInBand",
+        "integration:single": "cd test && ./run-against-dist.mjs npx jest --runInBand",
+        "integration:typescript": "cd test && ./run-against-dist.mjs npx jest --runInBand typescript",
+        "integration:python": "cd test && ./run-against-dist.mjs npx jest --runInBand python",
+        "integration:csharp": "cd test && ./run-against-dist.mjs npx jest --runInBand csharp",
+        "integration:java": "cd test && ./run-against-dist.mjs npx jest --runInBand java",
+        "integration:go": "cd test && ./run-against-dist.mjs npx jest --runInBand go",
+        "integration:windows": "cd test && node run-against-dist.mjs npx jest --runInBand",
```
gabegorelick: We could probably delete these windows scripts if we wanted. But we'd have to make `yarn edge:install` cross-platform.

Saving that for a separate effort.

gabegorelick:
Updated to port `run-against-dist` to JS.

gabegorelick:
Sorry, it's taking some back and forth to get the Windows tests working since I don't run Windows locally. Let me put this PR into draft while I fix it.

gabegorelick:
I think the only remaining test failures are flakes. So this is ready for review.

test/provider-tests/test-provider.bat
```diff
@@ -5,4 +5,4 @@ setlocal EnableDelayedExpansion
 set scriptdir=%~dp0
 
 call node %scriptdir%\generate-provider-test.js %~1
-call %scriptdir%\..\run-against-dist.bat npx jest ".\providers\%~1"
```
gabegorelick, line 8: Claude is claiming this was wrong and caused no tests to run.

package.json
jsteinich: Let's make a follow-up issue so that doesn't get lost.

gabegorelick: https://github.com/open-constructs/cdk-terrain/issues/259

jsteinich (approved):

PR #252, opened 2026-06-07
chore: rename CdktfProject to CdktnProject

gabegorelick:
Rename internal "project" symbols from cdktf to cdktn.

This should *not* affect external consumers. This is an internal refactoring only.

### Related issue

Fixes # 

### Description

`CdktfProject` -> `CdktnProject`

### Checklist

### DIFF
--- a/packages/@cdktn/cli-core/src/test/lib/cdktn-project.test.ts
+++ b/packages/@cdktn/cli-core/src/test/lib/cdktn-project.test.ts
@@ -2,11 +2,11 @@
 // SPDX-License-Identifier: MPL-2.0
 import * as fs from "fs-extra";
 import * as path from "path";
-import { CdktfProject, get, init } from "../../lib/index";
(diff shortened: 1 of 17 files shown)


packages/@cdktn/cli-core/src/test/lib/cdktn-project.test.ts
```diff
@@ -26,17 +26,17 @@ function installFixturesInWorkingDirectory(
     workingDirectory: string;
   },
 
-  fixtureName: string
+  fixtureName: string,
```
gabegorelick, line 29: The build seems to force us to fix pre-existing formatting issues whenever a file is edited.

sakul-learning:
Small follow-up: please double-check the ESLint ignore path after the rename. If any generated/template output path moved, the ignore entry should move with it so ESLint doesn’t start scanning generated artifacts — or keep silently ignoring an obsolete path.

PR #253, opened 2026-06-07
chore: rename cdktfVersion context field

gabegorelick:
Christen it `cdktnVersion`.

### Related issue

Fixes # 

### Description

Rename the `cdktfVersion` field we save on node context.

### Checklist

### DIFF
--- a/packages/cdktn/src/app.ts
+++ b/packages/cdktn/src/app.ts
@@ -117,7 +117,7 @@ export class App extends Construct {
       node.setContext(DISABLE_STACK_TRACE_IN_METADATA, true);
     }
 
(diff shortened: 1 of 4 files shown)


packages/cdktn/src/app.ts
```diff
@@ -117,7 +117,7 @@ export class App extends Construct {
       node.setContext(DISABLE_STACK_TRACE_IN_METADATA, true);
     }
 
-    node.setContext("cdktfVersion", version);
+    node.setContext("cdktnVersion", version);
```
gabegorelick, line 120: Should we consider this a breaking change? I think the context is technically visible to consumers, although not sure why they'd use this.

PR #260, opened 2026-06-10
chore: add uv to mise

gabegorelick:
We currently have `pipx:pipenv` in our mise tools. But that fails to install unless uv or pipx are already installed. The solution is to add uv to our mise tools so that dev environments without any existing python tooling can be bootstrapped.

### Related issue

Fixes # 

### Description

The `pipx:pipenv tool` relies on mise's `pipx` backend, which shells out to either pipx or uv. So one of those tools should be in mise.toml.

See https://mise.jdx.dev/dev-tools/backends/pipx.html

### Checklist

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,6 +1,3 @@
-# SpecLedger mise configuration
-# Issue tracking is now built into sl CLI (sl issue commands)
-
 [settings]
 idiomatic_version_file_enable_tools = ["node"]
 
@@ -9,6 +6,7 @@ python = '3.11'
 go = '1.25'
 java = "prefix:corretto-20"
 dotnet = "6.0"  # Match jsii-superchain DOTNET_CHANNEL
+uv = "0.11.19" # Used by the pipx backend
 "pipx:pipenv" = "latest"
 
 terraform = "1.7.5"


mise.toml
```diff
@@ -1,6 +1,3 @@
-# SpecLedger mise configuration
-# Issue tracking is now built into sl CLI (sl issue commands)
```
gabegorelick, line 2: This is more than just a mise config for SpecLedger.

mise.toml
```diff
@@ -9,6 +6,7 @@ python = '3.11'
 go = '1.25'
 java = "prefix:corretto-20"
 dotnet = "6.0"  # Match jsii-superchain DOTNET_CHANNEL
+uv = "0.11.19" # Used by the pipx backend to install pipenv
 "pipx:pipenv" = "latest"
```
gabegorelick, line 10: The `pipx` backend actually recommends uv, so that's what we're adding.

See https://mise.jdx.dev/dev-tools/backends/pipx.html

mise.toml
so0k: agreed, sorry specledger CLI took ownership of too many files

i’m leveraging specledger heavily for the Telemetry work - but with models like Fable coming out it might now be needed much longer

jsteinich (approved):

Issue #206, opened 2026-05-24
Migrate jsii-terraform image off Docker Hub (GHCR for CI, OCF Public ECR for devcontainer)


so0k:
### Description

https://github.com/TerraConstructs/terraform-cdk/blob/main/Dockerfile is the current docker image we are using as it was the first fork.

contributors are confused where the image lives and it should simply live in the open-constructs/cdk-terrain repo.

Dockerhub is not cheap and it's probably more consistent with the current CDK Terrain project to use the OCF Public ECR to host it.

### Help Wanted

- [ ] I'm interested in contributing a fix myself

### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

so0k:
## Scope expansion: this also covers CI reliability, not just the devcontainer

The same `terraconstructs/jsii-terraform` Docker Hub image is the **job `container:` for ~11 CI workflows**, and Docker Hub's anonymous pull throttling on GitHub's shared runner IP pool is now causing intermittent CI failures. Example from PR #62 (head `6ee638bf0`), at container init before any project code runs:

```
docker pull terraconstructs/jsii-terraform@sha256:c46ac91…
Error response from daemon: Get "https://registry-1.docker.io/v2/":
  net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
##[warning]Docker pull failed with exit code 1, back off 9.69 seconds before retry.   (×3 retries, then the job fails)
```

This flaked `Build / build-and-package` and several `Examples Integration Tests` jobs across two consecutive runs (a *different* example each round — classic rate-limit roulette). Digest-pinning does **not** exempt anonymous pulls from the cap.

### How the image is consumed today (audit)

[...]

PR #261, opened 2026-06-10
fix(gha): pull jsii-terraform CI image from GHCR instead of Docker Hub

so0k:
Refs #206 (GHCR-for-CI half; OCF Public ECR for the devcontainer is a follow-up).

## Why

Docker Hub's anonymous pull throttling on GitHub's shared runner IP pool is causing intermittent container-init failures across CI (see the audit in #206). All ~27 `container:` references across 12 workflows pull `terraconstructs/jsii-terraform` anonymously; GHCR public-package pulls from Actions runners are free, unmetered, and unauthenticated.

## What

- **`docker.yml`**: also pushes to `ghcr.io/open-constructs/jsii-terraform` (`packages: write` + `docker/login-action` against `ghcr.io` with `GITHUB_TOKEN`). Docker Hub publishing stays until consumers have moved.
- **`Dockerfile`**: adds `org.opencontainers.image.source` label so future pushes auto-link the GHCR package to this repo (plus description/licenses labels).
- **12 consumer workflows**: `container.image` repointed to `ghcr.io/open-constructs/jsii-terraform@sha256:c46ac91…` — the **same digest** as today, because the image is mirrored registry-to-registry (digest-preserving). No rebuild needed for this PR's CI to pass.

## Prerequisites before marking ready (one-time, maintainer)

- [x] Mirror the current image to GHCR (digest-preserving copy):
  ```sh
  echo $GHCR_TOKEN | docker login ghcr.io -u <user> --password-stdin   # PAT with write:packages
  docker buildx imagetools create \
    --tag ghcr.io/open-constructs/jsii-terraform:latest \
    docker.io/terraconstructs/jsii-terraform@sha256:c46ac91c8be20947a2b0e27a517eb4a5bfc015c16a27341876ed0b04d734ddf5
  ```
- [x] Set the new package **public** (org → Packages → jsii-terraform → settings → Change visibility). New GHCR packages default to private; anonymous CI pulls get 403 until this is done.
- [x] Manually link the package to this repo in package settings (the mirrored manifest predates the source label; the label takes over on the next `docker.yml` push to main).
- [x] Re-run CI on this PR to confirm GHCR pulls resolve, then mark ready.

## Follow-ups (tracked in #206)

- Mirror to OCF Public ECR for devcontainer/contributor distribution; update `.devcontainer.json` + `CONTRIBUTING.md`.
- Retire Docker Hub publishing and the `DOCKERHUB_*` secrets once consumers have moved.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
   build-and-package:
     runs-on: ubuntu-latest
     container:
-      image: terraconstructs/jsii-terraform@sha256:c46ac91c8be20947a2b0e27a517eb4a5bfc015c16a27341876ed0b04d734ddf5
+      image: ghcr.io/open-constructs/jsii-terraform@sha256:c46ac91c8be20947a2b0e27a517eb4a5bfc015c16a27341876ed0b04d734ddf5
     env:
       CHECKPOINT_DISABLE: "1"
     timeout-minutes: 60
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -20,15 +20,10 @@ jobs:
       packages: write
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
-      - name: Cache Docker layers
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-buildx-${{ hashFiles('/Dockerfile', '.terraform.versions.json') }}
-          restore-keys: |
-            ${{ runner.os }}-buildx-
       - name: Login to DockerHub
         uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4.2.0
         with:
@@ -66,8 +61,12 @@ jobs:
             ghcr.io/open-constructs/jsii-terraform:latest
             ghcr.io/open-constructs/jsii-terraform:${{ steps.cdktn-version.outputs.cdktn }}
             ghcr.io/open-constructs/jsii-terraform:${{ steps.cdktn-version.outputs.cdktn }}-${{ steps.git-sha.outputs.git-sha }}
-          cache-from: type=local,src=/tmp/.buildx-cache
-          cache-to: type=local,dest=/tmp/.buildx-cache
+          # Registry-backed layer cache instead of the GitHub Actions cache:
+          # writing to it requires packages:write, which only this workflow's
+          # token has, so it cannot be poisoned from less-privileged contexts
+          # (zizmor cache-poisoning).
+          cache-from: type=registry,ref=ghcr.io/open-constructs/jsii-terraform:buildcache
+          cache-to: type=registry,ref=ghcr.io/open-constructs/jsii-terraform:buildcache,mode=max
           build-args: |
             DEFAULT_TERRAFORM_VERSION=${{ steps.tf-versions.outputs.default }}
             AVAILABLE_TERRAFORM_VERSIONS=${{ steps.tf-versions.outputs.available }}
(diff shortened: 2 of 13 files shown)


so0k:
image
migrated the docker hub image to GHCR for this repo

sakul-learning:
### 🔒 Cache poisoning risks in workflow changes

This PR is clean and correct in intent, but the `docker.yml` cache configuration (pre-existing) has two issues worth flagging:

**1. Bare `restore-keys` prefix — cache poisoning vector**

https://github.com/open-constructs/cdk-terrain/blob/fix/ghcr-ci-image/.github/workflows/docker.yml#L26-L28

```yaml
key: ${{ runner.os }}-buildx-${{ hashFiles('/Dockerfile', '.terraform.versions.json') }}
restore-keys: |
  ${{ runner.os }}-buildx-
```

The `restore-keys` has no hash suffix — it matches ANY previous `buildx` cache regardless of Dockerfile or TF version changes. A stale/poisoned cache from a different Dockerfile version can be silently restored, injecting different layers into the build.

**Fix:** add a hash fallback:
```yaml
restore-keys: |
  ${{ runner.os }}-buildx-${{ hashFiles('/Dockerfile', '.terraform.versions.json') }}
  ${{ runner.os }}-buildx-
```

This prefers the exact-hash match but falls back to any `buildx` cache as a last resort.

**2. `hashFiles('/Dockerfile')` — absolute path may not resolve**

The leading `/` in `hashFiles('/Dockerfile', '.terraform.versions.json')` is an absolute path. In GHA, `hashFiles` patterns are typically workspace-relative. If `/Dockerfile` resolves to nothing, the hash is constant (empty), making all builds share the same cache key — defeating cache isolation.

**Fix:** use `hashFiles('Dockerfile', '.terraform.versions.json')` or `hashFiles('**/Dockerfile', '.terraform.versions.json')`.

---

These are pre-existing, not introduced by this PR, but since this PR touches `docker.yml` it's a good moment to address them.

so0k:
Thanks for flagging this — addressed in 676fcbd, though with a different fix than the suggested key tightening.

I ran `zizmor` to pin down the exact rule: it reports the caching as a high-severity **`cache-poisoning`** finding, and the trigger is the *combination* of a GitHub Actions cache feeding a workflow that **publishes** an image (`docker/build-push-action` with `push: true`). Tightening `restore-keys` or fixing the `hashFiles('/Dockerfile')` path wouldn't clear it: the Actions cache itself is writable from less-privileged contexts (any branch workflow in the repo can populate entries under the same scope), so however precise the key, a poisoned entry could be baked into the published image. That's why I moved off the GHA cache mechanism entirely rather than patching the key.

**Resolution — registry-backed buildx cache:**

```yaml
cache-from: type=registry,ref=ghcr.io/open-constructs/jsii-terraform:buildcache
cache-to: type=registry,ref=ghcr.io/open-constructs/jsii-terraform:buildcache,mode=max
```

- Writing to `:buildcache` requires `packages: write`, which only this workflow's `GITHUB_TOKEN` has (it runs on push-to-`main` / `workflow_dispatch` only; fork PR tokens are read-only and same-repo PRs never run `docker.yml`) — so no less-privileged context can poison it.
- `mode=max` exports all intermediate layers, so caching is as good as before. Better, actually: your two observations were spot-on — the absolute `hashFiles('/Dockerfile')` never resolved and the bare `restore-keys` prefix matched any prior cache, so the old setup had the poisoning exposure *without* meaningful cache isolation.
- First run after merge is a cold build that seeds `:buildcache`; the cache ref shows up alongside the image tags in the GHCR package, which is the standard pattern.

Also set `persist-credentials: false` on the checkout while in there (`zizmor` `artipacked`). `zizmor` now reports **no findings** on `docker.yml`.

sakul-learning:
> *Retracting the cache-poisoning concerns from my earlier comment — 676fcbd removes `actions/cache` entirely and replaces it with registry-backed Buildx cache (`cache-from/cache-to: type=registry`).*

**Re-verified at 676fcbd — this is the right approach:**

- ✅ `actions/cache` removed — no GHA-cache poisoning surface at all
- ✅ Registry cache requires `packages:write` scope, same as the push step — no privilege escalation possible
- ✅ `persist-credentials: false` on checkout prevents token leakage
- ✅ `cache-from: type=registry` pulls cache layers from GHCR, `cache-to: type=registry,mode=max` pushes all layers (not just final image layers) for maximal reuse

**First-run cold cache:** the `:buildcache` tag won't exist in GHCR until the first push-to-main after merge. That first build does a full rebuild. Every subsequent run hits the registry cache — which is also available to any branch that has `packages:read` access to the GHCR package.

jsteinich (approved):

PR #262, opened 2026-06-10
chore: fix generate-function-bindings tool

X-Guardian:
### Description

Repairs the `tools/generate-function-bindings` tool, which broke during the yarn→pnpm migration, and removes its dead library/test/publish scaffolding. No runtime/library behaviour changes — the two generated files (`packages/cdktn/src/functions/terraform-functions.generated.ts` and `packages/@cdktn/hcl2cdk/src/function-bindings/functions.generated.ts`) regenerate **byte-for-byte identically** to what's already committed, so this is tooling-only.

**Why it was broken:** the migration bumped Prettier v2→v3 and tightened pnpm's dependency isolation, and an earlier `@cdktn` restructure had moved sources to `src/`. None of that was reflected in this tool, so `pnpm run generate-function-bindings` failed.

**Fixes (each surfaced as a failure when running the tool):**

1. **Prettier v3 is async.** `prettier.format()` now returns `Promise<string>`; the tool was writing the unresolved promise to disk. Added `await` at both call sites in `generate.ts` and the one in `fetch-metadata.ts`.
2. **Prettier v3's `babel` parser rejects TypeScript.** The generator emits TS type annotations (e.g. `static "abs"(num: number)`), so formatting failed with a syntax error. Switched both `generate.ts` calls from `parser: "babel"` to `parser: "babel-ts"`.
3. **Output paths pointed at `lib/`.** Sources live in `src/` (`lib`/`build` are gitignored tsc output); the tool wrote a stray file into `cdktn/lib/` and failed with `ENOENT` for `hcl2cdk/lib/` (whose build dir is `build/`, not `lib/`). Repointed both output paths to `src/`.
4. **Missing `@babel/*` type declarations.** `@types/babel__generator` and `@types/babel__template` were present at the workspace root but hidden from the tool by pnpm's strict isolation, so `ts-node` failed to type-check. Declared both as direct devDependencies.

**Cleanup of dead scaffolding** (this is a private, run-via-`ts-node` script collection — no tests, no build output, no consumers):

- Deleted `jest.config.js` (no `*.test.ts` files exist — jest never ran), `package.sh` + `.npmignore` (the tool is `private` and outside lerna/nx scope, so it's never packed/published), `ambient.d.ts` (augmented `SpawnOptions.signal`, which nothing here uses — `fetch-metadata` only calls `execSync`, and `@types/node` 20.x already declares it), and `.gitignore` (every rule was a no-op or vestigial boilerplate copied from the `@cdktn/*` packages; the root `.gitignore` already covers anything this tool produces).
- Trimmed `package.json` to the two scripts that are actually invoked (`fetch-metadata`, `generate`); removed the never-run `build`/`watch`/`test`/`package`/`dist-clean` lifecycle and the `main`/`types` fields (which pointed at a non-existent `src/index.*`).
- Fixed `tsconfig.json` `include` and the `lint-staged` glob from `src/**` to `scripts/**` so they point at the real code (both previously matched nothing); dropped the now-obsolete `files: [ambient.d.ts]` and `*.test.tsx` exclude.
- Updated the root `generate-function-bindings` script to drive the tool by workspace filter (`pnpm --filter generate-function-bindings run …`) instead of `cd … && pnpm install`, since the tool is a workspace member whose deps are installed by the root.

### Checklist


so0k:
merged via https://github.com/open-constructs/cdk-terrain/pull/273

PR #266, opened 2026-06-10
chore: add pnpm to mise

gabegorelick:
We switched to pnpm in #170. Adding it to mise means devs don't have to manually install it.

Mise could also eventually take over from corepack/nvm in CI. But that is not implemented here.

EDIT: this change also required updating lerna, nx, and knip to support pnpm v11's multi-doc lock file format.

### Related issue

Fixes # 

### Description

Add pnpm to mise so devs don't have to manually install pnpm themselves.

### Checklist

### DIFF
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -76,7 +76,6 @@
         "@types/express",
         "express",
         // TODO: triage and add to package.json — each is imported but undeclared.
-        "cdktn",
         "constructs"
       ]
     }
--- a/package.json
+++ b/package.json
@@ -50,9 +50,9 @@
     },
     "devDependencies": {
         "@eslint/js": "^9.39.4",
-        "@nx/eslint": "^20.8.2",
-        "@nx/eslint-plugin": "^20.8.2",
-        "@nx/jest": "^20.8.2",
+        "@nx/eslint": "^22.7.5",
+        "@nx/eslint-plugin": "^22.7.5",
+        "@nx/jest": "^22.7.5",
         "@swc/core": "~1.15.5",
         "@swc/helpers": "~0.5.18",
         "@swc/jest": "~0.2.38",
@@ -71,10 +71,10 @@
         "husky": ">=6",
         "infer-owner": "^1.0.4",
         "jest": "^30.3.0",
-        "knip": "^6.6.3",
-        "lerna": "^8.2.1",
+        "knip": "^6.16.1",
+        "lerna": "^9.0.7",
         "lint-staged": "^15.5.2",
-        "nx": "20.8.2",
+        "nx": "22.7.5",
         "prettier": "^3.8.1",
         "publib": "^0.2.1047",
         "semver": "^7.7.4",
@@ -87,5 +87,12 @@
         "!(.claude|.agents|.specledger|specledger)/**/*.{ts,tsx,js,css,md}": "prettier --write",
         "examples/**": "pnpm run lint:examples"
     },
-    "packageManager": "pnpm@11.5.2+sha512.71c631e382066efc25625d5cf029075de07b61b37f6e27350fbd84b1bda5864c8c1967adc280776b45c30a715c0359a3be08fef42d5bb09e2b99029979692916"
+    "packageManager": "pnpm@11.5.2+sha512.71c631e382066efc25625d5cf029075de07b61b37f6e27350fbd84b1bda5864c8c1967adc280776b45c30a715c0359a3be08fef42d5bb09e2b99029979692916",
+    "devEngines": {
+        "packageManager":{
+            "name": "pnpm",
+            "version": "11.5.2",
+            "onFail": "download"
+        }
+    }
 }
--- a/package.json
+++ b/package.json
@@ -50,9 +50,9 @@
     },
     "devDependencies": {
         "@eslint/js": "^9.39.4",
-        "@nx/eslint": "^22.7.5",
-        "@nx/eslint-plugin": "^22.7.5",
-        "@nx/jest": "^22.7.5",
+        "@nx/eslint": "22.7.5",
+        "@nx/eslint-plugin": "22.7.5",
+        "@nx/jest": "22.7.5",
         "@swc/core": "~1.15.5",
         "@swc/helpers": "~0.5.18",
         "@swc/jest": "~0.2.38",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -214,13 +214,13 @@ importers:
         specifier: ^9.39.4
         version: 9.39.4
       '@nx/eslint':
-        specifier: ^22.7.5
+        specifier: 22.7.5
         version: 22.7.5(@babel/traverse@7.29.0)(@nx/jest@22.7.5(@babel/traverse@7.29.0)(@swc/core@1.15.40(@swc/helpers@0.5.21))(@types/node@20.19.37)(babel-plugin-macros@3.1.0)(nx@22.7.5(@swc/core@1.15.40(@swc/helpers@0.5.21)))(ts-node@10.9.2(@swc/core@1.15.40(@swc/helpers@0.5.21))(@types/node@20.19.37)(typescript@5.4.5))(typescript@5.4.5)(verdaccio@6.6.0(encoding@0.1.13)(typanion@3.14.0)))(@swc/core@1.15.40(@swc/helpers@0.5.21))(@zkochan/js-yaml@0.0.7)(eslint@9.39.4(jiti@2.7.0))(nx@22.7.5(@swc/core@1.15.40(@swc/helpers@0.5.21)))(verdaccio@6.6.0(encoding@0.1.13)(typanion@3.14.0))
       '@nx/eslint-plugin':
-        specifier: ^22.7.5
+        specifier: 22.7.5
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -87,5 +87,12 @@
         "!(.claude|.agents|.specledger|specledger)/**/*.{ts,tsx,js,css,md}": "prettier --write",
         "examples/**": "pnpm run lint:examples"
     },
-    "packageManager": "pnpm@11.5.2+sha512.71c631e382066efc25625d5cf029075de07b61b37f6e27350fbd84b1bda5864c8c1967adc280776b45c30a715c0359a3be08fef42d5bb09e2b99029979692916"
+    "packageManager": "pnpm@11.5.2+sha512.71c631e382066efc25625d5cf029075de07b61b37f6e27350fbd84b1bda5864c8c1967adc280776b45c30a715c0359a3be08fef42d5bb09e2b99029979692916",
+    "devEngines": {
+        "packageManager":{
+            "name": "pnpm",
```
gabegorelick, line 93: devEngines.packageManager is what mise uses. It's a newer standard used by pnpm, npm, and lots of other tools to replace corepack's `packageManager` field.

We need to keep the `packageManager` field for corepack for now. But if/when we get rid of corepack (which we should do at some point since it's quasi deprecated), then we can drop the redundant `packageManager` field.

package.json
```diff
@@ -87,5 +87,12 @@
...
+            "version": "11.5.2",
```
gabegorelick, line 94: Notice that there is no hash here. The hash is encoded by consumers like pnpm and mise, i.e. the hash lives in pnpm-lock.yaml and mise's lock file.

Note that we don't yet use a lock file in mise. But that can be added in a followup.

pnpm-lock.yaml
```diff
@@ -1,3 +1,202 @@
+---
+lockfileVersion: '9.0'
```
gabegorelick, line 2: pnpm v11 uses multiple yaml documents to track devEngines tools.

gabegorelick:
Looks like we need to update a bunch of tooling to support pnpm v11's new multi-doc lock file format.

package.json
```diff
@@ -50,9 +50,9 @@
     },
     "devDependencies": {
         "@eslint/js": "^9.39.4",
-        "@nx/eslint": "^20.8.2",
-        "@nx/eslint-plugin": "^20.8.2",
-        "@nx/jest": "^20.8.2",
+        "@nx/eslint": "^22.7.5",
```
gabegorelick: Upgrading nx for multi-document pnpm-lock.yaml support. See https://github.com/nrwl/nx/pull/35271.

X-Guardian: The nx version need to match the version used by lerna. Enjoy the rabbit hole...

gabegorelick: Yeah I am realizing that now...

The alternative is to drop the devEngines changes that make the pnpm lock file multi-doc. The only downside is then we'd duplicate the pnpm version in mise.toml, which isn't the end of the world.

gabegorelick: @X-Guardian as part of the pnpm migration did you consider dropping lerna and just using `pnpm --filter`?

tools/documentation-generation/project.json
```diff
@@ -2,5 +2,12 @@
   "$schema": "../../node_modules/nx/schemas/project-schema.json",
   "name": "@tools/documentation-generation",
   "root": "tools/documentation-generation",
-  "projectType": "library"
+  "projectType": "library",
+  "targets": {
+    "lint": {
+      "options": {
+        "no-error-on-unmatched-pattern": true
```
gabegorelick, line 9: eslint is failing on `tools/documentation-generation` because there are no .ts files. Apparently newer versions of lerna/nx pass through the eslint error instead of swallowing it.

gabegorelick:
I updated lerna, nx, and knip to support the new lockfile. Let me know if we'd prefer that be a separate PR.

knip.jsonc
```diff
@@ -76,7 +76,6 @@
         "@types/express",
         "express",
         // TODO: triage and add to package.json — each is imported but undeclared.
-        "cdktn",
```
gabegorelick, line 79: Newer version of knip wants us to remove this.

jsteinich:
> I updated lerna, nx, and knip to support the new lockfile. Let me know if we'd prefer that be a separate PR.

This PR is still pretty small, so I don't see a need to split them out. Can close out #76 as well.



Going to let @X-Guardian take a peek and answer some questions before myself or @so0k give a final review.

package.json
X-Guardian: For info, corepack is not deprecated. It is just not being bundled with Node from v25+

package.json
X-Guardian: I tried to keep the pnpm migration to the minimum changes required, but I'm sure we can replace some uses of lerna now with `pnpm --filter`. In my view, this would be a good move so that we can eventually migrate away from lerna completely.

package.json
gabegorelick: I suppose technically that's true, but what is the value prop of corepack if you need to manually install it? At that point, just install a package manger (pnpm or npm) and use that to bootstrap everything else.

That's why I called it "quasi" deprecated.

package.json
gabegorelick: I got the lerna upgrade working so no need to drop it now. But good to know there's no objection to going down that road in the future.

tools/documentation-generation/project.json
```diff
@@ -2,5 +2,12 @@
...
+      }
+    }
+  }
```
X-Guardian, line 12: This isn't needed in this PR. The `documentation-generation` tool is currently not linted as it is a stand-alone project in the repo. When @so0k updates the tool to be a workspace within the repo, he can make sure the linting is correct then.

package.json
```diff
@@ -50,9 +50,9 @@
...
+        "@nx/eslint-plugin": "^22.7.5",
+        "@nx/jest": "^22.7.5",
```
X-Guardian:
```suggestion
        "@nx/eslint": "22.7.5",
        "@nx/eslint-plugin": "22.7.5",
        "@nx/jest": "22.7.5",
```

These need to be pinned to exact version to match the Nx version

X-Guardian (dismissed):

tools/documentation-generation/project.json
gabegorelick: As I said in https://github.com/open-constructs/cdk-terrain/pull/266/changes#r3392027681, eslint was failing without this change.

>  The documentation-generation tool is currently not linted as it is a stand-alone project in the repo.

Did that change with the pnpm migration? It definitely was being run by nx/lerna previously, which is why I made this change.

package.json
gabegorelick: Happy to make that change, but we weren't pinning them before. And they will be pinned by the lockfile anyway.

tools/documentation-generation/project.json
gabegorelick: You can see here that nx is running it for the lint job:

Image

package.json
X-Guardian: It has the same value it always had, just with an extra one-time install. The correct package manager version is automatically used for the repo, and developers can't accidentally use the wrong package manager.

package.json
X-Guardian: Nx best practice is to pin these to explicit, matching versions so they don't get out-of-sync with the Nx version.

package.json
gabegorelick: > The correct package manager version is automatically used for the repo, and developers can't accidentally use the wrong package manager.

So the workflow is:
1. Download corepack
2. Run corepack to install pnpm

instead of
1. Download pnpm
2. Run pnpm to install pnpm

I'm not sure how that's better. pnpm will also automatically use the correct version of itself based on what's in `devEngines`. Except devEngines is much more flexible than `packageManager` (it accepts semver ranges for instance). 

Don't get me wrong, I think corepack was great when it shipped with Node as a way to bootstrap a package manager. But now that you can just as easily bootstrap with pnpm or mise or whatever, I think corepack is now just a very feature poor package manager. Unless there's something it enables that I'm missing?

so0k: ~i think what corepack offers on top is that it uses `.packageManager` field from `package.json` and thus supports package managers migrations.. but we don't regularly migrate packager manager and~ mise can do what corepack does on its own (no need for extra corepack if it's not shipped with Node anymore)

EDIT: just noticed `mise` can do the same

package.json
so0k: is pnpm on its own sufficient or does something like TurboRepo offer benefits (other than remote caches that could speed up CI/CD.. and avoid re-running jobs?

EDIT: Resolving conversation - that discussion can be on another PR, in slack or through an issue

tools/documentation-generation/project.json
so0k: > This isn't needed in this PR. The `documentation-generation` tool is currently not linted as it is a stand-alone project in the repo. When @so0k updates the tool to be a workspace within the repo, he can make sure the linting is correct then.

this has merged, I rebase this PR

so0k: oops, just realised I confused `documentation-generation` and `generate-function-bindings`, I'll resolve this and revisit when I work on docs gen

so0k (approved):
LGTM

package.json
so0k: NOTE: `devEngines` caused jsii-publib to trip up`EBADDEVENGINES` see comments in release-publish.yml

https://github.com/open-constructs/cdk-terrain/blob/48ab91f7fa268a9c15daa2d041fe30e31b6015bc/.github/workflows/release-publish.yml#L63-L71

- can't use `pnpm` to publish here because jsii-publib wraps `npm` under the hood, `npm` then throws
- python publish worked because it runs twine directly instead of jsii-publib

mise prefers `devEngines` but falls back to `packageManager` so reverted this on main to fix releases

PR #267, opened 2026-06-10
fix(cli): don't mistake terraform error output for a missing-variable prompt

### LINKED ISSUE #265: cli: cdktn deploy fails with missing variable
### Expected Behavior

When running `cdktn deploy --skip-synth` it should at least start the plan phase.

### Actual Behavior

The command exits with return code `1` and informing there was a missing variable, even though the variable has a default value as `null` in the module.

```
Missing variable: 'mirror'. You can provide it using the 'TF_VAR_mirror' environment variable.
```

### Steps to Reproduce

```
$ mkdir example
$ cd example
```

Then setup the project as described in the gist attached to this issue. Once you complete that you can build and run the docker container mounting the local `example` folder inside the container.

```bash
$ docker build --platform linux/amd64 -t cdktn-issue:latest .
$ docker run --rm -it --mount type=bind,src=$(pwd),dst=/home/nonroot/example cdktn-issue:latest /bin/bash -i
```

Then inside the container you can run:

```bash
$ poetry install
$ cdktn get
$ cdktn synth
$ cdktn deploy --skip-synth
```

### Versions

Unfortunately when I run `cdktn debug` in this example project I get the following error:

```bash
$ cdktn debug
External Error: Could not determine cdktf/cdktn version. Please make sure you are in a directory containing a cdktn project and have all dependencies installed.
```

What I can tell is:

```bash
$ cdktn --version
0.23.3

$ terraform --version
Terraform v1.14.5
on linux_amd64

$ uname -a
Linux 6d7f6f4467d7 6.8.0-47-generic #47-Ubuntu SMP PREEMPT_DYNAMIC Fri Sep 27 22:03:50 UTC 2024 x86_64 GNU/Linux
```

### Migration Context

I used cdktf and did NOT see this issue there (possible regression)

### Upstream cdktf Issue

_No response_

### Providers

Unfortunately, when I run `cdktn provider list` I get an error:

```bash
$ cdktn provider list
External Error: Could not determine cdktf/cdktn version. Please make sure you are in a directory containing a cdktn project and have all dependencies installed.
```

### Gist

https://gist.github.com/egonbraun/45aa79d25d5386892859593df5f574c1

### Possible Solutions

_No response_

### Workarounds

Right now I can confirm this issue does not happen on cdktn version `0.22.0`.

### Anything Else?

_No response_

### References

_No response_

### Help Wanted

### Community Note

so0k:
Fixes #265

## Problem

`cdktn deploy` (and `destroy`) transformed real terraform errors into a misleading `Missing variable` error, swallowing the actual failure.

In `packages/@cdktn/cli-core/src/lib/models/deploy-machine.ts`, `handleLineReceived` detected terraform's interactive variable prompt with:

```ts
} else if (noColorLine.includes("var.")) {
```

That substring check fires on **any** terraform output that references a module variable — including `Unsupported argument` error lines such as `92:   mirror = var.mirror`. As a result, the real error was hidden behind:

```
Missing variable: 'mirror'. You can provide it using the 'TF_VAR_mirror' environment variable.
```

(see #265 — this was a regression vs. `0.22.0`).

## Fix

Scope the detection to terraform's actual interactive prompt, which puts the variable name on its own line:

```
var.<name>
  Enter a value:
```

- New `isMissingVariablePrompt()` helper backed by `/^\s*var\.([A-Za-z_][A-Za-z0-9_-]*)\s*$/` replaces the loose `includes("var.")` check.
- `extractVariableNameFromPrompt` reuses the same regex and returns the captured group instead of `split("var.")`.

Now genuine missing-variable prompts are still detected, but error output that merely references a variable is forwarded to the user unchanged.

## Before / After

**Before** (`0.23.3`):
```
Missing variable: 'mirror'. You can provide it using the 'TF_VAR_mirror' environment variable.
```

**After**:
```
╷
│ Error: Unsupported argument
│   on assets/.../main.tf line 92, in resource "gitlab_project" "project":
│   92:   mirror = var.mirror
│ An argument named "mirror" is not expected here.
╵
```

## Tests

- Added regression test `should not treat terraform error output referencing var. as a missing variable (#265)` feeding the exact `Unsupported argument` block from the issue, asserting no `VARIABLE_MISSING` event fires and the real error is still forwarded.
- Existing `extractVariableNameFromPrompt` cases (`var.content`, `var.SCREAM_CASE`, `var.nested_content_578CD0EA`, `var.with-dashes`) and the `var.my_var` prompt test still pass under the new regex.

## Note

As the issue diagnosis points out, there is a second, **separate** problem: the reporter's module references `mirror`, which doesn't exist on `gitlabhq/gitlab ~> 19.0.0`. That's a module/provider version mismatch, not a cdktn bug — this PR simply makes cdktn surface that real terraform error instead of masking it.

https://claude.ai/code/session_01HYP7w7BmQM9F8xZTc9qoHD

---
_Generated by [Claude Code](https://claude.ai/code/session_01HYP7w7BmQM9F8xZTc9qoHD)_

### DIFF
--- a/packages/@cdktn/cli-core/src/lib/models/deploy-machine.ts
+++ b/packages/@cdktn/cli-core/src/lib/models/deploy-machine.ts
@@ -84,16 +84,43 @@ export type DeployState =
       context: DeployContext;
     };
 
+/**
+ * Terraform's interactive prompt for a missing variable looks like:
+ *
+ *   var.<name>
+ *     Enter a value:
+ *
+ * The variable name appears on its own line, prefixed with "var." (optionally
+ * indented). We match this precisely so we don't mistake terraform error output
+ * that merely references a variable (e.g. "92:   mirror = var.mirror" from an
+ * "Unsupported argument" error) for an actual prompt. Terraform variable names
+ * start with a letter or underscore and may contain letters, digits,
+ * underscores and dashes. (#265)
+ */
+const VARIABLE_PROMPT_REGEX = /^\s*var\.([A-Za-z_][A-Za-z0-9_-]*)\s*$/;
+
+/**
+ * Returns true if any line in the given (ANSI-stripped) output is terraform's
+ * interactive prompt for a missing variable.
+ */
+export function isMissingVariablePrompt(noColorLine: string): boolean {
+  return noColorLine
+    .split("\n")
+    .some((line) => VARIABLE_PROMPT_REGEX.test(line));
+}
+
 export function extractVariableNameFromPrompt(line: string) {
   const noColorLine = stripAnsi(line);
   const lines = noColorLine.split("\n");
-  const lineWithVar = lines.find((line) => line.includes("var."));
-  if (!lineWithVar) {
+  const match = lines
+    .map((l) => l.match(VARIABLE_PROMPT_REGEX))
+    .find((m): m is RegExpMatchArray => m !== null);
+  if (!match) {
     throw Errors.Internal(
       `Could not find variable name in prompt: ${line}. This is most likely a bug in cdktn. Please report it at http://cdktn.io/issues`,
     );
   }
-  return lineWithVar.split("var.")[1].trim();
+  return match[1];
 }
 
 interface BufferedReceiverFunction {
@@ -146,7 +173,7 @@ export function handleLineReceived(send: (event: DeployEvent) => void) {
       hideOutput = true;
       send({ type: "OUTPUT_RECEIVED", output });
       send({ type: "REQUEST_APPROVAL" });
-    } else if (noColorLine.includes("var.")) {
+    } else if (isMissingVariablePrompt(noColorLine)) {
       hideOutput = true;
 
       const variableName = extractVariableNameFromPrompt(output);
--- a/packages/@cdktn/cli-core/src/test/models/deploy-machine.test.ts
+++ b/packages/@cdktn/cli-core/src/test/models/deploy-machine.test.ts
@@ -303,6 +303,27 @@ describe("handleLineReceived", () => {
     });
   });
 
+  it("should not treat terraform error output referencing var. as a missing variable (#265)", () => {
+    const send = jest.fn();
+    const input = `
+│ Error: Unsupported argument
+│
+│   on .terraform/modules/project/main.tf line 92, in resource "gitlab_project" "project":
+│   92:   mirror                              = var.mirror
+│
+│ An argument named "mirror" is not expected here.
+`;
+    handleLineReceived(send)(input);
+    expect(send).not.toHaveBeenCalledWith(
+      expect.objectContaining({ type: "VARIABLE_MISSING" }),
+    );
+    // the real error output must still be forwarded to the user
+    expect(send).toHaveBeenCalledWith({
+      type: "OUTPUT_RECEIVED",
+      output: input,
+    });
+  });
+
   it("should detect destroy all resources in workspace message on windows", () => {
     const send = jest.fn();
     const input = `\u001b[25


sakul-learning (commented):
## How the regression was introduced

@jsteinich — this intersects with your PR #55 (Replace node-pty with simple interactive process). The fix here is correct, but I wanted to flag the full history so you can verify the regex approach aligns with your intent for that refactor.

### Timeline of the `var.` check

**1. Upstream original** (commit `1c8bc5446`, 2023-01-03 — Daniel Schmidt):
```ts
} else if (
  noColorLine.includes("var.") &&
  noColorLine.includes("Enter a value:")
) {
  hideLine = true;
  const variableName = extractVariableNameFromPrompt(line);
  send({ type: "LINE_RECEIVED", line: missingVariable(variableName) });
  send({ type: "VARIABLE_MISSING", variableName });
}
```
Both conditions in the same `if` — correct, but assumed both strings appear in the same line.

**2. bufferUnterminatedLines introduced** (upstream `2060f2f99` — #2827):
```
var.mirror
  Enter a value:
```
After this, terraform's prompt arrives as **two separate calls** to `handleLineReceived`. ~The combined `&&` check was now dead — `var.` and `Enter a value:` were never in the same line anymore.~ This sat silently because in practice, a missing-variable prompt still failed at the `var.` line (terraform exits), so nobody noticed the detection was broken.

**3. PR #55** (commit `46ab9fb86`, 2026-04-01 — @jsteinich):
```diff
-    } else if (
-      noColorLine.includes("var.") &&
-      noColorLine.includes("Enter a value:")
-    ) {
+    } else if (noColorLine.includes("var.")) {
       hideOutput = true;
       const variableName = extractVariableNameFromPrompt(output);
       send({ type: "OUTPUT_RECEIVED", output: missingVariable(variableName) });
       send({ type: "VARIABLE_MISSING", variableName });
+    } else if (noColorLine.includes("Enter a value:")) {
+      // This comes along with above block, but is a separate line
+      hideOutput = true;
     }
```
The split correctly recognized the two-line reality, but `VARIABLE_MISSING` was kept on the `var.` side — so it fires on **any** line containing `var.`, including terraform error output like `│   92:   mirror = var.mirror`.

### This PR's fix

The regex `/^\s*var\.([A-Za-z_][A-Za-z0-9_-]*)\s*$/` matches **only** standalone `var.<name>` prompt lines. It correctly handles the two-line reality without false-matching error output — arguably better than the original upstream code, which was already broken by `bufferUnterminatedLines` before PR #55 touched it.

### Verification

Reproduced with the exact Gist from #265. Before: `Missing variable: 'mirror'`. After: real `Unsupported argument` errors forwarded correctly. All 17 test suites pass (cli-core + cdktn-cli), including the new regression test.

so0k:
@sakul-learning thanks for the thorough history — the regression timeline is spot on. One correction on the dead-code call so we don't remove a load-bearing branch:

> ### Dead code
> After that split, the `Enter a value:` handler (lines 158-160 in the current base) is **dead code**:
> ```ts
> } else if (noColorLine.includes("Enter a value:")) {
>   // This comes along with above block, but is a separate line
>   hideOutput = true;
> }
> ```
> - For **genuine** missing-variable prompts: the `var.` block already sent `VARIABLE_MISSING` → state machine transitions to `exited` → this block is unreachable.
> - For **false matches** (error output like this issue): the `var.` block fires and sends `VARIABLE_MISSING` → this block never runs.

Both bullets are correct, but they only cover the **missing-variable** path. `Enter a value:` also shows up in the **approval** and **sentinel-override** flows, and that's where this branch does its work.

Since PR #55 switched to `stdio: "pipe"` (no TTY echo), terraform prints `  Enter a value: ` **unterminated**. `bufferUnterminatedLines` holds it until terraform's *next* output line arrives, then flushes it — at which point the state machine is back in `processing` (not `exited`), so the branch fires and hides the stray prompt line. In the variable path it never fires precisely because that line stays buffered and the machine exits first.

You can confirm on your machine with this standalone repro (no build needed — it inlines `bufferUnterminatedLines` + the branch ladder, including this PR's regex):

```js
// node deadcode-check.mjs
import { EOL } from "os";
function bufferUnterminatedLines(handler){let b="";function f(o){b+=o;const l=b.split(/[\n\r]+/);b=l.pop()||"";if(l.length>0)handler(l.join(EOL)+EOL);}f.getBuffer=()=>b;return f;}
const RX=/^\s*var\.([A-Za-z_][A-Za-z0-9_-]*)\s*$/;
const isVar=s=>s.split("\n").some(l=>RX.test(l));
function handle(o,log){let hide=false;const nc=o;
 if(nc.includes("Do you want to perform these actions")){hide=true;log.push(["REQUEST_APPROVAL(hide)",JSON.stringify(o)]);}
 else if(isVar(nc)){hide=true;log.push(["VARIABLE_MISSING(hide)",JSON.stringify(o)]);}
 else if(nc.includes("Enter a value:")){hide=true;log.push(["ENTER_A_VALUE_BRANCH(hide)",JSON.stringify(o)]);}
 if(!hide)log.push(["FORWARD",JSON.stringify(o)]);}
function run(name,chunks){console.log("\n=== "+name+" ===");const log=[];const r=bufferUnterminatedLines(o=>handle(o,log));for(const c of chunks)r(c);log.forEach(([k,v])=>console.log("  "+k+": "+v));console.log("  buffer left:",JSON.stringify(r.getBuffer()));}

// approval, stdio:pipe (no TTY echo): prompt is unterminated, flushed by terraform's next line
run("approval flow",[
 'Do you want to perform these actions in workspace "x"?\n  Only \'yes\' will be accepted to approve.\n\n  Enter a value: ',
 '\n\nApply complete! Resources: 1 added.\n',
]);

// genuine missing variable: Enter-a-value line stays buffered, branch never reached
run("missing variable",[ 'var.mirror\n', '  Enter a value: ' ]);
```

Output:

```
=== approval flow ===
  REQUEST_APPROVAL(hide): "Do you want to perform these actions in workspace \"x\"?\n  Only 'yes' will be accepted to approve.\n"
  ENTER_A_VALUE_BRANCH(hide): "  Enter a value: \nApply complete! Resources: 1 added.\n"
  buffer left: ""

=== missing variable ===
  VARIABLE_MISSING(hide): "var.mirror\n"
  buffer left: "  Enter a value: "
```

So in the approval (and likewise sentinel-override) flow the branch is reached and hides the `Enter a value:` line; removing it would leak that prompt into user-facing output on every interactive apply/destroy. In the variable flow it correctly never fires (`buffer left: "  Enter a value: "`). Net: not dead code — I'd leave it in place. Happy to be corrected if you see a path where the approval-flow flush doesn't happen on your end.

---
_Generated by [Claude Code](https://claude.ai/code)_

sakul-learning:
> Is the `Enter a value:` Branch Dead Code?

I was wrong (edited dead code paragraph out). 

Empirically tested with terraform v1.7.5 on Linux: `Enter a value: ` arrives **unterminated** (no trailing `\r\n`), so `bufferUnterminatedLines` holds it in the buffer. It gets flushed later when post-approval apply output arrives, at which point the branch **does fire** — the merged string matches `Enter a value:` and `hideOutput = true` suppresses it.

So the branch fires at a slightly different time than expected (during the next PTY output flush, not during the initial prompt processing). My dead-code claim was incorrect — retracted.

jsteinich (approved):

PR #268, opened 2026-06-11
feat(lib): validate Fn usage against Terraform/OpenTofu versions

so0k:
## Summary

> Stacked on #269 (declared `targetVersions` foundation). 

Terraform and OpenTofu have diverged in their built-in function sets, and the checked-in function metadata predated Terraform 1.8 — so `Fn` was missing every function added since, and using a function your targeted runtimes don't support only failed at `terraform plan` time. This PR fixes the bindings generator, regenerates the bindings to cover **both products**, and adds an opt-in synth-time validation of function usage against the project's declared targets.

Full Source data: [open-constructs/cdktn-planning - RFCS/03-function-availability/PROPOSAL.md](https://github.com/open-constructs/cdktn-planning/blob/f4fbc15510ab83d1f2378a04bcb1f9ef7908c89a/RFCS/03-function-availability/PROPOSAL.md)

Full function matrix: https://cdktn.io/function-matrix.html

### Commits (reviewable independently)

1. **chore(lib): function availability dataset + tooling** — swept `metadata functions -json` across every stable Terraform (1.5.7→1.15.6, 66 releases) and OpenTofu (1.6.0→1.12.1, 54 releases) release into `functions-matrix.json`. Includes a `pnpm update-function-matrix` delta updater for new releases (downloads only unseen versions) and the generator for the availability map. Baseline sweep scripts are marked linguist-generated and excluded from copywrite/prettier/eslint.
1. **feat(lib): regenerate function bindings from Terraform 1.15.6** — filters `core::` aliases (terraform ≥1.8 reports every builtin twice; not valid identifiers), merges OpenTofu-only functions from the matrix. New on `Fn` and the `cdktn convert` map: `convert`, `ephemeralasnull`, `issensitive`, `templatestring`, `base64gunzip`, `cidrcontains`, `urldecode`. None are variadic, so no new `INTERNAL_METHODS`. Upstream metadata also refined `format`/`formatlist` return types (`any` → `string`/`string[]`; rendered expressions unchanged).
1. **feat(lib): validate Fn usage against declared targets (`validateFunctionVersions` flag)** — a usage registry records `Fn` calls at the `terraformFunction()` chokepoint (~5ns/call, 0.04% of an `Fn` call); `ValidateFunctionVersionSupport` checks used functions from the generated availability map against the project's declared `targetVersions` via `checkFeatureSupportedByTargets` (range subset), with an availability hint pointing at the products/versions that do support the function. **Purely declarative — no binary is ever executed**; synth behaves identically in CI and environments without Terraform/OpenTofu installed. Added to `FUTURE_FLAGS` so `cdktn init` enables it for new projects.
1. **chore(tests): integration test + cross-version docs** — every synth runs with a nonexistent `TERRAFORM_BINARY_NAME`, proving no binary involvement: the default baseline targets deterministically reject `issensitive` (introduced TF 1.8 / Tofu 1.7), and raising the declared floor in `cdktf.json` makes it pass.

Example validation error:

```
Validation failed with the following errors:
  [MyStack/resource] Terraform function "templatestring" requires terraform >=1.9.0, but the project targets terraform >=1.5.7. It is available in terraform >=1.9.0 and opentofu >=1.7.0.
```

Companion docs PR: [open-constructs/cdk-terrain-docs](https://github.com/open-constructs/cdk-terrain-docs/pull/16/)

## Test plan

- [x] `pnpm nx test cdktn` — 433/434 pass (the one failure is `matchers.test.ts` `toPlanSuccessfully`, which requires a running Docker daemon; environmental, unrelated)
- [x] 8 new unit tests in `validations.test.ts`: baseline functions never resolve targets (execSync spy), default-baseline rejection, declared-targets pass (no exec), product-exclusive with availability hint, OpenTofu-only targeting, malformed declared targets, flag on/off wiring
- [x] jsii build clean (`nx run cdktn:build`)
- [x] Delta updater verified: rewinding the matrix one release and re-running reconstructs it byte-identically
- [ ] Integration test requires `pnpm package` dist (runs in CI); deterministic across the version matrix — no binary is executed


packages/@cdktn/hcl2cdk/src/function-bindings/functions.generated.ts
```diff
@@ -397,7 +439,7 @@ export const functionsMapGenerated = {
   },
   format: {
     name: "format",
-    returnType: "dynamic",
+    returnType: "string",
```
so0k, line 442: this is a fix that comes from newer terraform binary output in the `terraform metadata funcions -json` call

packages/@cdktn/hcl2cdk/src/function-bindings/functions.generated.ts
```diff
@@ -457,6 +499,15 @@ export const functionsMapGenerated = {
       },
     ],
   },
+  issensitive: {
```
so0k, line 502: resolves https://github.com/hashicorp/terraform-cdk/issues/3660

tools/generate-function-bindings/scripts/generate.ts
```diff
@@ -82,14 +83,53 @@ t.addComment(
...
+}> {
   const file = path.join(__dirname, FUNCTIONS_METADATA_FILE);
   const json = JSON.parse((await fs.readFile(file)).toString())
     .function_signatures as {
     [name: string]: FunctionSignature;
   };
 
+  const matrix: FunctionsMatrix = JSON.parse(
+    (await fs.readFile(FUNCTIONS_MATRIX_FILE)).toString(),
+  );
```
so0k, line 110: functions-matrix.json should be the single source of truth, this PR aims for a minimal diff but if you want I can re-write generate.ts to not depend on `functions.json` at all or split that off for a follow-up PR.

tools/generate-function-bindings/scripts/functions.json
so0k, line 1: the duplication here between `abs` and `core::abs` is due to a change in TF 1.8, `generate.ts` merges them and it has no impact on function binding generation aside from slight bloat in `functions.json`

packages/cdktn/src/features.ts
```diff
@@ -29,6 +29,14 @@
 export const FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS =
   "failOnConstructsOutsideOfStacks";
 
+/**
+ * When enabled, each TerraformStack validates that the Terraform functions
+ * used through `Fn` are supported by the version of the Terraform-compatible
+ * CLI (Terraform or OpenTofu) found on the system.
+ */
+export const VALIDATE_FUNCTION_VERSIONS = "validateFunctionVersions";
```
so0k, line 39: Preview docs 

- https://cdkterrain-generate-function-bindings.mintlify.app/release/function-availability
- https://cdkterrain-generate-function-bindings.mintlify.app/concepts/functions#function-availability-across-terraform-and-opentofu

sakul-learning:
I think a remaining design question is less about the function matrix itself and more about what target runtime we validate against.

The current implementation is careful — it only shells out when a constrained `Fn` is actually used, caches the detected CLI version, and downgrades a missing binary to a warning — but I’m not sure synth-time binary probing should be the default behavior. Synth often runs in CI, package builds, docs generation, or test environments where the installed Terraform/OpenTofu binary may be absent, incidental, or not representative of the runtime the project intends to support.

Could the project config declare the intended Terraform-compatible binary target and version constraints instead?

For example, conceptually, something shaped like the generated `FunctionVersionConstraints` rather than an array of version strings:

```ts
{
  terraform?: string;
  opentofu?: string;
}
```

where the values are semver constraints/ranges, e.g. `">=1.5.7"` or `">=1.6.0"`, not enumerated supported versions.

Then the validator can answer: “Are the `Fn`s used by this app supported by the project’s declared Terraform/OpenTofu target constraints?” without relying on whichever binary happens to be installed during synth.

If the project does not declare anything, the default could be the current baseline target, e.g. Terraform `1.5.7`, rather than probing the local binary. That gives existing and newly initialized projects deterministic behavior.

The split I’d suggest is a prerequisite PR:

1. First, prepare the reusable project/runtime validation foundation:
   - refactor the existing Terraform/OpenTofu feature-version validation helpers so they can validate against an injected target constraint source, not only a detected local binary;
   - add the project-level Terraform/OpenTofu target constraints to config;
   - thread that config through `cdktn init`;
   - add a feature flag / config switch for commands that should explicitly validate the installed local binary;
   - make installed-binary detection an opt-in command behavior rather than implicit synth-time behavior.

2. Then this PR can be draft/stacked on top of that prerequisite PR and focus on function availability:
   - generated function availability matrix,
   - `Fn` usage tracking,
   - validation plumbing behind `validateFunctionVersions`,
   - use the reusable target-constraint validation from the prerequisite PR.

Surfaces I’d expect the prerequisite PR to touch:

- `cdktn init`
  - `packages/cdktn-cli/src/bin/cmds/init.ts`
  - `packages/cdktn-cli/src/bin/cmds/helper/init.ts`
  - `packages/@cdktn/cli-core/src/lib/init.ts`
  - `packages/@cdktn/cli-core/templates/*/cdktf.json`

  `init` could write the default target compatibility config, or expose a flag/prompt if we want projects to choose Terraform vs OpenTofu explicitly.

- `cdktf.json` / config parsing
  - `packages/@cdktn/commons/src/config.ts`
  - `packages/@cdktn/cli-core/src/lib/cdktf-config.ts`

  This would need a typed config field for the declared target constraints, plus validation for malformed product names and malformed semver ranges.

- feature flag / command config surface
  - `packages/cdktn/src/features.ts`
  - `packages/cdktn/src/app.ts`
  - `packages/cdktn/src/terraform-stack.ts`
  - relevant `packages/cdktn-cli/src/bin/cmds/*` commands

  `validateFunctionVersions` can still gate function availability validation, but the target source should be project config/context first. A separate opt-in flag/config path can control whether a command also validates the actually installed local binary.

- validation helpers
  - `packages/cdktn/src/validations/validate-function-version-support.ts`
  - `packages/cdktn/src/validations/validate-terraform-feature-version.ts`
  - `packages/cdktn/src/functions/function-availability.generated.ts`
  - `packages/cdktn/src/functions/usage-registry.ts`

  I’d separate:
  - collecting the used constrained functions,
  - checking those functions against Terraform/OpenTofu constraints,
  - resolving the target runtime source.

  Installed-binary detection can then become one possible target resolver, not the default resolver used by synth.

I do still think local binary checks are useful, but probably only where the user explicitly asks CDKTN to interact with or verify a local Terraform-compatible CLI. For example, the opt-in validation flag/config could apply to commands such as:

- `cdktn deploy`
- `cdktn diff`
- `cdktn destroy`
- `cdktn plan` / equivalent command if present
- `cdktn get` or provider-schema flows when they actually require the CLI
- an explicit diagnostic/check command or flag, e.g. “validate against installed binary”

Important distinction: declared project constraints should not be treated as proof that the user’s installed binary satisfies them. They only state the project’s intended supported target range. Commands that actually run Terraform/OpenTofu locally can explicitly verify the installed binary against those declared constraints when that opt-in behavior is enabled.

Expected tests for the prerequisite PR:

- Unit tests:
  - config parser accepts constraint-shaped targets such as `{ terraform: ">=1.5.7", opentofu: ">=1.6.0" }`.
  - config parser rejects malformed product names and malformed semver constraints.
  - undeclared target resolves to the chosen baseline default, e.g. Terraform `1.5.7`.
  - reusable feature-version validation checks a feature against injected declared constraints without calling `execSync`.
  - Terraform and OpenTofu constraints are evaluated independently.
  - binary detection remains covered separately as an opt-in resolver.
  - command-level opt-in binary validation verifies the installed binary against declared project constraints.
  - synth/default validation does not call the binary unless explicitly configured to do so.

- Integration tests:
  - `cdktn init` writes or implies the expected target compatibility config.
  - synth succeeds/fails based on declared project constraints even when `TERRAFORM_BINARY_NAME` is absent or points to a nonexistent binary.
  - synth does not call out to Terraform/OpenTofu by default.
  - explicit CLI/runtime commands that use a local binary verify the installed binary against declared project constraints when the opt-in validation flag/config is enabled.
  - existing projects without the new field preserve the intended baseline/default behavior.

Then the stacked function-availability PR would add tests specific to this feature:

- used `Fn`s are checked against the declared Terraform/OpenTofu constraints;
- unsupported functions under the declared target constraint fail with a clear message;
- supported functions under the declared target constraint pass;
- `validateFunctionVersions` disabled means no function validation regardless of declared constraints;
- constrained function unused means no target validation work is needed.

So in short: the generated matrix and function usage tracking are useful, but I’d prefer the “what runtime am I validating against?” decision to be driven by project config prepared in a prerequisite PR. This PR can then be draft/stacked on that foundation. Installed-binary probing should probably be an explicit command/flag behavior, not the default during synth.

so0k:
Converting to draft: this PR depends on #269 - declared `targetVersions` foundation and will be rebased once that has merged.

This ensures `ValidateFunctionVersionSupport`switches from probing the installed binary at `synth` time to checking used functions against the project's **declared target ranges** via `checkFeatureSupportedByTargets`, keeping `validateFunctionVersions` as the gating feature flag. 

Installed-binary verification stays opt-in per #269.

so0k:
Rebased onto the updated #269: `ValidateFunctionVersionSupport` now validates used functions against the declared `targetVersions` (range subset) — no binary probing, no warning path. The integration test runs every synth with a nonexistent `TERRAFORM_BINARY_NAME` to prove it.

so0k:
this PR absorbs some of the work in https://github.com/open-constructs/cdk-terrain/pull/262 and I could make absorb all of it cc @X-Guardian

X-Guardian:
> this PR absorbs some of the work in #262 and I could make absorb all of it cc @X-Guardian

Up to you Vincent.

My only comment on this Pr would be why are we using Python scripts rather than Node?

so0k:
> > this PR absorbs some of the work in #262 and I could make absorb all of it cc @X-Guardian
> 
> Up to you Vincent.
> 
> My only comment on this Pr would be why are we using Python scripts rather than Node?

hahaha so hiding them in the PR view with gitattributes didn’t get them past you? :)))

they were kept only for reference on a full sweep, the actual delta is done through nodejs

i was in between a local full gitignore and not including them or converting them because I read your other PR comments.. decided it wasn’t worth the effort to rewrite (wasn’t going to re-run now anyway) and i know they worked .. and still felt they helped as a reference 

i’ll delete them.. ~I'll put them in a gist linked to this PR if anyone ever wants to rebuild the full sweep~

I've moved them all to a planning and RFC repo

PS: 
- it’s not just python, it’s bunch of bash too
- PPS: there’s tagged follow up PR work

X-Guardian:
Personally, I would merge #262 first, and then refactor your changes on top, otherwise these PRs gain more and more bloat and are then difficult to review.

so0k:
> Personally, I would merge #262 first, and then refactor your changes on top, otherwise these PRs gain more and more bloat and are then difficult to review.

Done! #262 merged, #269 rebased to main (Ready to merge) and this PR stacked on that to land the missing function support from TF 1.5.x -> TF(latest)/OTF(latest) with synth time validation

packages/cdktn/src/functions/function-availability.generated.ts
```diff
@@ -0,0 +1,36 @@
...
+ * at any version.
+ */
+export interface FunctionVersionConstraints {
+  readonly terraform?: string;
+  readonly opentofu?: string;
+}
+
+/**
+ * Terraform functions that are not universally available across all
+ * supported Terraform (>=1.5.7) and OpenTofu
```
jsteinich, line 21: We do technically support older versions still, but I don't think we need to backfill this as everything else has already existed.

packages/cdktn/src/features.ts
jsteinich: As we are only applying this to new functions, is there actually a need for a flag to disable them? Users will be able to adjust their target versions appropriately to signify which functions are valid.

tools/generate-function-bindings/scripts/generate.ts
jsteinich: Let's split that out separately

packages/cdktn/src/functions/usage-registry.ts
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) HashiCorp, Inc
+// SPDX-License-Identifier: MPL-2.0
+
+/**
+ * Records which Terraform functions have been called through the `Fn` class
+ * so that synthesis-time validations can check them against the version of
```
so0k, line 6: synth-time validations now use the cdktf.json targetVersions declaration instead of calling out to the binary.

test/Readme.md
```diff
@@ -47,3 +47,56 @@ The provider lives under `packages/@cdktn/provider-generator/lib/__tests__/edge-
...
+the unit-test matchers (`toPlanSuccessfully` and friends run
+`<binary> init/plan`).
+
+Which tests execute the binary:
+
+- **Integration tests** that call `driver.diff()` / `driver.deploy()` (e.g.
+  `typescript/variables`) run full `terraform init/plan/apply` cycles.
+  Synth-only tests (e.g. `typescript/synth-app`,
+  `typescript/function-version-validation`) execute the binary at most for a
+  `version` lookup.
```
so0k: I believe this is no longer the case

packages/cdktn/src/features.ts
so0k: For new projects, `validateFunctionVersions` will be on: `cdktn init` writes the future flags into `cdktf.json`, and the templates also declare default `targetVersions` from that PR.

For existing projects, enabling this unconditionally would be a behavior change on upgrade. A project that does not yet declare `targetVersions` would fall back to the baseline target ranges, and any use of a newer/constrained `Fn` could start failing synth until the declare `targetVersions` (even if the project is actually deployed with a newer Terraform/OpenTofu version.)

This makes `validateFunctionVersions` like the other future flag:
- new projects get it enabled;
- existing projects can opt in by adding the context flag and, ideally, explicit `targetVersions`;
- in the next major release we can remove/flip the flag and make declared-target validation the default behavior.

The important distinction is that `targetVersions` declares what runtimes the project supports, while `validateFunctionVersions` controls whether synth enforces `Fn` availability against that declaration during the current compatibility window.

tools/generate-function-bindings/scripts/generate.ts
so0k: Created https://github.com/open-constructs/cdk-terrain/issues/279 for tracking this

packages/cdktn/src/features.ts
jsteinich: As we only very recently added new functions, it's unlikely that anyone would hit a behavior change.

That said, I don't see it as worth the effort to remove since it's already been added.

jsteinich (approved):

PR #237, opened 2026-06-04
feat(cli): validate Terraform CLI feature versions

sakul-learning:
## Summary

Introduce a Terraform-compatible CLI feature/version validator that can be reused by feature PRs whose support depends on both the selected CLI product and its version.

This adds:

- `parseTerraformCliVersion()` to detect Terraform vs OpenTofu from the first line of plain `version` output:
  - `Terraform vX.Y.Z` -> `terraform`
  - `OpenTofu vX.Y.Z` -> `opentofu`
- `ValidateTerraformFeatureVersion`, a single validation class for the matrix:
  - feature -> CLI product (`terraform` / `opentofu`) -> semver constraint
- tests proving that `version -json`-style output is not used to infer the product, because OpenTofu keeps the Terraform-compatible `terraform_version` key.

## Intended downstream use

This is meant to unblock/reduce duplication for the S3 `useLockfile` work in:

- References #217
- References #234

A follow-up PR can add the backend-specific validation with:

```ts
new ValidateTerraformFeatureVersion("S3 native state locking", {
  terraform: ">=1.10.0",
  opentofu: ">=1.10.0",
});
```

If future features land at different times, the same class can encode different product constraints, for example:

```ts
new ValidateTerraformFeatureVersion("some future feature", {
  terraform: ">=1.12.0",
  opentofu: ">=1.10.0",
});
```

## Test plan

- [x] `yarn jest packages/cdktn/test/validations.test.ts --runInBand`
- [x] `yarn lerna run --scope cdktn build`
- [x] `yarn nx test cdktn --runInBand`

PR #269, opened 2026-06-11
feat: declared Terraform/OpenTofu target versions (validation foundation)

so0k:
## Summary

> Reworks unreleased https://github.com/open-constructs/cdk-terrain/issues/237 to not run terraform binary during synth

Prerequisite for #268. Instead of probing whichever Terraform/OpenTofu binary happens to be installed during synth, projects declare the runtimes they intend to support, and validations answer: *"is this feature supported by the project's declared target range?"* — deterministically, with no binary required. Installed-binary verification becomes an explicit opt-in for commands that execute the binary anyway.

```jsonc
// cdktf.json
{
  "targetVersions": {
    "terraform": ">=1.5.7",
    "opentofu": ">=1.6.0"
  },
  "validateInstalledBinary": true // opt-in, off by default
}
```

### Commits

1. **feat(cli): add targetVersions to cdktf.json config** — typed field in `@cdktn/commons`; `parseConfig` validates product names and npm-semver ranges (Terraform's `~>` syntax is rejected with a hint — npm semver would silently parse it as `~` with different semantics) and threads the declaration to apps through the existing `CDKTF_CONTEXT_JSON` channel. Also adds `parseTerraformCliVersion` to commons and the `validateInstalledBinary` flag.
2. **feat(lib): validate features against declared runtime targets** — `resolveTargetVersions(scope)` (context → dual-baseline default `terraform >=1.5.7` + `opentofu >=1.6.0`), `checkFeatureSupportedByTargets` (semver range **subset**: a feature passes only if supported across the *entire* declared range of every targeted product), and `ValidateFeatureTargetSupport` (an `IValidation` that never executes a binary). Synth-time binary probing is removed: `ValidateTerraformFeatureVersion` (merged unreleased in #237, zero call sites) is deleted together with its execSync machinery — the pure `parseTerraformCliVersion` parser and the `TerraformFeatureVersionConstraints` type remain. The long-released `ValidateBinaryVersion` and the resource-move probe built on it stay unchanged as the legacy exception, both marked `TODO(target-versions)`.
3. **feat(cli): opt-in installed-binary verification** — when `validateInstalledBinary: true`, `initalizeTerraform` (the diff/deploy/destroy chokepoint, where the binary is about to run anyway) verifies the installed product/version against the declared targets. Declared targets are deliberately *not* treated as proof about the local binary.
4. **feat(cli): write default targetVersions in init templates** — new projects get the dual baseline explicitly in `cdktf.json`.

### Semantics

- **Subset check**: declaring `terraform: ">=1.5.7"` while using a feature that needs `>=1.8.0` fails — the project claims to support 1.5.7..1.8.0 where the feature is missing. The fix is raising the declared floor (or narrowing the range), which keeps the declaration honest.
- **Missing product key in a feature's constraints** = unsupported by that product; **missing product key in declared targets** = the project doesn't target that product.
- **Undeclared** = dual product baseline (strictest, most portable interpretation).
- **Synth never executes a binary.** The only places a binary is run are commands that execute it anyway (opt-in `validateInstalledBinary`) and the legacy `ValidateBinaryVersion`/resource-move path kept for compatibility.

### Stacked work

#268 (function availability validation) is stacked on this branch: its `ValidateFunctionVersionSupport` checks used `Fn` functions against the declared targets via `checkFeatureSupportedByTargets`, gated by the `validateFunctionVersions` feature flag.

## Test plan

- [x] `@cdktn/commons`: 83/83 — parser accepts constraint-shaped targets, rejects unknown products / malformed ranges / `~>` / empty declarations / non-strings; context payload merge
- [x] `cdktn`: 425/426 — resolver default + context override + malformed-context errors; subset pass/fail; independent product evaluation; product-not-targeted skip; complex ranges; `ValidateFeatureTargetSupport` end-to-end with an execSync spy proving no binary call (the 1 failure is the pre-existing Docker-daemon-dependent `matchers.test.ts`, unrelated)
- [x] `@cdktn/cli-core`: 146 passed / 30 skipped (dist-dependent) — installed-binary check helper: satisfies/outside-range/undeclared-product/unparseable/default-baseline
- [ ] Integration: existing init tests exercise the new template field in CI

### DIFF
--- a/packages/@cdktn/cli-core/src/lib/cdktf-stack.ts
+++ b/packages/@cdktn/cli-core/src/lib/cdktf-stack.ts
@@ -3,7 +3,13 @@
 import { SynthesizedStack } from "./synth-stack";
 import { Terraform } from "./models/terraform";
 import { getConstructIdsForOutputs, NestedTerraformOutputs } from "./output";
-import { logger } from "@cdktn/commons";
+import {
+  Errors,
+  logger,
+  readConfigSync,
+  terraformBinaryName,
+} from "@cdktn/commons";
+import { checkInstalledBinaryAgainstTargets } from "./installed-binary-check";
 import { extractJsonLogIfPresent } from "./server/terraform-logs";
 import {
   TerraformCli,
@@ -250,12 +256,34 @@ export class CdktfStack {
     );
   }
 
+  /**
+   * When the project opts in via `validateInstalledBinary: true` in
+   * cdktf.json, verifies the installed Terraform-compatible binary against
+   * the project's declared `targetVersions` before executing it.
+   */
+  private async validateInstalledBinaryIfConfigured(terraform: Terraform) {
+    const config = readConfigSync();
+    if (!config.validateInstalledBinary) {
+      return;
+    }
+
+    const problems = checkInstalledBinaryAgainstTargets(
+      await terraform.version(),
+      config.targetVersions,
+      terraformBinaryName,
+    );
+    if (problems.length > 0) {
+      throw Errors.Usage(problems.join("\n"));
+    }
+  }
+
   public async initalizeTerraform(
     noColor?: boolean,
     skipProviderLock?: boolean,
     migrateState?: boolean,
   ) {
     const terraform = await this.terraformClient();
+    await this.validateInstalledBinaryIfConfigured(terraform);
     const needsLockfileUpdate = skipProviderLock
       ? false
       : await this.checkNeedsLockfileUpdate();
--- a/packages/cdktn/src/synthesize/synthesizer.ts
+++ b/packages/cdktn/src/synthesize/synthesizer.ts
@@ -30,6 +30,10 @@ export class StackSynthesizer implements IStackSynthesizer {
     invokeAspects(this.stack);
 
     if (this.stack.hasResourceMove()) {
+      // TODO(target-versions): this probes the locally installed binary during
+      // synth. Migrate to ValidateFeatureTargetSupport (declared cdktf.json
+      // targetVersions) and leave installed-binary verification to the opt-in
+      // validateInstalledBinary CLI behavior.
       this.stack.node.addValidation(
         new ValidateTerraformVersion(
           ">=1.5",
--- a/packages/cdktn/src/validations/target-versions.ts
+++ b/packages/cdktn/src/validations/target-versions.ts
@@ -9,8 +9,10 @@ import { TerraformFeatureVersionConstraints } from "./validate-terraform-feature
  * targets (set from the `targetVersions` field in cdktf.json by the CLI, or
  * directly via App context / CDKTF_CONTEXT_JSON for standalone synth).
  *
- * Kept in sync with TARGET_VERSIONS_CONTEXT_KEY in @cdktn/commons — the value
- * is duplicated because cdktn cannot depend on @cdktn/commons.
+ * This is the canonical definition. `@cdktn/commons` imports it from here
+ * (commons depends on cdktn; cdktn must not depend on commons), so the CLI
+ * writes and the construct library reads the exact same key. It is an
+ * internal wiring detail, deliberately not part of cdktn's public API.
  */
 export const TARGET_VERSIONS_CONTEXT_KEY = "targetVersions";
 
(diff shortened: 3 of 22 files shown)


packages/@cdktn/cli-core/src/lib/cdktf-stack.ts
```diff
@@ -250,12 +256,34 @@ export class CdktfStack {
...
+
+    const problems = checkInstalledBinaryAgainstTargets(
+      await terraform.version(),
+      config.targetVersions,
+      terraformBinaryName,
+    );
+    if (problems.length > 0) {
+      throw Errors.Usage(problems.join("\n"));
+    }
+  }
```
so0k, line 278: This is a project opt-in and not a Feature Flag because Feature flags in this repo are transitional, this isn't. 

`FUTURE_FLAGS` (per the docs in `features.ts`) exists for "capabilities from the future" - breaking behavior changes that new projects get enabled automatically via `cdktn init`, and that become unconditional defaults at the next major release, at which point the flag is deleted. The new `validateFunctionVersions` (function validation should eventually be default behavior) fits these semantics. But installed-binary probing should be explicit, opt-in behavior

> "only where the user explicitly asks CDKTN to interact with or verify a local Terraform-compatible CLI."

It's a permanent project preference, not a compatibility gate on its way to becoming default. Putting it in `FUTURE_FLAGS` would commit it to flipping on for every new project and becoming unconditional at the next major.

so0k:
~this PR absorbs some of https://github.com/open-constructs/cdk-terrain/pull/262 work (I could adopt all of it's efforts such as ambient file clean up) cc @X-Guardian~

EDIT: wrong PR

packages/cdktn/src/validations/validate-binary-version.ts
```diff
@@ -7,6 +7,13 @@ import * as semver from "semver";
 /**
  * A validation that can be applied to a construct that will error if the
  * construct is being used in an environment with a version of a binary lower than the specified version.
+ *
+ * TODO(target-versions): legacy synth-time binary probe that predates
```
jsteinich, line 11: We should make an issue to track switching these over

packages/cdktn/src/synthesize/synthesizer.ts
```diff
@@ -30,6 +30,10 @@ export class StackSynthesizer implements IStackSynthesizer {
     invokeAspects(this.stack);
 
     if (this.stack.hasResourceMove()) {
+      // TODO(target-versions): this probes the locally installed binary during
```
jsteinich, line 33: Add to tracking issue

packages/cdktn/src/validations/target-versions.ts
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Context key carrying the project's declared Terraform-compatible runtime
+ * targets (set from the `targetVersions` field in cdktf.json by the CLI, or
+ * directly via App context / CDKTF_CONTEXT_JSON for standalone synth).
+ *
+ * Kept in sync with TARGET_VERSIONS_CONTEXT_KEY in @cdktn/commons — the value
+ * is duplicated because cdktn cannot depend on @cdktn/commons.
+ */
+export const TARGET_VERSIONS_CONTEXT_KEY = "targetVersions";
```
jsteinich, line 17: We shouldn't be duplicating this. While cdktn cannot depend on @cdktn/commons, the other direction is completely valid.

jsteinich (changes requested):

packages/cdktn/src/validations/validate-binary-version.ts
so0k: https://github.com/open-constructs/cdk-terrain/issues/275

packages/cdktn/src/synthesize/synthesizer.ts
so0k: https://github.com/open-constructs/cdk-terrain/issues/275

so0k:
@jsteinich Good catch — addressed in 098e03c94.

**Duplication resolved.** `cdktn` is now the single canonical owner of the `targetVersions` vocabulary. `@cdktn/commons` no longer mirrors `TerraformTargetVersions` / `DEFAULT_TARGET_VERSIONS` / `TARGET_VERSIONS_CONTEXT_KEY` / the product list / `parseTerraformCliVersion` — it imports them from `cdktn` (commons already depends on cdktn) and re-exports for `cli-core`. The inaccurate "cannot import without pulling in the construct library" justifications are gone too.

**Kept internal, deliberately.** The shared symbols are reached via the `cdktn/lib/validations` subpath rather than re-exported from the package root, so they stay out of the public (jsii) API. The reason it doesn't need to be public: the actual validations are attached **inside cdktn base classes** (e.g. the upcoming `TerraformEphemeralResource`) that generated provider bindings *extend*. Generated code reaches the validation by **inheritance** — it never calls the validator itself, so the validator never has to cross the jsii contract into the Python/Go/Java/C# surface. Exposing a public/jsii export would commit us to a cross-language contract with no consumer today, so we've deferred it until a genuine third-party (non-cdktn) need appears. The end-to-end design — `cdktn` owning the `providerFeatureConstraints` matrix + the validators, generators carrying no version knowledge of their own — is tracked in RFC 04 (provider-feature-availability).

jsteinich (approved):

PR #273, opened 2026-06-13
fix(tools): repair generate-function-bindings tool

sakul-learning:
## Summary
- Supersedes #262 by carrying forward the existing X-Guardian `generate-function-bindings` fix history and adding a minimal follow-up fix.
- Preserves the original X-Guardian contribution commits in this branch history for author attribution, including `3e3741298 chore: fix generate-function-bindings tool`.
- Fixes the root generator command to run through the pnpm workspace instead of running a nested install.
- Avoids requiring `@cdktn/commons` as a tool package workspace dependency by using a type-only source import for `AttributeType`.

## Test Plan
- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm run generate-function-bindings`
- `git diff --exit-code -- packages/cdktn/src/functions/terraform-functions.generated.ts packages/@cdktn/hcl2cdk/src/function-bindings/functions.generated.ts`
- `pnpm --filter generate-function-bindings exec tsc --noEmit`
- `pnpm exec prettier --check package.json tools/generate-function-bindings/package.json tools/generate-function-bindings/scripts/generate.ts pnpm-lock.yaml`

## Notes
- I could not push the branch directly to `open-constructs/cdk-terrain` with the available `sakul-learning` GitHub token (`403 Permission denied`), so this PR uses `sakul-learning:chore/supersede-262-generate-function-bindings` as the head while targeting `open-constructs/cdk-terrain:main`.
- The submitted branch preserves the original commit SHAs where possible instead of cherry-picking/squashing them.

Supersedes #262.


tools/generate-function-bindings/scripts/generate.ts
```diff
@@ -5,7 +5,7 @@
 
 import fs from "fs/promises";
 import * as path from "path";
-import { AttributeType } from "@cdktn/commons";
+import type { AttributeType } from "../../../packages/@cdktn/commons/src/provider-schema";
```
so0k: 1. Root script now runs the workspace package through pnpm filtering instead of cd ... && pnpm install:

   ```json
   "generate-function-bindings": "pnpm --filter generate-function-bindings run fetch-metadata && pnpm --filter generate-function-bindings run generate"
   ```

2. Removed @cdktn/commons as a runtime workspace dependency from the tool package.

3. Changed the generator import to type-only source import:

   ```typescript
   import type { AttributeType } from "../../../packages/@cdktn/commons/src/provider-schema";
   ```
   This avoids requiring built `@cdktn/commons` output just to run the generator, while preserving the shared `AttributeType` shape.

4. Updated pnpm-lock.yaml accordingly.

so0k (approved):
LGTM, let's get this merged to unblock function availability matrix

so0k:
this resolved upstream issues:
- https://github.com/hashicorp/terraform-cdk/issues/3660

PR #277, opened 2026-06-21
chore: remove cdktf from tests

### LINKED ISSUE #8: Revisit cli provider commands integration tests
### Description

Previously the provider commands integration tests checked a branch built cli against a previously released library version. As we don't currently have any released versions and as part of the rename transition, we switched the tests to use a released version of cdktf instead.

Those tests are useful in the transition, but longer term we'll probably want to go back to the way it worked previously or test against the branch built library as well (with some version faking to still match providers)

jsteinich:
### Related issue

Fixes #8 

### Description

- Removes `cdktf` from the provider x command integration tests and instead tests using `cdktn`.
- Exact versions used and sometimes providers used needed to be updated to match actual availability
- Fixes an issue with C# package search.
- Fixes two issues with Java installed package identifications
  - Hardcoded to cdktf group
  - Regex match incorrect in the case where one value is missing (only the version portion could be missing)

### Checklist


.prettierignore
```diff
@@ -9,7 +9,7 @@ packages/@cdktn/cli-core/src/test/lib/cdktf-project.test.ts
 packages/@cdktn/cli-core/src/test/lib/terraform-logs.test.ts
 packages/@cdktn/cli-core/src/test/lib/execution-logs.test.ts
 
-packages/cdktf-cli/templates/*/cdktf.json
+packages/cdktn-clie/templates/*/cdktf.json
```
so0k:
```suggestion
packages/cdktn-cli/templates/*/cdktf.json
```

jsteinich:
I don't recommend reviewing individual commits (we'll squash them). I had tried one path and then rolled it back. I've also been fighting with making my local environment actually function.

test/csharp/provider-upgrade-command/test.ts
```diff
@@ -48,23 +44,19 @@ describe("provider upgrade command", () => {
...
       "installs pre-built provider using nuget",
       async () => {
-        await driver.exec("cdktn", [
-          "provider",
-          "add",
-          "random@=3.4.2", // this is not the latest version, but theres v2.0.52 of the pre-built provider resulting in exactly this package
-        ]);
+        await driver.exec("cdktn", ["provider", "add", "random@=3.7.2"]);
 
         expect(driver.readLocalFile("MyTerraformStack.csproj")).toContain(
```
sakul-learning, line 49: **Side note (non-blocking):** The Windows `onWindows` variant expects `Io.Cdktn.Providers.Random` versions `12.1.0` → `12.1.1` after `random@=3.7.2` → `random@=3.8.1`, but neither NuGet version exists (only `14.0.0` and `14.1.0` are published). The Linux equivalent correctly uses the published `14.0.0` → `14.1.0` pair. This is latent while the `windows_integration` lane is skipped, but should be fixed (use the same provider/version pair as POSIX, or an equivalent published mapping) before re-enabling Windows integration.

sakul-learning (approved):
**APPROVE** — clean chore, 23 files, +149/−197. All Linux CI passes; the implementation fixes for NuGet owner remapping and Gradle group/coordinate handling are correct and well-covered by the updated integration tests across C#, Go, Java, and TypeScript.

One non-blocking side note in the Windows C# upgrade fixture (see inline comment) — not a blocker since the Windows integration lane is skipped.

sakul-learning (approved):
**APPROVE** — re-review after force-push `e8163659 → b1c6f69b9`. The single new commit fixes the Windows C# upgrade fixture to use the same published `random@=3.8.1`/`14.0.0` → `random@=3.9.0`/`14.1.0` pair as POSIX, resolving the non-blocking side note from the previous review.

All configured checks pass: validations 31/31 ✓, nx build cdktn ✓, nx test cdktn 473/473 ✓.

No new issues. All prior findings addressed.

so0k (approved):

PR #280, opened 2026-06-25
fix(cli): include dev dependencies in npm version lookup

### LINKED ISSUE #278: cli: cdktn `debug`, `provider add` and `provider list` commands fail when cdktn is not in "dependencies" at root of npm workspace
### Expected Behavior

`cdktn debug`, `cdktn provider add` and `cdk provider list` should successfully execute.

### Actual Behavior

The following error is observed:

Could not determine cdktf/cdktn version. Please make sure you are in a directory containing a cdktn project and have all dependencies installed.

### Steps to Reproduce

Reproduction repository: https://github.com/forivall/cdktn-workspace-bug

1. Install
2. run `npx cdktn debug`
3. Observe the error.

### Versions

language: typescript
cdktn-cli: 0.23.3
node: v26.0.0
cdktn: 0.23.3
cdktf: null
constructs: null
jsii: null
terraform: 1.15.5
arch: arm64
os: darwin 25.5.0

### Migration Context

Not sure / don't remember

### Upstream cdktf Issue

_No response_

### Providers

N/A

### Gist

https://github.com/forivall/cdktn-workspace-bug

### Possible Solutions

_No response_

### Workarounds

with the dependencies in `devDependencies`, `NODE_ENV=dev npx cdktn debug` - this is because `npm list` behaves differently when `NODE_ENV=production` is passed, which [cdktn-cli does when `NODE_ENV` is unset](https://github.com/open-constructs/cdk-terrain/blob/main/packages/cdktn-cli/src/bin/cdktn.ts?plain=1#L3) [which then gets passed to the npm cli](https://github.com/open-constructs/cdk-terrain/blob/main/packages/@cdktn/commons/src/debug.ts?plain=1#L139).

or move the dependencies to `dependencies` at the root package.json

### Anything Else?

_No response_

### References

_No response_

### Help Wanted

### Community Note

sakul-learning:
## Summary

- Include dev dependencies when npm is used to detect the installed `cdktn` / `cdktf` package version.
- Add regression coverage for the npm package-version lookup arguments.

Fixes: #278

## Testing

- `pnpm nx test @cdktn/commons --runInBand --testFile=src/debug.test.ts`
- `pnpm nx build cdktn-cli`
- Validated the linked repro with the locally built CLI:
  - `cdktn debug`
  - `cdktn provider list`
  - `cdktn provider add random`

### DIFF
--- a/packages/@cdktn/commons/src/debug.test.ts
+++ b/packages/@cdktn/commons/src/debug.test.ts
@@ -1,0 +1,76 @@
+// Copyright (c) HashiCorp, Inc
+// SPDX-License-Identifier: MPL-2.0
+import * as fs from "fs";
+import * as path from "path";
+
+jest.mock("./terraform", () => ({
+  terraformVersion: Promise.resolve("1.7.5"),
+}));
+
+import { collectDebugInformation } from "./debug";
+import { withTempDir } from "./util";
+
+function writeInstalledPackage(packageName: string, version: string) {
+  const packageDir = path.join(process.cwd(), "node_modules", packageName);
+  fs.mkdirSync(packageDir, { recursive: true });
+  fs.writeFileSync(
+    path.join(packageDir, "package.json"),
+    JSON.stringify({ name: packageName, version }, null, 2),
+  );
+}
+
+describe("collectDebugInformation()", () => {
+  let previousNodeEnv: string | undefined;
+
+  beforeEach(() => {
+    previousNodeEnv = process.env.NODE_ENV;
+  });
+
+  afterEach(() => {
+    if (previousNodeEnv === undefined) delete process.env.NODE_ENV;
+    else process.env.NODE_ENV = previousNodeEnv;
+  });
+
+  it("detects TypeScript packages installed as npm dev dependencies in production", async () => {
+    await withTempDir("debug.test", async () => {
+      fs.writeFileSync(
+        "cdktf.json",
+        JSON.stringify({ language: "typescript" }),
+      );
+      fs.writeFileSync(
+        "package.json",
+        JSON.stringify(
+          {
+            name: "debug-dev-dependency-fixture",
+            version: "1.0.0",
+            devDependencies: {
+              cdktn: "0.23.3",
+              cdktf: "0.20.11",
+              constructs: "10.3.0",
+              jsii: "5.5.0",
+            },
+          },
+          null,
+          2,
+        ),
+      );
+
+      writeInstalledPackage("cdktn", "0.23.3");
+      writeInstalledPackage("cdktf", "0.20.11");
+      writeInstalledPackage("constructs", "10.3.0");
+      writeInstalledPackage("jsii", "5.5.0");
+
+      process.env.NODE_ENV = "production";
+
+      await expect(collectDebugInformation()).resolves.toEqual(
+        expect.objectContaining({
+          language: "typescript",
+          cdktn: "0.23.3",
+          cdktf: "0.20.11",
+          constructs: "10.3.0",
+          jsii: "5.5.0",
+        }),
+      );
+    });
+  });
+});
--- a/packages/@cdktn/commons/src/debug.ts
+++ b/packages/@cdktn/commons/src/debug.ts
@@ -135,9 +135,13 @@ async function getNpmNodeModuleVersion(
   let output;
 
   try {
-    output = await exec("npm", ["list", packageName, "--json"], {
-      env: { ...process.env },
-    });
+    output = await exec(
+      "npm",
+      ["list", packageName, "--json", "--include=dev"],
+      {
+        env: { ...process.env },
+      },
+    );
   } catch (e) {
     logger.debug(`Unable to run 'npm list ${packageName} --json': ${e}`);
     return undefined;


packages/@cdktn/commons/src/debug.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  it("includes dev dependencies when reading npm package versions", async () => {
+    mockExec.mockResolvedValue(
+      JSON.stringify({
+        dependencies: {
+          cdktn: {
+            version: "0.23.3",
+          },
+        },
+      }),
+    );
```
so0k: this mocks the response, not sure if this is very useful and not already covered in the integration tests?

forivall:
@sakul-learning  Could you also validate with the repro repo using yarn instead of npm?

sakul-learning:
@forivall I validated the repro against Yarn classic as well.

What I tested:

- Fresh clone of `forivall/cdktn-workspace-bug`
- Removed `package-lock.json`
- Installed with Yarn classic (`yarn 1.22.22`), producing `yarn.lock`

Results:

- Baseline `cdktn@0.23.3` still reproduces the original failure:
  - `npx --no-install cdktn debug`
  - exits `1`
  - `External Error: Could not determine cdktf/cdktn version...`
- Using the locally built CLI from this PR in the same Yarn-installed repro:
  - `cdktn debug` exits `0`
  - `cdktn provider list` exits `0`

So the PR fix is validated for the Yarn version-lookup repro: the `npm list ... --include=dev` change lets the CLI find `cdktn` even when the project was installed with Yarn classic and the dependency is only in `devDependencies`.

One caveat: `cdktn provider add random` gets past the version-detection problem and finds the prebuilt provider, but then fails differently because Yarn classic refuses to add a dependency at the workspace root unless the command uses `-W` / `--ignore-workspace-root-check`:

```text
Running this command will add the dependency to the workspace root rather than the workspace itself...
```

That looks like a separate Yarn/workspace behavior, not the original version-lookup bug fixed here.

Also, on the CI failures: I was able to reproduce intermittent Terraform Registry failures right now, even though the HashiCorp status page reports "fully operational". The failed integration logs show Terraform timing out while querying provider download metadata, e.g.:

```text
could not query provider registry for registry.terraform.io/hashicorp/kubernetes
Client.Timeout exceeded while awaiting headers
```

I also hit a timeout locally against the same registry endpoint, and then it responded later. That makes the current integration failures look like Terraform Registry/service instability rather than something caused by this PR.

EDIT: Confirmed [Hashicorp services incident]( https://status.hashicorp.com/incidents/01KVZVEWTKPS0WBSQ40P5JYAEK)

so0k:
> @sakul-learning Could you also validate with the repro repo using yarn instead of npm?

just curious, yarn classic is in maintenance mode and lacks quite a few supply chain attack safeguards, cdktn migrated to pnpm due to that

forivall:
Yeah, I would have anticipated it use yarn 2+

sakul-learning:
I validated the #278 repro converted to Yarn Berry with `nodeLinker: node-modules`, keeping `cdktn` and `cdktn-cli` as root `devDependencies`.

Setup:

- cloned `forivall/cdktn-workspace-bug`
- removed `package-lock.json`
- added `packageManager: "yarn@4.12.0"`
- added `.yarnrc.yml`:

```yaml
nodeLinker: node-modules
enableScripts: false
```

The package-version lookup part that this PR changes behaves as expected under `NODE_ENV=production`:

```text
npm list cdktn --json                 -> root cdktn hidden
npm list cdktn --json --include=dev   -> root cdktn 0.23.3 visible
```

So the `--include=dev` fix does address the original devDependency lookup failure in the Yarn Berry/node-modules shape too.

However, running the locally built PR CLI still fails later:

```text
Could not determine installed packages using 'yarn list --json'
Usage Error: Couldn't find a script named "list".
```

`cdktn debug` exited `7` in that Yarn Berry repro. I also checked the installed 0.23.3 CLI with `NODE_ENV=dev`, and it fails the same way, so this looks separate from the `NODE_ENV=production` / npm devDependency lookup fixed by this PR. The remaining blocker is in the provider-package listing path: `@cdktn/cli-core` detects `yarn.lock` and calls `yarn list --json`, which Yarn Berry no longer supports like Yarn classic.

so0k:
i think yarn berry support is a separate issue, should be separate PR? 

wdyt @jsteinich

packages/@cdktn/cli-core/src/test/lib/dependencies/dependency-manager.test.ts
```diff
@@ -5,6 +5,13 @@ import {
   ProviderConstraint,
   DependencyManager,
 } from "../../../lib/dependencies/dependency-manager";
+import { getLatestVersion } from "../../../lib/dependencies/registry-api";
+
+jest.mock("../../../lib/dependencies/registry-api", () => ({
+  getLatestVersion: jest.fn(),
+}));
+
+const mockedGetLatestVersion = jest.mocked(getLatestVersion);
```
jsteinich: Mocking the latest version seems completely unrelated and it's not clear why it would need changing.

jsteinich (changes requested):

sakul-learning:
I checked the test diff against the base branch before suggesting removals.

Current test changes in this PR are all PR-introduced changes:

- `packages/@cdktn/commons/src/debug.test.ts` is a new file in this PR (`+60`, did not exist on base).
- `packages/@cdktn/cli-core/src/test/lib/dependencies/dependency-manager.test.ts` existed already, but the `getLatestVersion` import/mock and `mockResolvedValue(null)` setup are new in this PR (`+9`). No pre-existing test body needs to be deleted to remove that change.

Suggested cleanup / reshape list:

1. **Remove from this PR:** the `dependency-manager.test.ts` `getLatestVersion` mock change.
   - Reason: it is unrelated to npm/debug devDependency detection. If it is needed to make that test deterministic, it deserves either a separate cleanup PR or a clear explanation in this PR.

2. **Replace, not just keep as-is:** the current `debug.test.ts` assertion that checks the exact npm argv (`--include=dev`).
   - Reason: that tests implementation details. The regression test should instead prove the behavior: debug/version detection still finds TypeScript packages when they are installed as root `devDependencies` under `NODE_ENV=production`.

3. **Preferred replacement coverage:** a behavior test around `getPackageVersion` / `collectDebugInformation` using a temp npm project where these are dev dependencies:
   - `cdktn`
   - `cdktf`
   - `constructs`
   - `jsii`

   The assertion should be on detected versions / non-null debug output, not on the exact npm flag.

4. **Do not add here:** Yarn Berry coverage or broad `provider add` / `provider list` command tests. Yarn Berry is a separate package-manager compatibility issue, and provider command integration coverage would be broader than this focused fix.

Are you okay with removing all of the unrelated/implementation-detail test changes above and replacing them with the behavior-level devDependency detection test? Or is there any piece you want to keep? If so, can you say why so we keep the test scope intentional?

jsteinich:
> i think yarn berry support is a separate issue, should be separate PR?
> 
> wdyt @jsteinich

I agree that yarn berry fixes is a separate issue.

sakul-learning:
Rebased this onto current `origin/main` and reshaped the diff based on the review feedback.

What changed:
- Removed the unrelated `dependency-manager.test.ts` / `getLatestVersion` mock change from the PR.
- Replaced the exact npm argv assertion with behavior-level coverage around `collectDebugInformation()`: a temp TypeScript npm project with `cdktn`, `cdktf`, `constructs`, and `jsii` installed as root dev dependencies is detected correctly under `NODE_ENV=production`.
- Kept Yarn Berry support out of scope as discussed.

Local validation:
- `pnpm exec nx test @cdktn/commons --runInBand --testFile=src/debug.test.ts --skip-nx-cache`
- `pnpm nx build cdktn-cli --skip-nx-cache`

so0k (approved):

jsteinich (approved):

PR #281, opened 2026-06-26
chore: Remove unused dependencies

X-Guardian:
### Description

Removes four declared dependencies with no usage. Each was verified before removal:

- **Root `package.json` devDependencies:**
  - `infer-owner` — no source references anywhere in the repo.
  - `semver` — the root declaration has no importer. The `semver` deps in `packages/cdktn`, `packages/cdktn-cli`, and `packages/@cdktn/cli-core` are used and are **left in place**; only the redundant root-level declaration is removed.
- **`test/package.json` devDependencies:**
  - `express` and `@types/express` — no usage.

`changelog-parser` is kept: it's required by `tools/extract-changelog` (invoked from `tools/release-github.sh`). `knip` can't see this because it traces dependencies through the import graph, and a shell-invoked script isn't in that graph — so it carries a `knip.jsonc` ignore entry documenting the consumer.

`knip.jsonc` is updated to remove the `ignoreDependencies` entries for the deleted deps.

### Checklist


sakul-learning:
Thanks for the cleanup here. One important release-path detail: `release-please` has **not** taken over the full release job yet; it only maintains the release PR / changelog / manifest in the current setup. The GitHub Release is still created by the existing release workflow, which still depends on the tool that imports `changelog-parser`.

Current dependency trace:

1. `.github/workflows/release-please.yml` documents that release-please runs with `skip-github-release: true`, so it does **not** create the tag or GitHub Release. Its comments say that responsibility remains with `release.yml`'s `release_github` job.
2. `.github/workflows/release.yml` has the active `release_github` job (`Release to Github`) which runs `pnpm run release-github`.
3. `package.json` maps `release-github` to `tools/release-github.sh`.
4. `tools/release-github.sh` calls `tools/extract-changelog` to build the GitHub Release body.
5. `tools/extract-changelog` does `require("changelog-parser")`.

So removing the root `changelog-parser` dependency breaks the currently configured GitHub Release creation path. If the goal is to retire that path because release-please should own GitHub releases too, this PR should also update the release ownership intentionally (for example, flip release-please off `skip-github-release` and remove/update the `release_github` job and the now-obsolete tools). Otherwise, `changelog-parser` should stay declared.

package.json
```diff
@@ -59,7 +59,6 @@
         "@types/jest": "30.0.0",
         "@types/prettier": "2.6.0",
         "@typescript-eslint/parser": "^8.58.0",
-        "changelog-parser": "^2.8.1",
```
so0k, line 62: still needed

so0k (changes requested):
either define package.json in tools and move `changelog-parser` as a dependency for the tools, or keep the root `changelog-parser` dependency

package.json
X-Guardian: Sigh, extension-less files: `tools/extract-changelog`

`changelog-parser` restored

so0k (approved):

PR #284, opened 2026-06-27
fix(cli): release terraform state lock when interrupting diff/deploy

### LINKED ISSUE #283: cli: interrupting diff/deploy leaves the terraform state lock held
### Expected Behavior

Pressing Ctrl-C while `cdktn diff` or `cdktn deploy` is running terraform should let terraform shut down gracefully and release its state lock, so the next command can acquire the lock without error.

### Actual Behavior

The interrupt kills terraform mid-operation without giving it a chance to shut down gracefully (no `Interrupt received... Gracefully shutting down...` is printed), so the state lock is left held. The next `cdktn diff`/`deploy` then fails with `Error acquiring the state lock`, and the user has to `force-unlock` manually.

Interrupting a `cdktn diff` while terraform is refreshing — terraform is killed with no graceful-shutdown output:

```
my-stack  example_resource.web: Refreshing state... [id=1234]

⠹  Processing
Exit status 1
```

Re-running immediately fails because the lock from the interrupted run was never released:

```
│ Error: Error acquiring the state lock
│
│ Error message: ConditionalCheckFailedException: The conditional request
│ failed
│ Lock Info:
│   ID:        eabbee29-11fe-42bd-d213-c76fe98cb6db
│   Path:      example-terraform-state/my-stack/terraform.tfstate
│   Operation: OperationTypePlan
│   Who:       user@host
│   Version:   1.5.5
│   Created:   2026-06-27 08:26:55 +0000 UTC
│
│ Terraform acquires a state lock to protect the state from being written
│ by multiple users at the same time. Please resolve the issue above and try
│ again. For most commands, you can disable locking with the "-lock=false"
│ flag, but this is not recommended.
```

### Steps to Reproduce

1. Run `cdktn diff` (or `cdktn deploy`) on a stack with a backend that locks state, against a slow/remote backend so the operation runs long enough to interrupt.
2. Press Ctrl-C while terraform is refreshing/applying.
3. Re-run the same command.
4. It fails to acquire the state lock.

### Versions

cdktn-cli: 0.23.3
node: v22.16.0
terraform: 1.5.5
os: darwin

### Migration Context

### Upstream cdktf Issue

_No response_

### Providers

N/A

### Gist

_No response_

### Possible Solutions

Don't re-signal terraform on abort — it already got the interrupt via the process group. Instead, wait for terraform's own exit before resolving, so the run only returns once the lock is released:

- diff/`exec`: stop forwarding spawn's `signal` option; settle on the child's own `close`.
- deploy: add an intermediate state that waits for terraform's `EXITED` before reaching `stopped`.

### Workarounds

_No response_

### Anything Else?

_No response_

### References

_No response_

### Help Wanted

### Community Note

X-Guardian:
### Related issue

Fixes #283

### Description

Interrupting `cdktn diff` or `cdktn deploy` (Ctrl-C) left terraform's state lock held, so the next run failed with `Error acquiring the state lock`.

The cause was that the CLI's abort handler tore terraform down instead of letting it exit on its own. Terraform shares the CLI's process group, so Ctrl-C already reaches it directly and starts its graceful shutdown (which releases the lock) — but the CLI killed it before that shutdown could finish. The fix is to stop tearing terraform down on abort and instead wait for it to exit:

- **diff/plan** ([`@cdktn/commons/src/util.ts`](packages/@cdktn/commons/src/util.ts)): `exec()` no longer forwards the abort to `child_process.spawn`'s `signal` option, which killed the child with SIGTERM and rejected the promise immediately (synthetic `AbortError`) before terraform had exited — so terraform never ran its graceful shutdown and the lock leaked. It now settles on the child's own `close`, so the run resolves only once terraform has exited and released the lock.

- **deploy/destroy** ([`@cdktn/cli-core/src/lib/models/deploy-machine.ts`](packages/@cdktn/cli-core/src/lib/models/deploy-machine.ts)): the xstate machine previously went straight to the final `stopped` state on `STOP`, killing the pty and resolving the run before terraform had finished shutting down. It now passes through an intermediate `stopping` state that marks the run cancelled and waits for terraform's natural `EXITED` before reaching `stopped`. The pty stays alive throughout, so terraform exits on its own and the run only resolves afterwards.

**Why no extra signal is needed:** every abort trigger in the codebase (`runCdktfProject`, `watch`) originates from `process.on(SIGINT/SIGTERM/SIGQUIT)`, so terraform — sharing the process group — always receives the signal directly. There is no abort path that requires the CLI to deliver the signal itself.

**Verification:** confirmed after the fix that a single Ctrl-C produces only `Interrupt received... Gracefully shutting down...` and terraform exits cleanly on its own. The full lock-release path (interrupt a locking operation, then re-run) still needs a final check against a backend that locks state.

### Checklist


sakul-learning:
Thanks for taking this on. I ran the PR head through our local termless `cdktn-e2e` path, including the unmerged e2e test from https://github.com/open-constructs/cdktn-e2e/pull/3, and I’m seeing the diff/http-lock case still reproduce.

Exact setup/checks I ran:

```bash
cd /data/repos/cdktn-e2e

# build/publish/install the cdk-terrain PR head into a local sandbox CLI
CDKTN_MONOREPO=/data/repos/hermes-pr-reviewer/worktrees/cdk-terrain/open-constructs__cdk-terrain/pr-284 pnpm provision cdktn-prhead

# static check for the e2e repo
pnpm typecheck

# run the relevant termless interrupt/lock tests against the provisioned PR-head CLI
CLI_ID=cdktn-prhead pnpm exec vitest run tests/diff-lock-release.test.ts tests/ctrl-c-teardown.test.ts
```

Observed results:

- `pnpm provision cdktn-prhead`: passed; installed a sandbox CLI from this PR head (`0.23.3+prhead`).
- `pnpm typecheck`: passed.
- `CLI_ID=cdktn-prhead pnpm exec vitest run tests/diff-lock-release.test.ts tests/ctrl-c-teardown.test.ts`: failed, with 2 failing tests and 3 passing tests.

The important failure is the unmerged PR #3 e2e test:

```text
FAIL tests/diff-lock-release.test.ts
Ctrl-C during diff waits for terraform plan to unlock the HTTP backend
```

That test uses a deterministic in-process Terraform HTTP backend, so it directly records the backend `LOCK` / `UNLOCK` requests. Against this PR head I saw the backend receive `lock` but never `unlock`, so it reproduces the exact `cdktn diff` / HTTP-state-lock regression this PR is intended to fix.

I also saw this existing interrupt e2e fail against the same provisioned PR-head CLI:

```text
FAIL tests/ctrl-c-teardown.test.ts
Ctrl-C at the deploy approval prompt exits cleanly and restores the cursor
```

That one hung after Ctrl-C at the Approve/Dismiss/Stop prompt (`waitExit(..., 30_000)` returned `null`).

Could you share how you validated the fix locally, and whether you tried the `cdktn-e2e` PR #3 test? It’s also possible our e2e harness is modeling the scenario incorrectly, so if you think the deterministic HTTP backend test is wrong I’d like to understand that before treating this as a PR blocker.

X-Guardian:
I have tested this locally. This is not interrupting Terraform at the approve prompt, but during the plan or apply process.

jsteinich:
@X-Guardian I think this looks good, but curious what the behavior of a 2nd Ctrl-C press is. My expectation is that it would abort gracefully shutdown.

X-Guardian:
I tested this and both Ctrl-C's get passed through to Terraform which then responds with:

```
Two interrupts received. Exiting immediately. Note that data loss may have occurred.
```

jsteinich (approved):

PR #285, opened 2026-06-28
fix(cli): resolve prebuilt Java provider versions via repo1.maven.org

X-Guardian:
### Description

The Java integration tests (`provider-add`, `provider-list`, `provider-upgrade`) have been intermittently failing, resolving a prebuilt provider package version one or two below the expected maximum (e.g. `0.2.55` → `0.2.53`/`0.2.54`).

#### Example Failed jobs

- https://github.com/open-constructs/cdk-terrain/actions/runs/28284215519/job/83805855926?pr=284
- https://github.com/open-constructs/cdk-terrain/actions/runs/28284215519/job/83805856150?pr=284

#### Cause

When adding a prebuilt Java provider, the resolver walks the npm candidate versions highest-to-lowest and installs the first that exists on Maven Central, using `JavaPackageManager.isNpmVersionAvailable` to decide existence. That check queries the `search.maven.org/solrsearch` index — a separate service from the artifact repository, and a flaky one: under load it returns `HTTP 502` and can take ~90s to respond. A failed search response reads as "version not found", so the resolver drops to an older package. 

#### Fix

This PR points the existence check at the artifact repository directly:

- **Probe `repo1.maven.org`** ([`package-manager.ts`](packages/@cdktn/cli-core/src/lib/dependencies/package-manager.ts)): `JavaPackageManager.isNpmVersionAvailable` requests the artifact's `.pom` directly (`https://repo1.maven.org/maven2/<group>/<artifact>/<version>/<artifact>-<version>.pom`) with a plain `fetch`, matching the shape of the other package managers' probes — `200` means present, anything else means not available. `repo1.maven.org` is the Fastly-fronted artifact CDN every Maven build already relies on, so it is both the source of truth for "is this artifact published" and reliably fast (~0.05s).

**Why a direct file check rather than the `central.sonatype.com` search API:** the artifact CDN is the most reliable signal and the fastest, with the fewest moving parts. The Sonatype Central search API is the documented successor to `search.maven.org`, but it is still a search index in front of the same artifacts; querying the artifact directly depends only on the artifact itself.

### Checklist


packages/@cdktn/cli-core/src/lib/dependencies/dependency-manager.ts
```diff
@@ -333,21 +333,15 @@ export class DependencyManager {
     );
 
     for (const { name, version } of prebuiltProviderNpmVersions) {
-      try {
```
jsteinich, line 336: While removing the `try` block makes sense for this particular change context, it's a bigger scope than that.

`tryGetLanguagePackageName` can throw (primarily related to network errors).

`isNpmVersionAvailable` has implementations for all the other supported package managers that also have the potential to fail.

Moving more error handling further down could result in better checks / messaging, but I think we should leave it here until other areas have been validated.

packages/@cdktn/cli-core/src/lib/dependencies/package-manager.ts
```diff
@@ -87,6 +87,142 @@ const pipPackageSchema = z.array(
   z.object({ name: z.string(), version: z.string() }).nonstrict(),
 );
 
+/**
+ * Marks failures worth retrying (network errors, 408/429, 5xx). A definitive non-transient response — e.g. a 4xx
+ * from the registry — is not wrapped and so surfaces immediately without burning retries.
+ */
+class TransientFetchError extends Error {}
```
jsteinich: Can we pull the generic url retry into a separate file? It's not really specific to this and could potentially be reused elsewhere in the codebase.

jsteinich (changes requested):

packages/@cdktn/cli-core/src/lib/dependencies/dependency-manager.ts
X-Guardian: I've reverted this code change.

packages/@cdktn/cli-core/src/lib/dependencies/package-manager.ts
X-Guardian: I've decided to remove the retry code and keep the PR to only changing the Java resolver to `repo1.maven.org`.

jsteinich (approved):

PR #292, opened 2026-06-30
fix(provider-generator): copy only publishable files into the jsii compile bundle

X-Guardian:
### Description

The `@cdktn/provider-generator:build` step (its `postbuild` runs the edge-provider-schema CLI, which generates bindings for all languages) intermittently fails in CI with:

```
[Error: ENOENT: no such file or directory, stat
 '/__w/cdk-terrain/cdk-terrain/packages/cdktn/.pack-staging/node_modules/semver/ranges/outside.js']
```

Example failed job: https://github.com/open-constructs/cdk-terrain/actions/runs/28436357153/job/84263776058?pr=285

The cause is in `generateJsiiLanguage` ([`constructs-maker.ts`](packages/@cdktn/provider-generator/src/get/constructs-maker.ts)). To build the jsii compile bundle, it copies each compile dependency into a staging `node_modules`. One of those deps is the `cdktn` package directory itself, and the copy was a blanket `fs.copy(dir, targetdir, { dereference: true })` of the **whole directory**.

`packages/cdktn/.pack-staging` is a transient tree that the `pnpm package` step (`packages/cdktn/scripts/pack.mjs`) creates and rewrites — it `rmSync`s it and runs `npm install` into it. Because `.pack-staging` is a subdirectory of `packages/cdktn`, the blanket copy recursed into it. When the two steps overlapped in CI, `fs.copy` listed a file and then `stat`ed it after `pack.mjs` had already removed it mid-install — hence the `ENOENT`. Beyond the race, copying `.pack-staging` (and `dist`) into the bundle is pure waste: they are build outputs, never jsii compile inputs.

The fix copies only each dependency's **publishable** files — the set `npm pack` would ship — instead of the whole directory:

- **Add `resolvePublishableFiles(dir)`** ([`constructs-maker.ts`](packages/@cdktn/provider-generator/src/get/constructs-maker.ts)): runs `npm pack --dry-run --json` for the dep and returns its publishable file list (dropping the `..`-escape paths npm emits when following pnpm's `node_modules` symlinks). This is the same technique already used by `packages/cdktn/scripts/pack.mjs`.

- **Copy that file list into staging** instead of the whole directory. Build outputs — `.pack-staging`, `dist`, `node_modules`, `*.tsbuildinfo` — fall away because they are not in the package's published surface, so the copy no longer walks `.pack-staging` and the race is gone.

**Verification:** rebuilt `@cdktn/provider-generator` end-to-end (`tsc` + the `postbuild` edge-CLI that was crashing) with `.pack-staging` present locally — it passes, and all four language targets (Python, Java, C#, Go) plus the TypeScript providers are generated, confirming nothing needed was dropped from the copy.

### Checklist


sakul-learning:
Review verdict: **Approve** ✅

I reviewed the PR diff and the change looks well-scoped: `generateJsiiLanguage()` now stages only the files that `npm pack --dry-run --json` says each compile dependency would publish, instead of recursively copying the entire dependency directory. That directly addresses the `.pack-staging` race described in the PR without adding broader machinery.

Checks I ran locally on PR head `700f3c572133f715fe34c17cc3b0766bc0907753`:

- `pnpm exec jest packages/cdktn/test/validations.test.ts --runInBand` — passed: 31 tests, 9 snapshots.
- `pnpm exec lerna run --scope cdktn build` — passed via Nx cache.
- `pnpm exec nx test cdktn --runInBand` — passed: 38 suites / 438 tests / 282 snapshots.
- `pnpm exec nx run @cdktn/provider-generator:build --skip-nx-cache` — passed, including dependent builds and the provider-generator `tsc` + `postbuild` path (`node ./build/__tests__/edge-provider-schema/cli.js ./edge-provider-bindings`).
- As an extra regression check, I created a temporary `packages/cdktn/.pack-staging/node_modules/semver/ranges/outside.js` sentinel and reran `pnpm exec nx run @cdktn/provider-generator:postbuild --skip-nx-cache`; it passed, then I removed the temporary sentinel.

No blocking correctness/security/test-value findings from my review.

so0k (approved):

PR #298, opened 2026-07-02
fix(cli): don't downgrade prebuilt providers on a transient registry failure

### LINKED ISSUE #297: cli: prebuilt-provider version resolution silently downgrades on a transient registry failure
### Expected Behavior

When `cdktn provider add` resolves which prebuilt-provider version to install, a *transient* failure while checking a registry (network error, rate limit, 5xx) should not be mistaken for "this version doesn't exist". Either the check should retry, or the failure should surface — so the resolver installs the newest version that genuinely matches, not an older one that happened to be checked after the network recovered.

### Actual Behavior

`DependencyManager.getLanguageSpecificPackageVersion` walks the candidate versions highest-to-lowest and returns the first that `isNpmVersionAvailable` reports as present. For most languages a transient registry failure is indistinguishable from "not available", so the resolver skips the (present) newest version and silently installs an older one. There is no error and no warning that a downgrade happened.

### Steps to Reproduce

1. `cdktn provider add` a prebuilt provider for a Go / Python / .NET target.
2. Have the availability check hit a transient failure on the newest candidate version (e.g. GitHub API rate limit for Go, a PyPI/NuGet 5xx).
3. Observe an *older* version gets installed with no indication the newest was skipped due to an infrastructure hiccup.

### Versions

cdktn-cli: 0.23.3

### Migration Context

Not sure / don't remember

### Upstream cdktf Issue

_No response_

### Providers

_No response_

### Gist

_No response_

### Possible Solutions

The availability probes conflate "transient failure" with "definitively absent", and the resolution loop's `catch` swallows anything that does throw:

- **Go** ([`package-manager.ts`](packages/@cdktn/cli-core/src/lib/dependencies/package-manager.ts) `GoPackageManager.isNpmVersionAvailable`) — GitHub returns a JSON body on rate-limit / 5xx, so `response.json()` succeeds, `json.ref` is undefined, and the method returns `false`. A throttled request reads as "version absent".
- **Python** (`PythonPackageManager`) and **.NET** (`NugetPackageManager`) — bare `fetch` with no status check; a non-JSON error body makes `response.json()` throw.
- The loop in `DependencyManager.getLanguageSpecificPackageVersion` ([`dependency-manager.ts`](packages/@cdktn/cli-core/src/lib/dependencies/dependency-manager.ts)) wraps the body in `try { … } catch (err) { logger.info("… Skipping…") }`, so any throw from the above (or from `tryGetLanguagePackageName`, which itself makes a network call) is logged at `info` and the loop moves on to an older version.

Net effect: neither a `false`-returning probe nor a throwing one can distinguish "not published" from "couldn't reach the registry", and the caller treats both as "try the next-oldest version".

Give the probes a shared, deliberate transient-vs-definitive distinction, then decide at the loop level what a transient failure should do (retry, or abort rather than downgrade):

- Introduce a small retry helper (retry only network errors / `408` / `429` / `5xx`; treat `404`/`numFound: 0`/missing-ref as definitive absence) and route each `isNpmVersionAvailable` through it.
- For Go, check `response.ok` / rate-limit headers explicitly instead of inferring absence from a missing `json.ref`.
- Narrow the resolution loop's `catch` so a genuine "registry unreachable" propagates (or is retried) rather than being logged-and-skipped, while a definitive "absent" still advances to the next candidate.
- Validate each endpoint's real failure responses (Go/GitHub, Python/PyPI, .NET/NuGet) before committing to the shared shape — the failure bodies differ per registry.

### Workarounds

_No response_

### Anything Else?

_No response_

### References

_No response_

### Help Wanted

### Community Note

X-Guardian:
### Related issue

Fixes #297

### Description

When `cdktn provider add` resolves which prebuilt-provider version to install, it walks the npm candidate versions highest-to-lowest and installs the first that `isNpmVersionAvailable` reports as present. The problem: for most languages a *transient* registry failure (network error, rate limit, 5xx) was indistinguishable from "this version doesn't exist", so the resolver skipped the present newest version and silently installed an older one — no error, no warning.

The probes conflated the two failure modes, and the resolution loop's `catch` swallowed anything that threw:

- **Go** — GitHub returns a JSON body on rate-limit / 5xx, so `response.json()` succeeded, `json.ref` was `undefined`, and the probe returned `false`. A throttled request read as "version absent".
- **Python** (PyPI) and **.NET** (NuGet) — bare `fetch` with no status check; a non-JSON error body made `response.json()` throw.
- The loop in `DependencyManager.getLanguageSpecificPackageVersion` wrapped the body in `try { … } catch { logger.info("… Skipping…") }`, so any throw was logged at `info` and the loop moved on to an older version.

This PR gives the probes a shared, deliberate transient-vs-definitive distinction and decides at the loop level what a transient failure should do:

- **Shared retry helper** ([`package-manager.ts`](packages/@cdktn/cli-core/src/lib/dependencies/package-manager.ts)): a new `fetchWithRetry` classifies transient failures — network errors and HTTP `408` / `429` / `5xx` — and retries them (3 attempts, short exponential backoff), while definitive responses (`2xx`, `404`, other `4xx`) return immediately. If the registry stays unreachable after all attempts, it throws an `External` error rather than reporting the version as absent.
- **Status-based probes**: each `isNpmVersionAvailable` now routes through `fetchWithRetry` and keys off the HTTP status instead of inferring absence from a missing field. Go checks `200` vs `404` (fixing the rate-limit-reads-as-absent bug), Python and .NET check `404` explicitly, and Java (already probing `repo1.maven.org` directly) shares the same retry behaviour.
- **Narrowed the resolution loop's `catch`** ([`dependency-manager.ts`](packages/@cdktn/cli-core/src/lib/dependencies/dependency-manager.ts)): a registry-unreachable (`External`) error now propagates and aborts the resolution, so the CLI never silently downgrades on an infrastructure hiccup. A definitive "not available" (a `false` return, no throw) still advances to the next-oldest candidate as before.

**Why abort rather than retry-forever or degrade silently:** installing an older version than is actually available is a silent correctness problem that's hard to notice after the fact. Surfacing a clear "could not reach the registry" error lets the user retry against a recovered network and get the version they expect.

### Checklist

### DIFF
--- a/packages/@cdktn/cli-core/src/lib/dependencies/dependency-manager.ts
+++ b/packages/@cdktn/cli-core/src/lib/dependencies/dependency-manager.ts
@@ -3,6 +3,7 @@
 import {
   Language,
   Errors,
+  IsErrorType,
   logger,
   TerraformDependencyConstraint,
 } from "@cdktn/commons";
@@ -345,6 +346,9 @@ export class DependencyManager {
           return { name: languagePackageName, version };
         }
       } catch (err) {
+        if (IsErrorType(err, "External")) {
+          throw err;
+        }
         logger.info(
           `Could not find version ${version} for package ${name}: '${err}'. Skipping...`,
         );
--- a/packages/@cdktn/cli-core/src/lib/dependencies/package-manager.ts
+++ b/packages/@cdktn/cli-core/src/lib/dependencies/package-manager.ts
@@ -14,8 +14,108 @@ import path from "path";
 import { xml2js, js2xml, Element } from "xml-js";
 import * as fs from "fs-extra";
 import * as semver from "semver";
-import fetch from "node-fetch";
+import fetch, { RequestInit, Response } from "node-fetch";
 import * as z from "zod";
+
+/**
+ * Number of attempts (initial try + retries) for a registry probe before we give up and treat the registry as
+ * unreachable. Transient failures (network errors / 408 / 429 / 5xx) are retried; definitive responses are not.
+ */
+const REGISTRY_PROBE_ATTEMPTS = 3;
+
+/** Base backoff between retries, doubled each attempt (250ms, 500ms). */
+const REGISTRY_PROBE_BACKOFF_MS = 250;
+
+/**
+ * HTTP 408 Request Timeout — a transient status worth retrying.
+ */
+const HTTP_REQUEST_TIMEOUT = 408;
+
+/**
+ * HTTP 429 Too Many Requests (rate limit) — a transient status worth retrying.
+ */
+const HTTP_TOO_MANY_REQUESTS = 429;
+
+/**
+ * Lowest HTTP 5xx status — any status at or above this is a server error and treated as a transient status worth
+ * retrying.
+ */
+const HTTP_SERVER_ERROR_MIN = 500;
+
+/**
+ * HTTP 404 Not Found — a definitive "the version is not published" answer from a registry (not transient).
+ */
+const HTTP_NOT_FOUND = 404;
+
+/**
+ * Reports whether an HTTP status is transient and therefore worth retrying: request timeout (408), rate limit (429),
+ * or any server error (5xx). Definitive statuses (2xx, 404, other 4xx) are not transient.
+ *
+ * @param status - The HTTP status code from a registry response.
+ * @returns `true` if the status is transient and the request should be retried, `false` otherwise.
+ */
+function isTransientStatus(status: number): boolean {
+  return (
+    status === HTTP_REQUEST_TIMEOUT ||
+    status === HTTP_TOO_MANY_REQUESTS ||
+    status >= HTTP_SERVER_ERROR_MIN
+  );
+}
+
+/**
+ * Resolves after `ms` milliseconds; used to back off between registry-probe retries.
+ *
+ * @param ms - The number of milliseconds to wait.
+ * @returns A promise that resolves once the delay has elapsed.
+ */
+function delay(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+/**
+ * Fetches `url`, retrying only on transient failures — network errors and HTTP 408/429/5xx — so a registry hiccup is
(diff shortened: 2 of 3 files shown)


so0k (approved):
Improves reliability & correctness: it stops transient registry failures from being misclassified as “version absent,” avoids silent provider downgrades, and surfaces a detailed `External` error instead.

### Concerns checked and resolved

1. **Is `fetchWithRetry` logic already pre-existing in CLI/package-manager?**  
   Resolved: no. Search only did not surface existing shared CLI/package-manager fetch retry helper to reuse.

2. **Are `promise-retry` / `@gar/promise-retry` already dependencies?**  
   Resolved: they are present only as **transitive dev/tooling dependencies**, not as direct `@cdktn/cli-core` dependencies. They come from npm/lerna/pacote tooling paths such as `lerna -> pacote`, `npm-registry-fetch`, `make-fetch-happen`, `@npmcli/git`, and `sigstore`.  
   `@cdktn/cli-core` directly depends on `node-fetch`, not these retry packages. So the local `fetchWithRetry` helper earns its keep and does not duplicate an existing project-level dependency-manager abstraction.

3. **Is a local `delay` helper duplicating an existing CLI helper?**  
   Resolved: no shared reusable CLI delay helper was found. There are isolated `setTimeout` / `sleep` patterns elsewhere, but nothing obvious in `cli-core` that this code should import. The tiny local helper is acceptable.

4. **Does the retry logic already use exponential backoff?**  
   Resolved: yes. It uses bounded exponential backoff:

   ```ts
   REGISTRY_PROBE_ATTEMPTS = 3
   REGISTRY_PROBE_BACKOFF_MS = 250
   delay = 250ms, then 500ms
   ```

   That is appropriate for this CLI path.

5. **Should `429 Retry-After` be honored?**  
   Considered a non-issue for this PR. For an interactive CLI, failing fast with a clear `External` error after bounded retries is acceptable — often preferable to silently blocking for `5s`/`60s`. Not honoring `Retry-After` does not reintroduce the correctness bug and is not worth a follow-up cleanup here.

6. **Do the tests provide meaningful coverage?**  
   Resolved: yes. The tests exercise definitive success/absence and transient-failure paths across Maven, GitHub/Go, PyPI, and NuGet. They are behavior-focused and run in the normal test path.

PR #305, opened 2026-07-04
fix(docs): fix stale constructs pin breaking with()/IMixin docs

so0k:
## Summary

- `tools/documentation-generation` pinned `constructs@10.3.0`, while every other package in the repo (`packages/cdktn`, `@cdktn/cli-core`, `@cdktn/provider-generator`, `cdktn-cli`, all examples) uses `10.6.0`. The pin dates back to the `cdktf` era (2024-05-06) and was never bumped since.
- `jsii-docgen` documents inherited `Construct` members by resolving the base-class assembly it's pointed at via `assembliesDir`. Since that assembly came from the stale `constructs@10.3.0`, which predates the `with(...mixins: IMixin[])` mixin API, none of the generated construct docs (across all 5 languages) documented `with()`/`IMixin` at all.
- **Root cause**: `tools/documentation-generation` was excluded from `pnpm-workspace.yaml` — a holdover from Yarn Classic days, when workspace hoisting could genuinely flatten `constructs` to a conflicting version. pnpm doesn't hoist, so that constraint stopped applying after the pnpm migration, but the exclusion was never revisited. Being outside the workspace made the package invisible to `lerna exec`-based tooling, including the **existing** `upgradeJSII` CI job that already keeps `constructs`/`jsii-docgen`/`jsii-rosetta` in lockstep everywhere else — which is exactly how this pin went stale in the first place.

## Changes

1. Bump `constructs` to `10.6.0` in `tools/documentation-generation/package.json`.
2. Add `tools/documentation-generation` to `pnpm-workspace.yaml`.
3. Switch its `cdktn` dependency to `workspace:*` (was a manual `link:../../packages/cdktn`).
4. Align `jsii-rosetta` to `~5.9.0` (matches `@cdktn/hcl2cdk`) — this drift was invisible before and only surfaced once `sherif`/lerna could actually see the package.
5. Drop its standalone `pnpm-lock.yaml` in favor of the root lockfile; simplify `generate-docs:api` now that a plain `pnpm install` covers this package too.

## Test plan

- [x] Ran the full `pnpm generate-docs:api` pipeline (install + build + regenerate docs) twice — once for the version-bump-only fix, once after the workspace restructuring — from a clean worktree each time
- [x] Confirmed `tools/documentation-generation/node_modules/constructs/package.json` resolves to `10.6.0` and `node_modules/cdktn` symlinks correctly via the workspace protocol
- [x] Confirmed generated docs include `with()`/`IMixin` — verified across all 5 languages' `constructs/` topic (40 construct files each, 0 before the fix)
- [x] `website/docs/cdktn/**` output is gitignored, so no generated-doc diff is part of this PR
- [x] Confirmed `@tools/documentation-generation` now shows up in `lerna ls --all` (so future `upgradeJSII` runs will enumerate it) and in `npx sherif` output (so future drift will be flagged) — verified by simulating the regression (reverting the pin) and confirming detection tooling behaves correctly both ways


jsteinich (approved):

so0k:
Thanks for review 

I will add constructs as a peer dependency of `documentation-generation`, so i don't have to list it in Knip `ignoreDependencies`. I will also remove the redundant .gitignore.
and update the PR title too.

as suggested by @X-Guardian

so0k:
Rebased onto main and addressed the follow-ups from above:

- `constructs` is now a `peerDependency` of `@tools/documentation-generation` instead of a regular dependency.
- Removed `constructs` from Knip's `ignoreDependencies` for this workspace, per the suggestion — Knip already recognizes a declared peer dependency as "used" without needing to statically import it.
- Removed the now-redundant `tools/documentation-generation/.gitignore` (the file it un-ignores, `generate-documentation.js`, is already tracked, so the negation has no effect).
- Updated the PR title to better describe the actual bug being fixed.

One thing I found while testing this out: simply adding a `peerDependencies` field to this package.json (regardless of the Knip ignore-list change) made Knip's default entry-point inference stop tracing `generate-documentation.js`, which caused it to falsely flag all 6 of the tool's other devDependencies (`jsii-docgen`, `jsii-rosetta`, `remark-parse`, `remark-stringify`, `unified`, `unist-util-visit`) as unused. Added an explicit `"entry": ["generate-documentation.js"]` to this workspace's Knip config to fix that — verified `pnpm exec knip` is clean again.

Re-ran the full `pnpm generate-docs:api` pipeline after these changes and confirmed all 5 languages' `constructs` docs (40 files each) still include `with()`/`IMixin`.

PR #306, opened 2026-07-06
chore(cli)!: replace node-fetch with undici

X-Guardian:
### Description

`@cdktn/cli-core` depended on the legacy `node-fetch` package and its proxy helper `https-proxy-agent`. This PR replaces that with the `undici` package — a modern, actively maintained `fetch` implementation with first-class proxy and mocking support — and removes `node-fetch`, `@types/node-fetch`, and `https-proxy-agent`.

#### Changes

- **Fetch → `undici`.** All three files now `import { fetch } from "undici"` instead of `node-fetch`.  We use undici's exported `fetch` rather than the global `fetch` deliberately: it keeps the whole HTTP stack — code and test mocks — on a single undici instance. `globalThis.fetch` uses Node's *internal* undici, which is a different instance from the `undici` npm package, so mocking has to target the same one the code uses.

- **Proxy → undici `ProxyAgent` via the `dispatcher` option.** `node-fetch`'s `agent` option (a Node `http.Agent` from `https-proxy-agent`) is replaced by an undici `ProxyAgent` built from the same `http_proxy` / `HTTP_PROXY` env vars and passed as the request `dispatcher`.

- **Proxy support extended to `package-manager.ts`.** Previously only `registry-api.ts` and `prebuilt-providers.ts` honoured a proxy; `package-manager.ts` (the PyPI / NuGet / Maven Central / GitHub / npm availability probes) never did. It now uses the same `dispatcher`, so all registry traffic respects a corporate proxy consistently.

- **`undici@8.6.0` dependency + Node floor.** `undici` is added to `@cdktn/cli-core`. undici 8 requires Node ≥ 22.19.0, so both published packages (`@cdktn/cli-core` and `cdktn-cli`) declare `"engines": { "node": ">=22.19.0" }` — surfacing a clear install-time warning on older Node instead of an opaque runtime failure.

- **Tests → undici `MockAgent`.** `prebuilt-providers.test.ts` and `package-manager.test.ts` exercise undici's `fetch`, so their `nock` scopes are replaced with an undici `MockAgent` wired via `setGlobalDispatcher`.

### Checklist

### DIFF
--- a/packages/@cdktn/cli-core/package.json
+++ b/packages/@cdktn/cli-core/package.json
@@ -49,14 +49,13 @@
     "constructs": "10.6.0",
     "cross-spawn": "7.0.6",
     "fs-extra": "8.1.0",
-    "https-proxy-agent": "5.0.1",
     "indent-string": "4.0.0",
     "minimatch": "5.1.9",
-    "node-fetch": "2.7.0",
     "pkg-up": "3.1.0",
     "semver": "7.7.4",
     "sscaff": "1.2.274",
     "strip-ansi": "6.0.1",
+    "undici": "8.6.0",
     "xml-js": "1.6.11",
     "xstate": "4.38.3",
     "zod": "3.25.76"
@@ -70,11 +69,13 @@
     "@types/fs-extra": "8.1.5",
     "@types/minimatch": "5.1.2",
     "@types/node": "20.19.1",
-    "@types/node-fetch": "2.6.13",
     "@types/semver": "7.7.1",
-    "nock": "13.5.6",
+    "nock": "^14.0.0",
     "prettier": "2.8.8",
     "tsc-files": "1.1.4",
     "typescript": "^5.0.0"
+  },
+  "engines": {
+    "node": ">=22.19.0"
   }
 }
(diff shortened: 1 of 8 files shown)


jsteinich (approved):

so0k (approved):
LGTM, noticed this is flagged as breaking change (due to Node version bump?)

X-Guardian:
Shouldn't be a breaking change. The `engines.node` property only creates a warning on install.

jsteinich:
> Shouldn't be a breaking change. The `engines.node` property only creates a warning on install.

I had marked as a breaking change due to the part in the PR description: `undici 8 requires Node ≥ 22.19.0`. If that's accurate, then this does seem like a breaking change as we haven't officially dropped support for Node 20 yet.

I don't really see that as a problem as I expect the next release to come with a major version bump regardless.

X-Guardian:
OK Jon, I've restored the breaking change marker in the PR title.

so0k:
LGTM, added docs PR - see preview environment - merging now

Docs: open-constructs/cdk-terrain-docs#28 — bumps the Node.js prerequisite to 22.19+ in the install and deploy-applications tutorials.
Preview: https://cdkterrain-docs-node-22-19-prereqs.mintlify.app

sakul-learning:
Non-blocking follow-up: undici v8.7.0 is now available and retains the same Node.js >=22.19.0 requirement as the v8.6.0 version introduced here. After this transport migration merges, a small dependency-only update to v8.7.0 would be straightforward.

PR #321, opened 2026-07-11
fix(lib)!: preserve symlinks in TerraformAsset walkers

### LINKED ISSUE #320: TerraformAsset: AssetType.ARCHIVE follows symlinks (statSync), duplicating content and crashing on cyclic links
### Expected Behavior

`new TerraformAsset(this, id, { path, type: AssetType.ARCHIVE })` over a directory that contains symlinks (e.g. a pnpm‑style `node_modules`) should produce a zip that is a faithful, bounded representation of the source tree — either by **preserving symlink entries** in the archive, or at minimum by **de‑duplicating by real path** so shared targets are stored once. It should never crash on a valid on‑disk tree.

### Actual Behavior

`archiveSync` walks the tree with `fs.statSync` (which *follows* symlinks) rather than `fs.lstatSync`, so every symlink is silently dereferenced. This causes three distinct problems:

1. **Content duplication / archive bloat.** A target reachable through _N_ symlink paths is copied _N_ times into the zip. With pnpm's heavily‑symlinked `node_modules` this is severe — in a real Lambda bundle it inflated the archive **15 MB → 34 MB (~2.3×)** for byte‑identical logical content.
2. **Symlinks are not preserved.** The output contains **zero** symlink entries; links become full copies. (AWS Lambda's `nodejs22.x` runtime *does* support symlinked deployment packages — verified by a live deploy+invoke — so preserving them would be both smaller and correct.)
3. **Hard crash on circular symlinks.** pnpm trees legitimately contain self/peer‑referential links. Following them with `statSync` eventually hits `ELOOP: too many symbolic links encountered`, which aborts the entire synth via `assetCanNotCreateZipArchive` — not just the one asset.

Root cause, `cdktn/lib/private/fs.js` (`archiveSync` → `walk`):
```js
if (fs.statSync(full).isDirectory()) {   // follows symlinks; should be lstatSync
    walk(full, zipPath);
} else {
    files[zipPath] = fs.readFileSync(full);
}
```

### Steps to Reproduce

Minimal, isolated against cdktn's own archiver (no full project needed):

```js
// repro.js — run with: node repro.js
const fs = require('fs');
const os = require('os');
const path = require('path');
const { archiveSync } = require('cdktn/lib/private/fs'); // internal, for isolation

const src = fs.mkdtempSync(path.join(os.tmpdir(), 'cdktn-archive-'));
fs.mkdirSync(path.join(src, 'real-pkg'));
fs.writeFileSync(path.join(src, 'real-pkg/index.js'), 'x'.repeat(200 * 1024));
fs.symlinkSync('real-pkg', path.join(src, 'link-a')); // two links -> same target
fs.symlinkSync('real-pkg', path.join(src, 'link-b')); // (pnpm shares deps this way)

archiveSync(src, path.join(src, 'out.zip'));
// out.zip contains real-pkg/index.js, link-a/index.js, link-b/index.js
//   => 3 copies of the 200 KB payload, 0 symlink entries preserved.

// Circular link (as pnpm produces) -> crashes the synth:
const cyc = fs.mkdtempSync(path.join(os.tmpdir(), 'cdktn-cyc-'));
fs.mkdirSync(path.join(cyc, 'pkg'));
fs.symlinkSync('../pkg', path.join(cyc, 'pkg/self'));
archiveSync(cyc, path.join(cyc, 'cyc.zip'));
//   => throws: "... failed: Error: ELOOP: too many symbolic links encountered ..."
```

Observed:
- `link-a/index.js` and `link-b/index.js` are full copies (3 total), not symlinks.
- The circular case throws `ELOOP`, failing synth.

Same effect end‑to‑end: `TerraformAsset(..., { path: '<dir with pnpm node_modules>', type: AssetType.ARCHIVE })` → `cdktn synth`.

### Versions

```
cdktn: 0.23.3
constructs: 10.6.0
node: v22 (also repros on v20)
os: darwin arm64
```

**Not OS‑specific** — the defect is `statSync`‑follows‑symlink in `archiveSync`, so it reproduces identically on Linux/CI. (This is distinct from client‑side packagers that flatten symlinks only on some platforms.)

### Migration Context

I used cdktf and saw the same issue there — `packages/cdktf/lib/terraform-asset.ts` carries the same `archiveSync` implementation, so this is inherited from upstream, not a cdktn regression.

### Upstream cdktf Issue

https://github.com/hashicorp/terraform-cdk/issues/3858

### Possible Solutions

- In `archiveSync`'s `walk`, use `fs.lstatSync(full)`; when the entry is a symlink, add a **symlink entry** to the zip (store `readlinkSync(full)` as the entry data with the Unix `S_IFLNK` mode bits) instead of recursing/copying. fflate supports symlink entries via `os`/`attrs`. This preserves fidelity, shrinks the archive, and removes the `ELOOP` crash.
- If preserving symlinks is undesirable, at least **track visited real paths** (`fs.realpathSync`) to dedupe and break cycles, and skip absolute / out‑of‑root link targets.

### Workarounds

- Pre‑zip the directory with a symlink‑aware archiver and pass the resulting file with `AssetType.FILE` (what we did — cut the Lambda bundle 34 MB → 15 MB and eliminated the crash).
- Or produce a dereferenced, hoisted (`node-linker=hoisted`) `node_modules` with no symlinks before handing the dir to `AssetType.ARCHIVE`.

### References

- https://github.com/hashicorp/terraform-cdk/issues/3858
- https://pnpm.io/symlinked-node-modules-structure
- https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/terraform-asset.ts

### Community Note

vincenthsh:
### Description

Closes #320.

`TerraformAsset` walked source trees with `fs.statSync`, which follows symlinks. On symlinked trees (e.g. pnpm's `node_modules`) this meant:

- **Archive bloat**: a target reachable through N symlink paths was copied N times into the zip (15 MB → 34 MB on a real Lambda bundle for byte-identical logical content).
- **No symlink fidelity**: archives contained zero symlink entries — links became full copies.
- **Hard synth crash**: circular symlinks (legitimate in pnpm trees) hit `ELOOP`. Notably this crashed for **every** asset type, not just `ARCHIVE`, because `hashPath` runs in the `TerraformAsset` constructor.

This is a fork regression, not inherited from cdktf: upstream archives with `archiver`, which walks with `lstat` and emits real symlink entries (cycles are structurally impossible there). The regression entered when `archiver` was replaced by yazl in #95 — yazl has no symlink awareness at all (`addFile` stats-and-follows) — and survived the yazl → fflate rewrite in #148, which kept the hand-rolled `statSync` walk.

### The fix

All three walkers in `packages/cdktn/src/private/fs.ts` now use `lstatSync` and treat symlinks first-class, restoring `archiver`/cdktf parity:

- **`archiveSync`** emits real symlink entries: the `readlink` target as STORED entry data with `S_IFLNK | perms` in the unix external attributes and version-made-by host = Unix — exactly what `archiver` produces and what Info-ZIP `unzip`, Go `archive/zip`, and the AWS Lambda runtime (verified live on `nodejs22.x` in #320) recreate as symlinks. Never recursing through links makes cycles unreachable by construction. fflate 0.8.2 supports this natively via per-file `[data, { os, attrs, level }]` tuples (attrs must be caller-pre-shifted, `(mode << 16) >>> 0`).
- **`hashPath`** hashes a symlink by its target path instead of following it — retargeting a link still changes the asset hash, but shared targets are no longer double-counted and cycles no longer crash the constructor.
- **`copySync`** (`AssetType.DIRECTORY`) recreates symlinks with `fs.symlinkSync`.

Two latent zip-metadata gaps fixed along the way (both also `archiver`/cdktf parity):

- Regular-file unix modes are now preserved (executable bits were silently stripped — matters for Lambda binaries and `.bin` scripts).
- Entry mtimes are pinned to a fixed 1980 date, making archives byte-reproducible across synths (fflate defaults `mtime` to `Date.now()`, so every synth previously produced different zip bytes — perpetual drift for anyone hashing the archive, e.g. `filebase64sha256`). The pin uses the local-time `Date` constructor deliberately: fflate encodes DOS dates from local getters and rejects years < 1980, so a UTC midnight date would underflow to 1979 in timezones west of UTC.

Prior art considered and rejected: `terraform-provider-archive` always dereferences and has no cycle detection — it accumulated the opt-in `exclude_symlink_directories` band-aid (v2.4.0, hashicorp/terraform-provider-archive#183; follow-up defect fixed in v2.4.2, #298) and still `ELOOP`s on cycles. Reverting to `archiver` would reintroduce the `execSync` child-process bridge that #148 removed (its API is async-only).

### Hash compatibility (updated after review — downscoped)

Review surfaced a domain collision in the first cut of the `hashPath` change (file bytes and symlink-target bytes shared one unframed stream). Per the requested downscope this PR now carries only the compatibility-preserving resolution (reviewer's option 3):

- the legacy content hash is kept byte-identical for symlink-free trees (proven by a test that recomputes the historical md5 independently);
- symlink metadata (relative path + target) is framed into a separate digest and combined under a tagged outer hash **only when symlinks exist** — so hashes still change only for symlink-bearing trees, which today either bloat 2-3× or crash outright.

The canonical entry-framed scheme and its `canonicalAssetHashes` feature flag moved to the stack: design issue #322 → implementation #323 → docs #324.

`TerraformModuleAsset` stays in scope per review: it shares the corrected copier (its private copier sent directory symlinks into `copyFileSync` → `EISDIR`) with a regression test.

### Testing

- `packages/cdktn/test/archive-symlink.test.ts`: 10 tests — the issue's repros (file/dir symlink preservation via system-`unzip` round-trip + `lstat`, dedup of shared targets, `ELOOP` on cycles for both `archiveSync` and `hashPath`, hash-changes-on-retarget), `copySync` symlink recreation, executable-bit preservation, and byte-identical archives across runs. 5 of the 6 core repros fail on `main`.
- Full `cdktn` package suite green (453 passed; the one pre-existing `matchers.test.ts` → `toPlanSuccessfully` failure shells out to a real `terraform plan` and fails identically without this change).
- Issue #320's literal repro script against the built lib: `zipinfo` shows `lrwxr-xr-x ... stor` entries for `link-a`/`link-b`, one payload copy (538-byte zip vs 3× 200 KiB before), and the cyclic tree archives to a 120-byte zip instead of aborting synth.

### Checklist

- [ ] I have made corresponding changes to the documentation if applicable

### DIFF
--- a/packages/cdktn/src/private/fs.ts
+++ b/packages/cdktn/src/private/fs.ts
@@ -117,37 +117,56 @@ export function archiveSync(src: string, dest: string) {
 
 /**
  * Compute a stable MD5 hash of a file or directory's contents.
- * Directories are hashed by recursively folding each file's contents into
- * the same digest, in directory-listing order.
+ * File contents fold into one digest in directory-listing order, exactly as
+ * earlier releases did, so trees without symlinks keep their historical
+ * hashes. Symlinks are hashed by their metadata (path + target) instead of
+ * being followed, so shared targets are not double-counted and cycles cannot
+ * recurse; that metadata goes into a second digest, and only when symlinks
+ * exist are the two combined under a tagged outer hash — the tag keeps
+ * symlink metadata in a separate domain from file bytes, so a file
+ * containing `foo` can never collide with a symlink targeting `foo`.
  * @param src - path to a file or directory to hash
  * @returns uppercased hex digest, truncated to HASH_LEN characters
  */
 export function hashPath(src: string): string {
-  const hash = crypto.createHash("md5");
+  const content = crypto.createHash("md5");
+  const links = crypto.createHash("md5");
+  let linkCount = 0;
 
   /**
-   * Walk `p`, feeding any file contents into the enclosing hash accumulator.
-   * Symlinks are hashed by their target path instead of being followed, so
-   * shared targets are not double-counted and cycles cannot recurse.
+   * Walk `p`, feeding file contents and symlink metadata into the enclosing
+   * accumulators.
    * @param p - path to walk
+   * @param relPath - path of `p` relative to the walk root, `/`-separated
    * @param isRoot - follow a symlink only at the root, matching how the
    * asset's own path is resolved when it is read
    */
-  function hashRecursion(p: string, isRoot = false) {
+  function hashRecursion(p: string, relPath: string, isRoot = false) {
     const stat = isRoot ? fs.statSync(p) : fs.lstatSync(p);
     if (stat.isSymbolicLink()) {
-      hash.update(fs.readlinkSync(p));
+      links.update(`${relPath}\0${fs.readlinkSync(p)}\0`);
+      linkCount++;
     } else if (stat.isFile()) {
-      hash.update(fs.readFileSync(p));
+      content.update(fs.readFileSync(p));
     } else if (stat.isDirectory()) {
       fs.readdirSync(p).forEach((filename) =>
-        hashRecursion(path.resolve(p, filename)),
+        hashRecursion(
+          path.resolve(p, filename),
+          relPath ? `${relPath}/${filename}` : filename,
+        ),
       );
     }
   }
 
-  hashRecursion(src, true);
-  return hash.digest("hex").slice(0, HASH_LEN).toUpperCase();
+  hashRecursion(src, "", true);
+  if (linkCount === 0) {
+    return content.digest("hex").slice(0, HASH_LEN).toUpperCase();
+  }
+  const outer = crypto.createHash("md5");
+  outer.update("cdktn/asset-hash/symlinks/v1\0");
+  outer.update(content.digest("hex"));
+  outer.update(links.digest("hex"));
+  return outer.digest("hex").slice(0, HASH_LEN).toUpperCase();
 }
 
 /**
--- a/packages/cdktn/src/private/fs.ts
+++ b/packages/cdktn/src/private/fs.ts
@@ -3,10 +3,32 @@
 import * as fs from "fs";
 import * as path from "path";
 import * as crypto from "crypto";
-import { zipSync } from "fflate";
+import { strToU8, zipSync } from "fflate";
+import type { ZipOptions } from "fflate";
 import { assetCanNotCreateZipArchive } from "../errors";
 
 const HASH_LEN = 32;
+
+// Unix file-type bits for zip external attributes (st_mode upper nibble)
+const S_IFLNK = 0o120000;
+const S_IFREG = 0o100000;
+const PERM_MASK = 0o7777;
+// "version made by" host byte; unzip only honors unix mode attrs when set
+const ZIP_OS_UNIX = 3;
+// Pinned so archives are byte-reproducible across synths. fflate encodes
+// DOS dates from local-time getters and rejects years outside 1980-2099,
+// so this must be a local-time 1980 date (a UTC one underflows to 1979 in
+// timezones west of UTC).
+const ZIP_ENTRY_MTIME = new Date(1980, 0, 1);
+
+/**
+ * Zip external attributes for a unix mode: file-type and permission bits
+ * belong in the high 16 bits of the 32-bit field.
+ * @param mode - unix st_mode bits (type | permissions)
+ */
+function zipAttrs(mode: number): number {
+  return (mode << 16) >>> 0;
+}
 
 // Full implementation at https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy/copy-sync.js
 /**
@@ -21,11 +43,12 @@ export function copySync(src: string, dest: string) {
    */
   function copyItem(p: string) {
     const sourcePath = path.resolve(src, p);
-    const stat = fs.statSync(sourcePath);
-    if (stat.isFile()) {
+    const stat = fs.lstatSync(sourcePath);
+    if (stat.isSymbolicLink()) {
+      fs.symlinkSync(fs.readlinkSync(sourcePath), path.resolve(dest, p));
+    } else if (stat.isFile()) {
       fs.copyFileSync(sourcePath, path.resolve(dest, p));
-    }
-    if (stat.isDirectory()) {
+    } else if (stat.isDirectory()) {
       walkSubfolder(p);
     }
   }
@@ -51,20 +74,42 @@ export function copySync(src: string, dest: string) {
  */
 export function archiveSync(src: string, dest: string) {
   try {
-    const files: Record<string, Uint8Array> = {};
+    const files: Record<string, [Uint8Array, ZipOptions]> = {};
     const walk = (dir: string, prefix: string) => {
       for (const entry of fs.readdirSync(dir)) {
         const full = path.join(dir, entry);
         const zipPath = prefix ? `${prefix}/${entry}` : entry;
-        if (fs.statSync(full).isDirectory()) {
+        const stat = fs.lstatSync(full);
+        if (stat.isSymbolicLink()) {
+          // Store the link target as the entry data with S_IFLNK attrs so
+          // extractors recreate the symlink instead of a copy of the target.
+          // Never recursing through links also makes cycles unreachable.
+          files[zipPath] = [
+            strToU8(fs.readlinkSync(full)),
+            {
+              os: ZIP_OS_UNIX,
+              attrs: zipAttrs(S_IFLNK | (stat.mode & PERM_MASK)),
+              level: 0,
+            },
+          ];
+        } else if (stat.isDirectory()) {
           walk(full, zipPath);
         } else {
-          files[zipPath] = fs.readFileSync(full);
+          files[zipPath] = [
+            fs.readFileSync(full),
+            {
+              os: ZIP_OS_UNIX,
+              attrs: zipAttrs(S_IFREG | (stat.mode & PERM_MASK)),
+            },
+          ];
         }
       }
     };
     walk(src, "");
-    fs.writeFileSync(dest, zipSync(files, { level: 9 }));
+    fs.writeFileSync(
+      dest,
+      zipSync(files, { level: 9, mtime: ZIP_ENTRY_MTIME }),
+    );
   } catch (err: any) {
     throw assetCanNotCreateZipArchive(src, dest, err);
   }
@@ -82,11 +127,17 @@ export function hashPath(src: string): string {
 
   /**
    * Walk `p`, feeding any file contents into the enclosing hash accumulator.
+   * Symlinks are hashed by their target path instead of being followed, so
+   * shared targets are not double-counted and cycles cannot recurse.
    * @param p - path to walk
+   * @param isRoot - follow a symlink only at the root, matching how the
+   * asset's own path is resolved when it is read
    */
-  function hashRecursion(p: string) {
-    const stat = fs.statSync(p);
-    if (stat.isFile()) {
+  function hashRecursion(p: string, isRoot = false) {
+    const stat = isRoot ? fs.statSync(p) : fs.lstatSync(p);
+    if (stat.isSymbolicLink()) {
+      hash.update(fs.readlinkSync(p));
+    } else if (stat.isFile()) {
       hash.update(fs.readFileSync(p));
     } else if (stat.isDirectory()) {
       fs.readdirSync(p).forEach((filename) =>
@@ -95,7 +146,7 @@ export function hashPath(src: string): string {
     }
   }
 
-  hashRecursion(src);
+  hashRecursion(src, true);
   return hash.digest("hex").slice(0, HASH_LEN).toUpperCase();
 }
 
--- a/packages/cdktn/src/terraform-module-asset.ts
+++ b/packages/cdktn/src/terraform-module-asset.ts
@@ -9,7 +9,7 @@ import * as os from "os";
 import * as fs from "fs";
 import { TerraformStack } from "./terraform-stack";
 import { AssetType, TerraformAsset } from "./terraform-asset";
-import { hashPath } from "./private/fs";
+import { copySync, hashPath } from "./private/fs";
 
 const TERRAFORM_MODULE_ASSET_SYMBOL = Symbol.for("cdktf.TerraformModuleAsset");
 
@@ -126,20 +126,4 @@ export function findLowestCommonPath(paths: string[]): string | undefined {
 
   const relativePath = path.relative(process.cwd(), absolutePathPrefix);
   return relativePath === "" ? "." : relativePath;
-}
-
-/**
- * Copies a file or directory recursively
- * @param from
- * @param to
- */
-function copySync(from: string, to: string) {
-  if (fs.lstatSync(from).isDirectory()) {
-    fs.mkdirSync(to, { recursive: true });
-    for (const file of fs.readdirSync(from)) {
-      copySync(path.join(from, file), path.join(to, file));
-    }
-  } else {
-    fs.copyFileSync(from, to);
-  }
 }
--- a/packages/cdktn/test/archive-symlink.test.ts
+++ b/packages/cdktn/test/archive-symlink.test.ts
@@ -141,6 +141,33 @@ describe("hashPath symlink handling (#320)", () => {
 
     expect(before).not.toBe(after);
   });
+
+  test("a regular file and a symlink with the same payload hash differently", () => {
+    // both trees contain a `current` entry whose payload bytes are `a.txt`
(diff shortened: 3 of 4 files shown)


so0k:
flagged as breaking change as it changes the checksums / SHA created by the `TerraformAsset` Archiver, and I agree that is should not be put behind a feature flag because:

> Asset hashes change only for trees that contain symlinks — trees that today either bloat 2-3× or crash outright.

Altho I'm not sure if we want this (or it should be possible to opt-in / default opt-out to this `mtime` behavior)

> Entry mtimes are pinned to a fixed 1980 date, making archives byte-reproducible across synths (fflate defaults mtime to Date.now(), so every synth previously produced different zip bytes — perpetual drift for anyone hashing the archive, e.g. filebase64sha256). The pin uses the local-time Date constructor deliberately: fflate encodes DOS dates from local getters and rejects years < 1980, so a UTC midnight date would underflow to 1979 in timezones west of UTC.

packages/cdktn/src/private/fs.ts
```diff
@@ -82,11 +127,17 @@ export function hashPath(src: string): string {
...
+   * @param isRoot - follow a symlink only at the root, matching how the
+   * asset's own path is resolved when it is read
    */
-  function hashRecursion(p: string) {
-    const stat = fs.statSync(p);
-    if (stat.isFile()) {
+  function hashRecursion(p: string, isRoot = false) {
+    const stat = isRoot ? fs.statSync(p) : fs.lstatSync(p);
+    if (stat.isSymbolicLink()) {
+      hash.update(fs.readlinkSync(p));
```
sakul-learning: **Blocking:** symlink targets and regular-file contents are fed into the same unframed hash stream. A regular file containing `foo` and a symlink whose target text is `foo` both contribute exactly the bytes `foo`, so replacing one with the other can leave `assetHash` unchanged even though `archiveSync`/`copySync` emit materially different artifacts. `TerraformAsset` then reuses that hash in its synthesized path.

There are a few implementation choices with different compatibility costs:

1. **Frame every entry as `(type, relative path, payload length, payload)`** — cleanest and fixes type, filename, and concatenation ambiguities, but intentionally changes hashes for all assets.
2. **Add only a marker before symlink targets** — smallest diff and generally changes only symlinked trees, but it is not true domain separation because a regular file can contain the same marker bytes; it also leaves existing filename/boundary ambiguity.
3. **Keep the legacy content hash for compatibility, separately hash framed symlink metadata, and combine them under a tagged outer hash only when symlinks exist** — preserves hashes for trees without symlinks while giving symlink-bearing trees an unambiguous metadata domain. It leaves older non-symlink filename ambiguity untouched, but is the narrowest robust fix for this PR.

I favor option 3 if preserving unaffected hashes matters, or option 1 if this breaking release is the right opportunity to make `hashPath` canonical. Please add a regression proving a same-named regular file containing `foo` hashes differently from a symlink targeting `foo`.

packages/cdktn/src/private/fs.ts
```diff
@@ -21,11 +43,12 @@ export function copySync(src: string, dest: string) {
    */
```
sakul-learning, line 43: Could `TerraformModuleAsset` reuse this corrected copier (or receive the same symlink branch) in this PR? Its separate recursive copier at `terraform-module-asset.ts:136-144` uses `lstatSync`, but routes every non-directory—including a directory symlink—to `copyFileSync`; that can fail with `EISDIR` before construction reaches these fixed walkers. Since this PR is small and addresses symlink-induced synthesis failures across asset walking, one module-asset regression plus sharing this implementation would close the remaining path without broadening the design.

sakul-learning (changes requested):
## Review summary

The symlink fix itself is well targeted, and the new tests pass the behavior-oriented/value assessment: each core test exercises an observable regression rather than implementation text. In particular, the extraction tests prove real symlink fidelity; the shared-target size test catches archive materialization/bloat; the circular-link tests protect all three walkers from `ELOOP`; retargeting checks asset invalidation; and the directory-copy test proves links are recreated rather than dereferenced. The executable-mode test protects a real Lambda/CLI artifact property, while the byte-identical archive test protects deterministic output (subject to resolving the maintainer tradeoff already raised about forcing mtimes). These tests are included by the normal `cdktn` Nx/Jest target, and the configured full run passed: 39 suites, 454 tests, and 288 snapshots. The focused validations suite and `nx build cdktn` also passed.

I found one blocking correctness issue in the new hash representation: regular-file bytes and symlink-target bytes occupy the same unframed hash domain, so materially different artifacts can receive the same `assetHash` and synthesized path. See the inline comment for options and compatibility tradeoffs.

I also think this small PR should cover `TerraformModuleAsset`'s separate source copier. `packages/cdktn/src/terraform-module-asset.ts:136-144` sends every non-directory, including a directory symlink, to `copyFileSync`, which can fail with `EISDIR` before the corrected `TerraformAsset` walkers are reached. Reusing the corrected helper (or adding equivalent symlink handling) plus one module-asset regression would address the same class of synthesis failures without materially expanding the PR. GitHub cannot attach a review comment to those unchanged lines, so this item is recorded here and alongside the shared copier.

No workflow files changed, and I found no security concern in the touched code. Ponytail/artifact-value pass: the core regression tests earn their maintenance cost; no speculative abstraction or low-signal static test was added.

packages/cdktn/src/private/fs.ts
so0k: I'd go for option 3, the justification so far was that the changes only affected archives with symlinks which were broken since archiver was replaced. Spent more time looking into archiver handled this and if this a concern archiver had

vincenthsh: Resolved in 0d8c0555d, landing on a combination of options 3 and 1 behind the repo's existing feature-flag lifecycle:

- **Default (existing projects): option 3** — as @so0k preferred. The legacy content hash stays byte-identical for symlink-free trees; symlink metadata (`relPath\0target\0` per link) goes into a separate digest; the two are combined under a tagged outer hash (`cdktn/asset-hash/symlinks/v1`) only when symlinks exist. That gives symlink-bearing trees true domain separation — a file containing `foo` can no longer collide with a symlink targeting `foo` — while unaffected assets keep their hashes.
- **`canonicalAssetHashes` feature flag (new `FUTURE_FLAGS` entry, on for new projects, opt-in for existing): option 1** — every entry framed as `(type, relative path, payload size, payload)` in sorted order. This is the construction git trees, Nix NAR, and Go's dirhash all use, and it additionally fixes the pre-existing blindness to renames and entry-boundary shifts. It graduates to the unconditional default at the next major, at which point the hybrid is deleted.

On the "preserve archiver-era hashes" angle we double-checked upstream: archiver never hashed anything — cdktf's `hashPath` is the same `statSync`-following walk cdktn inherited, so it also ELOOPs on cyclic trees in the constructor and double-counts symlinked content. There was no working symlink hash behavior to preserve, only the symlink-free one, which the default scheme keeps exactly.

The requested regression is in (`a regular file and a symlink with the same payload hash differently`, parameterized over both schemes), plus rename/boundary tests for the canonical scheme and a flag test at the `TerraformAsset` level (note `Testing.app()` enables `FUTURE_FLAGS` by default, so the compat case uses `enableFutureFlags: false`).

**Follow-up required:** website docs for the new flag — the feature-flags page and the `TerraformAsset`/module-asset docs need to describe `canonicalAssetHashes`, its hash-migration impact when opting in, and the next-major graduation. Happy to do that in a docs PR right after this lands.

packages/cdktn/src/private/fs.ts
vincenthsh: Done in 0d8c0555d — `TerraformModuleAsset` now reuses the shared symlink-aware `copySync` (its private copier is deleted) and passes the `canonicalAssetHashes` flag through to `hashPath` like `TerraformAsset` does. Regression added: a module source containing a directory symlink constructs cleanly (previously `copyFileSync` → `EISDIR`).

vincenthsh:
> Altho I'm not sure if we want this (or it should be possible to opt-in / default opt-out to this `mtime` behavior)

Keeping the pin, on prior art: Go module zips do exactly this — the `golang.org/x/mod/zip` format spec says *"File permissions and timestamps are also ignored"* (entries get the zero DOS date; the format bans symlinks outright, so timestamps are its only nondeterminism to kill). Python wheels likewise clamp entry mtimes to 1980 for reproducibility (the zip format can't represent earlier dates). The reproducible-builds ecosystem treats embedded wall-clock timestamps in archives as a defect.

Two things bound the blast radius here: `assetHash` (and therefore the asset *path*) comes from `hashPath` over the source tree, never from the zip bytes, so the pin changes no paths; and the previous behavior was `fflate`'s `Date.now()` default — meaning nobody can be depending on meaningful mtimes inside these archives today, because every synth already produced fresh ones. The pin only removes noise for anyone hashing the artifact (`filebase64sha256` on the archive path currently drifts on every synth).

If someone does turn up with a real need for source mtimes we can add a knob then, but I'd rather not grow the JSII surface for it preemptively.

sakul-learning:
## Requested downscope: keep option 3 here; move canonical hashing out

The compatibility-preserving option 3 now resolves the blocker from the earlier review: trees without symlinks keep their historical content hash, while symlink-bearing trees combine separately framed symlink metadata under a tagged outer digest. That is sufficient for #320 and should remain in this PR.

Please remove the additional `canonicalAssetHashes` implementation, feature flag, and canonical-only tests from this PR. I opened #322 for the implementer to take as that independently reviewable, feature-flagged follow-up:

https://github.com/open-constructs/cdk-terrain/issues/322

The reason to separate it is not only scope. The current canonical representation has a metadata hole: it frames file/symlink type, relative path, payload length, and payload, but regular-file records omit the mode that `archiveSync` preserves, and directories contribute no records. I reproduced `0644` → `0755` producing identical canonical hashes while producing different archive bytes; adding an empty directory is likewise invisible even where the asset copier emits it.

The Git/Nix precedent did not expose this problem because their canonical formats serialize more than payload framing:

- Git tree entries include mode, name, and object ID; executable files (`100755`), ordinary files (`100644`), symlinks (`120000`), and subtrees are structurally distinct.
- Nix NAR represents directory entries/node types and an executable marker.

So the framing pattern was borrowed without the complete metadata model that makes those formats canonical. #322 captures the needed contract and regressions for modes, empty directories, paths, entry boundaries, and symlink metadata.

The `TerraformModuleAsset` copier fix should **stay** in this PR. It is not speculative canonical-hash work: its existing copier is another active walker over the same asset source class, and a directory symlink reaches `copyFileSync` and fails with `EISDIR` before the corrected `TerraformAsset` walkers run. Reusing the shared symlink-aware copier is a small deletion/consolidation that fixes the same concrete synthesis-failure class and now has a regression test.

vincenthsh:
Downscoped as requested — the branch was force-pushed so this PR is now exactly: the three symlink-aware walkers, the option-3 compatibility-preserving hash, and the shared `TerraformModuleAsset` copier with its `EISDIR` regression (kept in scope per your note). The `canonicalAssetHashes` flag, canonical scheme, and canonical-only tests are gone from this diff.

The follow-up is a stack, review-ready:

- **#323** implements #322 on top of this branch (only its last commit is new). It adopts your metadata-model finding directly: file **and** symlink records now frame the octal permission mask (`0o7777` bits — exactly what `archiveSync` emits into zip external attributes), and directories — including empty ones, which `copySync` materializes — contribute explicit `D <relPath>\0` records (no mode, since neither emitter preserves directory permissions). Every acceptance criterion in #322 has a dedicated test, including your reproduced `0644` → `0755` case and empty-directory visibility.
- **#324** (stacked on #323) adds `docs/feature-flags.md`, documenting the flag lifecycle, why canonical hashing is the recommended future default, and the complete-metadata-model lesson from your review.

Thanks for the git/Nix observation — you were right that the first draft borrowed the framing without the object model that makes those formats canonical.

packages/cdktn/src/private/fs.ts
so0k: canonical hash path support was scoped out of this PR and moved to stacked PR

jsteinich (approved):

sakul-learning (approved):
Approved after the downscope and successful checks on `ff817b2`. The option-3 hash fix and `TerraformModuleAsset` copier regression are in scope; canonical hashing remains correctly split into the stacked follow-up.

Documentation follow-up for `open-constructs/cdk-terrain-docs`: please explain the deterministic ZIP timestamp policy and cite `golang.org/x/mod/zip` as precedent (its format says “File permissions and timestamps are also ignored”). Phrase the comparison precisely: TerraformAsset archives similarly avoid meaningful source timestamps by pinning ZIP mtimes, but unlike Go module ZIPs they preserve regular-file permission bits and support symlinks.

PR #323, opened 2026-07-11
feat(lib): canonical asset hashes behind the canonicalAssetHashes feature flag

### LINKED ISSUE #322: feat(lib): define a complete canonical asset-hash representation
## Problem

PR #321 explored an opt-in `canonicalAssetHashes` scheme that frames entry type, relative path, payload size, and payload. That is useful follow-up work, but the current representation does not yet describe every artifact-affecting property:

- regular-file records omit the Unix mode that `archiveSync` preserves in ZIP external attributes;
- directories contribute no records, so adding or removing an empty directory is invisible.

A reproduced mode case changes the emitted archive bytes while retaining the same canonical asset hash:

```json
{
  "canonicalHashBefore": "0D3C9E0875B72106054DC334DF123995",
  "canonicalHashAfter": "0D3C9E0875B72106054DC334DF123995",
  "canonicalEqual": true,
  "archiveBytesEqual": false
}
```

This work was split conceptually from PR #321 so that its symlink fix can retain the narrower compatibility-preserving approach: keep the historical content hash for trees without symlinks and combine separately framed symlink metadata under a tagged outer digest only for symlink-bearing trees.

## Why Git/Nix do not have this hole

The framing idea is sound, but Git and Nix serialize a richer object model than `(type, path, payload length, payload)` alone:

- A Git tree records each child's mode, name, and object ID. Directories are represented by tree objects, regular executable files differ through mode `100755` versus `100644`, and symlinks use mode `120000`.
- Nix NAR serialization explicitly represents directory entries/node types and an executable marker for regular files.

Therefore copying only the payload framing pattern, without their metadata model, drops properties those formats include in their canonical representation.

## Proposed scope

Design and introduce the canonical scheme in a dedicated, feature-flagged PR:

1. Define which source-tree properties affect emitted `ARCHIVE`, `DIRECTORY`, and `FILE` assets.
2. Frame directory entries explicitly, including empty directories.
3. Include relevant preserved mode bits (at minimum executable state; preferably the exact permission mask if that is the emitted contract).
4. Keep record ordering and path normalization deterministic across supported platforms.
5. Specify root-symlink behavior.
6. Add compatibility and behavior tests before enabling the future flag.

## Acceptance criteria

- File rename changes the canonical hash.
- Entry-boundary shifts change the canonical hash.
- File/symlink swaps and symlink retargeting change the canonical hash.
- `0644` → `0755` changes the canonical hash when it changes emitted artifact metadata.
- Adding/removing an empty directory changes the canonical hash for directory/archive assets where that directory is emitted.
- Identical logical trees hash deterministically across repeated runs.
- Legacy hashing remains available until the intended compatibility boundary.

Follow-up from #321.

vincenthsh:
### Description

Closes #322. ~~Stacked on #321~~ — #321 has merged and this branch is rebased onto current `main`; the diff is now the single canonical-hash commit.

Adds the canonical entry-framed asset hash as a `canonicalAssetHashes` feature flag (`FUTURE_FLAGS`: enabled for new projects by `cdktn init`, opt-in via `cdktf.json` context for existing projects, becomes the default at the next major). The legacy scheme — including #321's compatibility-preserving symlink handling — stays untouched as the default.

### The canonical representation

Modeled on git trees and Nix NAR, which serialize a full metadata model rather than payload framing alone. Every record contains what affects the emitted artifact:

| entry | frame |
| --- | --- |
| file | `F <mode> <relPath>\0<size>\0` + content |
| symlink | `L <mode> <relPath>\0<target byte length>\0` + target |
| directory (incl. empty) | `D <relPath>\0` |

- `<mode>` is the octal permission mask (`0o7777` bits) — exactly the bits `archiveSync` preserves in zip external attributes, closing the hole found in review where `0644` → `0755` changed archive bytes but not the hash.
- Directories contribute explicit records (no mode — neither `archiveSync` nor `copySync` preserves directory permissions), so adding/removing an empty directory is visible.
- Entries are framed in sorted directory order with `/`-separated relative paths; a symlink at the root is followed, matching how the asset source path is opened when the artifact is emitted.

### Acceptance criteria from #322

Each has a dedicated test in `packages/cdktn/test/canonical-asset-hash.test.ts`:

- [x] File rename changes the canonical hash (and a paired assertion that legacy cannot see it)
- [x] Entry-boundary shifts change the canonical hash
- [x] File/symlink swaps and symlink retargeting change the canonical hash
- [x] `0644` → `0755` changes the canonical hash
- [x] Adding/removing an empty directory changes the canonical hash
- [x] Identical logical trees hash deterministically (two locations + repeated runs)
- [x] Legacy hashing remains available (flag off; `Testing.app({ enableFutureFlags: false })` covers the compat case per the `features.ts` test rule)

### Testing

Full `cdktn` suite green (464 passing; the pre-existing environmental `matchers.test.ts` → `toPlanSuccessfully` failure shells out to a real `terraform plan` and is unrelated). jsii build green. Docs for the flag are being handled separately in the docs repository (the stacked docs PR #324 was closed in favor of that).

### Checklist

- [x] I have made corresponding changes to the documentation if applicable (stacked docs PR follows)

### DIFF
--- a/packages/cdktn/src/private/fs.ts
+++ b/packages/cdktn/src/private/fs.ts
@@ -115,20 +115,43 @@ export function archiveSync(src: string, dest: string) {
   }
 }
 
+export interface HashPathOptions {
+  /**
+   * Use the canonical entry-framed hash instead of the legacy
+   * content-concatenation hash. Enabled through the `canonicalAssetHashes`
+   * feature flag.
+   */
+  readonly canonical?: boolean;
+}
+
 /**
  * Compute a stable MD5 hash of a file or directory's contents.
- * File contents fold into one digest in directory-listing order, exactly as
- * earlier releases did, so trees without symlinks keep their historical
- * hashes. Symlinks are hashed by their metadata (path + target) instead of
- * being followed, so shared targets are not double-counted and cycles cannot
- * recurse; that metadata goes into a second digest, and only when symlinks
- * exist are the two combined under a tagged outer hash — the tag keeps
- * symlink metadata in a separate domain from file bytes, so a file
- * containing `foo` can never collide with a symlink targeting `foo`.
+ * In both schemes symlinks are hashed by their metadata (path + target)
+ * instead of being followed, so shared targets are not double-counted and
+ * cycles cannot recurse; a symlink at the root itself is followed, matching
+ * how the asset source path is opened when the artifact is emitted.
  * @param src - path to a file or directory to hash
+ * @param options - hash scheme selection, see {@link HashPathOptions}
  * @returns uppercased hex digest, truncated to HASH_LEN characters
  */
-export function hashPath(src: string): string {
+export function hashPath(src: string, options: HashPathOptions = {}): string {
+  const digest = options.canonical
+    ? canonicalHashPath(src)
+    : legacyHashPath(src);
+  return digest.slice(0, HASH_LEN).toUpperCase();
+}
+
+/**
+ * Legacy-compatible hash: file contents fold into one digest in
+ * directory-listing order, exactly as earlier releases did, so trees without
+ * symlinks keep their historical hashes. Symlink metadata goes into a second
+ * digest, and only when symlinks exist are the two combined under a tagged
+ * outer hash — the tag keeps symlink metadata in a separate domain from file
+ * bytes, so a file containing `foo` can never collide with a symlink
+ * targeting `foo`.
+ * @param src - path to a file or directory to hash
+ */
+function legacyHashPath(src: string): string {
   const content = crypto.createHash("md5");
   const links = crypto.createHash("md5");
   let linkCount = 0;
@@ -160,13 +183,68 @@ export function hashPath(src: string): string {
 
   hashRecursion(src, "", true);
   if (linkCount === 0) {
-    return content.digest("hex").slice(0, HASH_LEN).toUpperCase();
+    return content.digest("hex");
   }
   const outer = crypto.createHash("md5");
   outer.update("cdktn/asset-hash/symlinks/v1\0");
   outer.update(content.digest("hex"));
   outer.update(links.digest("hex"));
-  return outer.digest("hex").slice(0, HASH_LEN).toUpperCase();
+  return outer.digest("hex");
+}
+
+/**
+ * Canonical hash, modeled on git trees and Nix NAR: every entry is framed
+ * with everything that affects the emitted artifact —
+ *
+ * - files:      `F <mode> <relPath>\0<size>\0` + content, where `<mode>` is
+ *               the octal permission mask (`0o7777` bits) that archiveSync
+ *               preserves in zip external attributes,
+ * - symlinks:   `L <mode> <relPath>\0<target byte length>\0` + target,
+ * - directories (including empty ones, which copySync materializes):
+ *               `D <relPath>\0`, with no mode — neither archiveSync nor
+ *               copySync preserves directory permissions,
+ *
+ * in sorted directory order with `/`-separated relative paths, so renames,
+ * entry-boundary shifts, permission changes, empty-directory changes, and
+ * file-vs-symlink swaps all change the digest.
+ * @param src - path to a file or directory to hash
+ */
+function canonicalHashPath(src: string): string {
+  const hash = crypto.createHash("md5");
+
+  /**
+   * Walk `p`, framing each entry into the enclosing hash accumulator.
+   * @param p - path to walk
+   * @param relPath - path of `p` relative to the walk root, `/`-separated
+   * @param isRoot - follow a symlink only at the root, matching how the
+   * asset's own path is resolved when the artifact is emitted
+   */
+  function hashRecursion(p: string, relPath: string, isRoot = false) {
+    const stat = isRoot ? fs.statSync(p) : fs.lstatSync(p);
+    const mode = (stat.mode & PERM_MASK).toString(8);
+    if (stat.isSymbolicLink()) {
+      const target = fs.readlinkSync(p);
+      hash.update(`L ${mode} ${relPath}\0${Buffer.byteLength(target)}\0`);
+      hash.update(target);
+    } else if (stat.isFile()) {
+      const data = fs.readFileSync(p);
+      hash.update(`F ${mode} ${relPath}\0${data.length}\0`);
+      hash.update(data);
+    } else if (stat.isDirectory()) {
+      if (relPath) {
+        hash.update(`D ${relPath}\0`);
+      }
+      for (const filename of fs.readdirSync(p).sort()) {
+        hashRecursion(
+          path.resolve(p, filename),
+          relPath ? `${relPath}/${filename}` : filename,
+        );
+      }
+    }
+  }
+
+  hashRecursion(src, "", true);
+  return hash.digest("hex");
 }
 
 /**
--- a/packages/cdktn/src/private/fs.ts
+++ b/packages/cdktn/src/private/fs.ts
@@ -122,6 +122,14 @@ export interface HashPathOptions {
    * feature flag.
    */
   readonly canonical?: boolean;
+  /**
+   * Frame for an archive artifact: archiveSync never emits ZIP directory
+   * entries, so canonical hashing omits directory records and the hash
+   * tracks the emitted archive bytes exactly. Non-empty directories stay
+   * visible through the relative paths of their contents. Has no effect on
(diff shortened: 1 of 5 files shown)


packages/cdktn/src/private/fs.ts
```diff
@@ -51,52 +74,177 @@ export function copySync(src: string, dest: string) {
...
+ * - files:      `F <mode> <relPath>\0<size>\0` + content, where `<mode>` is
+ *               the octal permission mask (`0o7777` bits) that archiveSync
+ *               preserves in zip external attributes,
+ * - symlinks:   `L <mode> <relPath>\0<target byte length>\0` + target,
+ * - directories (including empty ones, which copySync materializes):
+ *               `D <relPath>\0`, with no mode — neither archiveSync nor
+ *               copySync preserves directory permissions,
+ *
+ * in sorted directory order with `/`-separated relative paths, so renames,
+ * entry-boundary shifts, permission changes, empty-directory changes, and
```
sakul-learning, line 218: **Blocking:** Canonical hashing records every non-root directory, including empty ones, but `archiveSync()` only recurses into directories and never emits ZIP directory entries. For `AssetType.ARCHIVE`, adding an empty directory therefore changes `assetHash` and the synthesized asset path while producing identical ZIP bytes. That conflicts with #322’s criterion limiting this sensitivity to assets where the directory is emitted. Please either emit explicit directory entries in archives or make canonical hashing asset-type-aware and omit directory records for archives that do not emit them; an archive-level regression should prove the hash tracks the emitted artifact.

packages/cdktn/src/terraform-module-asset.ts
```diff
@@ -69,7 +70,11 @@ export class TerraformModuleAsset extends Construct {
     this.asset = new TerraformAsset(this, "asset", {
       path: tmpDir,
       type: AssetType.DIRECTORY,
-      assetHash: staticModuleAssetHash ?? hashPath(relativeAssetPath),
+      assetHash:
+        staticModuleAssetHash ??
+        hashPath(relativeAssetPath, {
```
sakul-learning: **Blocking:** The emitted module asset contains only `moduleSources` copied into `tmpDir`, but this hashes their lowest common source ancestor. Unrelated siblings under that ancestor can affect the automatic hash without appearing in the emitted asset; canonical path/mode/directory framing adds new churn from unrelated renames, mode changes, and empty directories. Please hash `tmpDir` after copying (or frame exactly the selected module sources), and add a regression proving unrelated siblings do not change the module asset hash.

sakul-learning (changes requested):
Two emitted-artifact/hash consistency issues need resolution; details are inline. I’ll post the configured-check and test-gap summary separately.

sakul-learning:
## Testing and coverage evaluation

All configured checks passed at `9d97256b344349da0a8218cb369c83e0892f0625`:

- `pnpm exec jest packages/cdktn/test/validations.test.ts --runInBand` — 1 suite, 31 tests, 9 snapshots passed.
- `pnpm exec nx build cdktn` — passed with 0 errors and 0 warnings.
- `pnpm exec nx test cdktn --runInBand` — 40 suites, 465 tests, and 288 snapshots passed.

The canonical hash suite runs through the normal Nx/Jest target and provides useful behavioral coverage for renames, entry-boundary shifts, permission changes, empty directories, file/symlink distinctions, symlink retargeting, and deterministic trees.

Two output-level gaps remain and correspond to the inline correctness findings:

1. The empty-directory test covers `copySync`/directory semantics but not `AssetType.ARCHIVE`, where `archiveSync` currently omits directory entries. An archive regression should establish that hash changes correspond to emitted ZIP changes.
2. The feature integration test covers `TerraformAsset`, but not `TerraformModuleAsset`, whose emitted tree is assembled separately in `tmpDir`. A module test should prove that only selected/copied module sources affect the automatic hash and that unrelated siblings do not cause asset-path churn.

Ponytail/artifact-value pass: the core framing tests earn their maintenance cost and run in normal CI. Some canonical symlink cases overlap the stacked #321 coverage, but they still exercise a separate implementation. I would prioritize the two missing emitted-artifact regressions over adding further helper-level cases.

No workflow files changed, and I found no new security issue in the touched code.

jsteinich (approved):

packages/cdktn/src/private/fs.ts
```diff
@@ -160,13 +183,68 @@ export function hashPath(src: string): string {
...
+      const target = fs.readlinkSync(p);
+      hash.update(`L ${mode} ${relPath}\0${Buffer.byteLength(target)}\0`);
+      hash.update(target);
+    } else if (stat.isFile()) {
+      const data = fs.readFileSync(p);
+      hash.update(`F ${mode} ${relPath}\0${data.length}\0`);
+      hash.update(data);
+    } else if (stat.isDirectory()) {
+      if (relPath) {
+        hash.update(`D ${relPath}\0`);
```
sakul-learning, line 247: **Blocking:** This canonical directory record applies to `AssetType.ARCHIVE` too, but `archiveSync()` only recurses into directories and never emits ZIP directory entries. Adding an empty directory therefore changes the archive’s `assetHash` and synthesized path while leaving the emitted ZIP unchanged. This conflicts with #322’s criterion limiting empty-directory sensitivity to assets where the directory is emitted. Please either emit explicit directory entries in archives or make canonical hashing asset-type-aware and omit directory records for archives that do not emit them; add an archive-level regression that correlates the hash with emitted ZIP content.

packages/cdktn/src/terraform-module-asset.ts
sakul-learning: **Blocking:** The emitted module asset contains only `moduleSources` copied into `tmpDir`, but this hashes their entire lowest common source ancestor. Unrelated siblings can affect the automatic hash without appearing in the emitted asset; canonical path/mode/directory framing additionally makes unrelated renames, mode changes, and empty directories cause asset-path churn. Please hash `tmpDir` after copying (or frame exactly the selected module sources), and add a regression proving unrelated siblings do not change the module asset hash.

sakul-learning (changes requested):
The rebase onto merged #321 leaves both previously identified emitted-artifact/hash consistency blockers unresolved. Reposting them on the current diff.

packages/cdktn/src/private/fs.ts
so0k: Fixed in 554fcbe54 by making canonical hashing asset-type-aware (the second remedy you offered): `HashPathOptions` gains an `archive` option and `TerraformAsset` sets it for `AssetType.ARCHIVE`, so directory records are omitted exactly where `archiveSync` emits none — the hash now tracks the emitted ZIP bytes. Non-empty directories stay visible through their contents' relative paths. We chose this over emitting ZIP directory entries because that would change emitted archive bytes for all users (including flag-off), which this hashing-scoped flag shouldn't do.

New regressions in `canonical-asset-hash.test.ts` ("canonical archive hashing tracks the emitted ZIP") correlate hash and ZIP bytes in both directions: an empty directory changes neither (while the DIRECTORY-asset hash still moves), and a permission change or a rename inside a non-empty directory changes both. A `TerraformAsset`-level test asserts ARCHIVE assets use the archive framing.

packages/cdktn/src/terraform-module-asset.ts
so0k: Fixed in 554fcbe54: under the canonical flag, `TerraformModuleAsset` now hashes `tmpDir` after copying — the exact emitted tree, with deterministic module-relative paths — so unrelated siblings under the sources' lowest common ancestor can no longer affect the hash, and canonical framing adds no churn from unrelated renames, mode changes, or empty directories. The legacy scheme intentionally keeps hashing the ancestor to preserve historical hashes/asset paths for existing projects, consistent with this PR's compatibility stance.

New regressions in `canonical-asset-hash.test.ts` ("TerraformModuleAsset with the canonicalAssetHashes flag"): sibling content changes and empty-sibling additions leave the module asset path unchanged while a change to a selected module source still moves it, plus a paired test documenting the preserved legacy behavior.

sakul-learning (approved):
All previously requested changes are resolved at c79fc67a2.

- Canonical archive hashes now omit directory records that ZIP emission omits.
- Canonical TerraformModuleAsset hashing uses the staged emitted tree.
- archiveSync traversal is sorted to match canonical hash ordering, with deterministic reversed-enumeration coverage.
- Configured validation, build, and full cdktn test checks pass (40 suites, 473 tests, 288 snapshots).

Approved.

PR #333, opened 2026-07-13
chore(deps): Update xstate to v5.32.4

X-Guardian:
### Description

Bumps `xstate` from `4.38.3` to `5.32.4` (latest) in `@cdktn/cli-core`, the only package that declares it.

This is a major version bump (4.x → 5.x) and required migrating the deploy state machine, its consumer, and its tests to the reworked v5 API.

#### State machine (`deploy-machine.ts`)

- **`createMachine<Ctx, Evt, State>(…)` → `setup({ types, actors }).createMachine(…)`.** The generic-parameter form was replaced by the `setup()` builder with a `types` block. The `predictableActionArguments` flag was dropped (predictable ordering is the default in v5) and `services` became `actors`.
- **Invoked callback service → `fromCallback`.** The v4 `(context, event) => (send, onReceive) => {…}` factory became `fromCallback(({ sendBack, receive, input }) => …)`. Since v5 invoked actors receive `input` rather than the triggering event, the spawn config now flows in through `invoke.input`. The actor is built by `makeTerraformPtyService(spawn)` so tests can inject a mock pty.
- **`send(…, { to })` → `sendTo(…)`; `assign<Ctx, Evt>` → `assign`.** Action creators were split and the `assign` implementation signature changed to a single argument.
- **`raise(…)` → `sendTo(({ self }) => self, …)`.** The two internally re-emitted events (`EXITED` on a missing variable, and the `OVERRIDE_REJECTED_EXTERNALLY` re-label of an external discard) are sent to self rather than raised. Raised events are internal micro-steps and do not surface through the inspection API, so the consumer would otherwise stop observing them.
- **Exit code as final-state output.** The terraform exit code is captured into context on the `EXITED` transitions and surfaced through the machine's root `output`, so consumers read it off the settled snapshot rather than scraping it from an event. This also puts the previously-unused `DeployContext.exitCode` field to work.
- Named helper types `DeploySnapshot` (`SnapshotFrom<typeof deployMachine>`) and `DeployActor` (`ActorRefFrom<typeof deployMachine>`) are exported for the consumer.

#### Consumer (`terraform-cli.ts`)

This needed the most care, because v5 snapshots no longer expose the triggering event (`state.event`):

- **Transitions** come from the actor's own `service.subscribe(…)`, which yields a fully-typed snapshot for the root actor only — replacing `service.onTransition(…)`. Because the factory starts the actor and sends `START` before we subscribe, the current snapshot is processed once immediately after subscribing so the initial `idle → running` transition is not missed.
- **Events** come from the `inspect` callback (the only source that still carries the triggering event), replacing `service.onEvent(…)`. The root actor is identified by `actorRef.sessionId === rootId` — self-contained within the inspection event, so it works even while inspection fires during startup before the `service` binding is initialized. Events are narrowed through the existing `isDeployEvent` guard.
- **`interpret(…)` → `createActor(…, { inspect })`.** The actor is started explicitly before events are sent.
- `waitFor` is imported from `xstate` and its predicate uses `snapshot.status === "done"`; the exit code is read from `snapshot.output`; `service.send("X")` calls became `service.send({ type: "X" })`.

#### Tests

- `deploy-machine.test.ts` was migrated to match: `interpret().onTransition` → `createActor().subscribe`, `withConfig({ services })` → `provide({ actors })`, `machine.transition` → `getNextSnapshot`, and the `state.event` assertions rebuilt on the inspection API.
- A `deploy` suite was added to `terraform-cli.test.ts` covering the previously-untested `deploy()`/`handleService` path: `running` is reported even when the process exits immediately (the subscribe-ordering fix), a zero exit resolves without cancellation, and a non-zero exit rejects with the exit code (the final-state-output path).

`tsc --noEmit` passes cleanly on the package, and all 28 tests across `deploy-machine.test.ts` and `terraform-cli.test.ts` pass.

### Checklist

### DIFF
--- a/packages/@cdktn/cli-core/package.json
+++ b/packages/@cdktn/cli-core/package.json
@@ -57,7 +57,7 @@
     "strip-ansi": "6.0.1",
     "undici": "8.7.0",
     "xml-js": "1.6.11",
-    "xstate": "4.38.3",
+    "xstate": "5.32.4",
     "zod": "4.4.3"
   },
   "lint-staged": {
--- a/packages/@cdktn/cli-core/src/lib/models/deploy-machine.ts
+++ b/packages/@cdktn/cli-core/src/lib/models/deploy-machine.ts
@@ -4,13 +4,13 @@
  */
 
 import {
-  createMachine,
-  send,
-  interpret,
-  EventObject,
+  setup,
+  createActor,
+  sendTo,
+  fromCallback,
   assign,
-  Sender,
-  Receiver,
+  EventObject,
+  InspectionEvent,
 } from "xstate";
 import { Errors, logger } from "@cdktn/commons";
 import { missingVariable } from "../errors";
@@ -53,40 +53,6 @@ export function isDeployEvent<DeployEventType extends DeployEvent["type"]>(
 ): event is DeployEvent & { type: DeployEventType } {
   return event.type === type;
 }
-
-export type DeployState =
-  | {
-      value: "idle";
-      context: DeployContext;
-    }
-  | {
-      value: "running";
-      context: DeployContext;
-    }
-  | {
-      value: { running: "processing" };
-      context: DeployContext;
-    }
-  | {
-      value: { running: "awaiting_approval" };
-      context: DeployContext;
-    }
-  | {
-      value: { running: "awaiting_sentinel_override" };
-      context: DeployContext;
-    }
-  | {
-      value: { running: "stopping" };
-      context: DeployContext;
-    }
-  | {
-      value: "exited";
-      context: DeployContext & { exitCode: number };
-    }
-  | {
-      value: "stopped";
-      context: DeployContext;
-    };
 
 /**
  * Terraform's interactive prompt for a missing variable looks like:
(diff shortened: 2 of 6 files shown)


sakul-learning:
I validated a non-blocking ordering regression in `TerraformCli.handleService()`: `createAndStartDeployService()` starts the XState actor and sends `START` before `handleService()` attaches `service.subscribe()`. A silent process that exits immediately can therefore miss the `running` callback (`[]` rather than `["running"]`); ordinary Terraform output produces a later snapshot that reports it, so I **would not block this XState update on that behavior**.

### What the repro shows

`createAndStartDeployService()` starts the actor and sends `START` before `handleService()` subscribes to snapshots. A silent process that exits immediately therefore never reports the initial `running` callback:

```ts
jest.mocked(spawnInteractive).mockReturnValue({
  actions: {
    write: jest.fn(),
    writeLine: jest.fn(),
    stop: jest.fn(),
  },
  exitCode: Promise.resolve(0),
});

const states: string[] = [];
await cli.deploy({}, (state) => states.push(state.type));

expect(states).toEqual(["running"]);
```

On this PR head, the assertion fails with `received: []`.

This is technically an observable behavior regression from v4, but normal Terraform apply/destroy invocations emit stdout/stderr. Those events produce a later snapshot while the machine is still in `running.processing`; because `previousState` is still `idle`, that later snapshot triggers the `running` callback and masks the missed initial transition. I also verified that output emitted during actor startup is observed, and that an immediate non-zero exit is still propagated correctly.

So I see this as a **non-blocking lifecycle hardening opportunity**, rather than evidence of a current customer-facing failure. The practical difference for a real Terraform run is that `running` is reported on its first output instead of immediately on entry. The silent/immediate process is useful for exposing the ordering, but is not representative enough to justify blocking the dependency update by itself.

### Minimal fix validated locally

Processing the actor's current snapshot once after subscribing restores the initial callback without changing the factory API:

```ts
const handleSnapshot = (
  snapshot: ReturnType<typeof service.getSnapshot>,
) => {
  // existing transition handling
};

service.subscribe(handleSnapshot);

// The actor was already started by createService, so consume the
// snapshot whose transition may have preceded the subscription.
handleSnapshot(service.getSnapshot());
```

I pushed the TDD sequence as two commits:

1. RED — repro test: https://github.com/sakul-learning/cdk-terrain/commit/8bcd8279a7aee622341484a37fd2f6346a68f9c0
2. GREEN — minimal fix: https://github.com/sakul-learning/cdk-terrain/commit/e4f72c3594a79129a44475bc8938b08065e20562

Branch: https://github.com/sakul-learning/cdk-terrain/tree/review/pr-333-xstate-running-repro

Validation:

- RED commit: focused suite fails exactly one test, `reports running before a silent process exits` (`1 failed, 5 passed`).
- GREEN commit: focused suite passes (`6/6`).
- Full `@cdktn/cli-core` suite with the fix: `163 passed`; 30 existing dist-dependent tests skipped.
- Prettier and `git diff --check`: pass.

Given the realistic execution path and green existing CI, this is a suggestion rather than a requested change.

so0k (approved):
Up to you if you want to address the non-block ordering concern (which may not be a real failure mode) - still need to rebase for conflict resolution

sakul-learning:
## Review: ordering and exit-code follow-up

`TerraformCli.handleService()` receives an already-started deploy/destroy actor from `createAndStartDeployService()` / `createAndStartDestroyService()`. Before this commit, those factories sent `START` before `handleService()` installed `service.subscribe(...)`, so a silent process that completed immediately could miss the initial `idle → running.processing` notification. The observable symptom was that the deploy callback received no `running` state for that path.

This commit fixes that ordering gap by subscribing and then processing `service.getSnapshot()`. The new `TerraformCli.deploy()` regression test uses a silent, immediately completing PTY and verifies that `running` is reported. It also moves Terraform's exit code from inspection-event capture to the machine's final snapshot output, with focused tests for successful and non-zero exits.

I independently re-reviewed the new head and found the fix correct and appropriately scoped. The added tests are behavioral, deterministic, and run in the affected Terraform unit-test CI jobs.

Validation run locally with the configured review environment:

- `pnpm exec jest packages/cdktn/test/validations.test.ts --runInBand` — 41/41 tests, 11 snapshots passed
- `pnpm exec nx build cdktn` — passed
- `pnpm exec nx test cdktn --runInBand` — 43/43 suites, 542/542 tests, 301 snapshots passed

The completed GitHub build, quality, and affected-unit jobs are also green; the remaining matrix jobs were pending when checked. No workflow or security-sensitive changes are included.

**Verdict: approve.**

PR #362, opened 2026-07-30
feat(lib)!: validate Terraform function versions by default

### LINKED ISSUE #342: Remove validateFunctionVersions feature flag before v0.24 release (graduate to always-on)
## Context

PR #296 (function version validation) introduced the `validateFunctionVersions` feature flag to gate `ValidateFunctionVersionSupport` validation behind an opt-in context key. This validation checks that Terraform functions used through `Fn.*` are supported by the declared `targetVersions` at synth time.

The flag was intentionally behind a feature flag to avoid breaking existing projects in a minor release. New projects created via `cdktn init` automatically get the flag enabled through `FUTURE_FLAGS`.

## Why now

The **v0.23 → v0.24** release is a pre-v1 breaking-change-equivalent version bump. Per the feature flag lifecycle in CONTRIBUTING.md, v0.24 is the correct window to graduate the flag to always-on. If we miss this window, the flag stays gated until v0.25.

**Must land BEFORE the v0.24 release is cut.** Depends on PR #303.

## What needs to change

1. **Always register validation** — `terraform-stack.ts:134`: remove the `if (this.node.tryGetContext(VALIDATE_FUNCTION_VERSIONS))` gate, always call `this.node.addValidation(new ValidateFunctionVersionSupport(this))`
2. **Remove from `FUTURE_FLAGS`** — `features.ts:43`: drop `VALIDATE_FUNCTION_VERSIONS` from the `FUTURE_FLAGS` map
3. **Stop injecting into new projects** — `cli-core/src/lib/init.ts`: `FUTURE_FLAGS` iteration will naturally skip it
4. **Update tests** — `validations.test.ts`: tests that pass `[VALIDATE_FUNCTION_VERSIONS]: "true"` need updating; default (no-flag) behavior becomes the validated behavior
5. **Update integration test config** — `test/typescript/function-version-validation/cdktf.json`: drop the flag from context

## Acceptance criteria

- [ ] `ValidateFunctionVersionSupport` runs on every `TerraformStack` regardless of context
- [ ] `validateFunctionVersions` is removed from `FUTURE_FLAGS`
- [ ] `cdktn init` no longer generates the flag in `cdktf.json`
- [ ] All tests pass with validation as the default behavior
- [ ] Tests cover both always-on validation and the no-targeted-functions path

Flagged during review of PR #296 by @so0k. /cc @jsteinich

### LINKED ISSUE #352: Enable Terraform function version validation by default (drop the validateTerraformFunctionVersions gate) in 0.24
#296 gates `ValidateFunctionVersionSupport` behind the pre-existing `validateTerraformFunctionVersions` context flag, while the new `ValidateProviderFunctionTargetSupport` and write-only target validation are always on (no feature flag).

Since the upcoming release is a 0.23 -> 0.24 bump (pre-v1 "breaking change" release), a follow-up PR should enable function version validation by default and drop the gate, so the `Fn.*` baseline validation and the provider-feature validations behave consistently.

Raised in #296 review: https://github.com/open-constructs/cdk-terrain/pull/296#discussion_r3611050821

so0k:
Closes #352
Closes #342

Graduates the `validateFunctionVersions` feature flag to always-on so it rides the 0.24 breaking-change release.

## Approach: delete the flag, no no-op shim

Following this repo's one existing graduation precedent, `3c323e76d` ("chore: remove existing feature flags", which graduated `EXCLUDE_STACK_ID_FROM_LOGICAL_IDS` and `ALLOW_SEP_CHARS_IN_LOGICAL_IDS").

"Keep it accepted but ignored" is already what deletion gives us: `cdktf.json`'s `context` is typed `readonly context?: { [key: string]: any }` (`packages/@cdktn/commons/src/config.ts:296`) and passed straight through as `CDKTF_CONTEXT_JSON` with no schema, allowlist, or unknown-key check anywhere in the CLI. So an existing `cdktf.json` that still sets `"validateFunctionVersions": "true"` keeps working — the key just becomes inert. There is no hard-fail to protect against, so a deliberate shim would be dead code (YAGNI > KISS).

**No `.jsii` API diff.** `features.ts` is not re-exported from `packages/cdktn/src/index.ts`, so `VALIDATE_FUNCTION_VERSIONS` was never part of the public/JSII surface. The only in-repo deep importer is `packages/@cdktn/cli-core/src/lib/init.ts`, which imports `FUTURE_FLAGS` (kept), not the constant.

## Changes

- `packages/cdktn/src/features.ts` — drop the constant, its JSDoc, and its `FUTURE_FLAGS` entry. The map keeps `failOnConstructsOutsideOfStacks` and `canonicalAssetHashes`.
- `packages/cdktn/src/terraform-stack.ts` — register `ValidateFunctionVersionSupport` unconditionally, alongside `ValidateProviderPresence` and `ValidateProviderFunctionTargetSupport`.
- `test/typescript/function-version-validation/cdktf.json` — remove the key (and the now-empty `context` object), so the integration fixture becomes the end-to-end proof that validation runs with zero context.
- `tools/generate-function-bindings/function-availability/README.md` — name the validation class instead of the removed context key.
- `packages/cdktn/src/validations/validate-provider-function-target-support.ts` — its JSDoc said "Registered unconditionally (no feature flag)", a contrast drawn against a flag that no longer exists.

`cdktn init` stops emitting the key automatically — `init.ts` iterates `Object.entries(FUTURE_FLAGS)` generically, as does `Testing.enableFutureFlags`.

## Why the test diff is larger than a find/replace

Six tests in the `ValidateFunctionVersionSupport` block registered the validation **manually** on a resource, precisely because the constructor gate was off by default. With the gate always-on, every `TerraformStack` also registers a stack-scoped instance, and both read the same `stack._getUsedFunctions()` — so each error would be reported **twice**, under `[MyStack/testResource]` and `[MyStack]`.

Dropping those manual registrations makes the tests exercise the real always-on path; the inline snapshots then re-prefix to `[MyStack]`. The pre-existing interleaved-Apps test, which already relied on automatic registration, confirms `[MyStack]` is the correct prefix.

Two tests are repurposed to cover migration instead: a stale truthy key is inert, and a falsy one no longer opts out.

## Verification

`pnpm nx test cdktn --skip-nx-cache -- --ci`, compared against untouched `origin/main` in a separate worktree:

| | this branch | `origin/main` |
|---|---|---|
| Test Suites | 1 failed, 42 passed | 1 failed, 42 passed |
| Tests | 1 failed, 541 passed | 1 failed, 541 passed |
| Snapshots | 300 passed | 299 passed |

Identical. The one failure is **pre-existing** — `matchers › toPlanSuccessfully › succeeds if the terraform config passed is valid`, which needs a real `terraform` binary. The extra snapshot is this branch's new assertion. `--skip-nx-cache` on every run so a cached green could not stand in for real work.

**Not verified:** `pnpm integration:single typescript/function-version-validation` was not run — it requires `pnpm package`, which needs `mvn` and `dotnet`, neither available on the machine used. CI should cover it. `@cdktn/cli-core` tests also could not run locally (`hcl2json`'s prebuild needs `go` on `PATH`).

## Note on #352's title

#352 names `validateTerraformFunctionVersions`, which does not exist anywhere in the tree — the real context key is `validateFunctionVersions`. #342 quotes the real code and was used as the source of truth. Flagging it so the wrong key name doesn't propagate into changelogs or migration notes.

## Blast radius

Only 7 functions carry version constraints (`base64gunzip`, `cidrcontains`, `convert`, `ephemeralasnull`, `issensitive`, `templatestring`, `urldecode`), and outside `validations.test.ts` the only usage in the repo is the dedicated integration fixture — so no other test, example, or fixture starts failing when validation becomes unconditional.

### DIFF
--- a/packages/cdktn/src/features.ts
+++ b/packages/cdktn/src/features.ts
@@ -30,15 +30,6 @@ export const FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS =
   "failOnConstructsOutsideOfStacks";
 
 /**
- * When enabled, each TerraformStack validates that the Terraform functions
- * used through `Fn` are supported by the runtime versions declared in
- * `targetVersions` (Terraform and/or OpenTofu). This is a purely static check
- * against the vendored function availability matrix; it never executes a
- * Terraform-compatible CLI.
- */
-export const VALIDATE_FUNCTION_VERSIONS = "validateFunctionVersions";
-
-/**
  * When enabled, TerraformAsset and TerraformModuleAsset compute asset hashes
  * with a canonical entry-framed scheme modeled on git trees and Nix NAR:
  * every entry contributes its type, permission mask (for files and
@@ -58,6 +49,5 @@ export const CANONICAL_ASSET_HASHES = "canonicalAssetHashes";
 
 export const FUTURE_FLAGS = {
   [FAIL_ON_CONSTRUCTS_OUTSIDE_OF_STACKS]: "true",
-  [VALIDATE_FUNCTION_VERSIONS]: "true",
   [CANONICAL_ASSET_HASHES]: "true",
 };
--- a/packages/cdktn/src/terraform-stack.ts
+++ b/packages/cdktn/src/terraform-stack.ts
@@ -20,7 +20,6 @@ import {
   ValidateProviderFunctionTargetSupport,
   ValidateProviderPresence,
 } from "./validations";
-import { VALIDATE_FUNCTION_VERSIONS } from "./features";
 import { App } from "./app";
 import { TerraformBackend } from "./terraform-backend";
 import { TerraformResourceTargets } from "./terraform-resource-targets";
@@ -131,9 +130,7 @@ export class TerraformStack extends Construct {
     );
     Object.defineProperty(this, STACK_SYMBOL, { value: true });
     this.node.addValidation(new ValidateProviderPresence(this));
-    if (this.node.tryGetContext(VALIDATE_FUNCTION_VERSIONS)) {
-      this.node.addValidation(new ValidateFunctionVersionSupport(this));
-    }
+    this.node.addValidation(new ValidateFunctionVersionSupport(this));
     this.node.addValidation(new ValidateProviderFunctionTargetSupport(this));
   }
 
--- a/packages/cdktn/src/validations/validate-provider-function-target-support.ts
+++ b/packages/cdktn/src/validations/validate-provider-function-target-support.ts
@@ -21,8 +21,8 @@ import { TerraformStack } from "../terraform-stack";
  * provider function has been used, and otherwise checked once, naming every
  * used function so the error is actionable.
  *
- * Registered unconditionally (no feature flag): this is new API surface, and
- * the check only ever fires when a provider function is actually used —
+ * Registered unconditionally: this is new API surface, and the check only
+ * ever fires when a provider function is actually used —
  * i.e. its token was resolved while rendering this validation's OWNING
  * STACK (`TerraformStack._getUsedProviderFunctions()`), not while rendering
  * some other stack in the same App: this validation is registered in
--- a/packages/cdktn/test/validations.test.ts
+++ b/packages/cdktn/test/validations.test.ts
@@ -11,12 +11,10 @@ import {
   resolveTargetVersions,
   ValidateBinaryVersion,
   ValidateFeatureTargetSupport,
-  ValidateFunctionVersionSupport,
 } from "../src/validations";
 import { TestProvider } from "./helper/provider";
 import { createTmpHelper } from "./helper/tmp";
 import { terraformBinaryName } from "../src/util";
-import { VALIDATE_FUNCTION_VERSIONS } from "../src/features";
 
 const tmp = createTmpHelper();
 
@@ -492,10 +490,7 @@ describe("ValidateFunctionVersionSupport", () => {
   }
 
   test("ignores universally available functions without resolving targets", () => {
-    const { app, stack, testResource } = appWithStack();
-    testResource.node.addValidation(
-      new ValidateFunctionVersionSupport(testResource),
-    );
+    const { app, stack } = appWithStack();
     new TestResource(stack, "usesBaselineFunction", {
       name: Fn.abs(-42).toString(),
     });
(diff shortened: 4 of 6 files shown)


sakul-learning:
## PR #362 Review — **APPROVE**

Reviewed latest head `fcd7178e821e471f1a304cd77ca70e80110cfd09` against `origin/main`. I also read the PR body and complete public comment/review thread; there are currently no comments or prior reviews.

### Findings

No blocking correctness, security, workflow, or maintainability issues found.

- `TerraformStack` now registers `ValidateFunctionVersionSupport` unconditionally.
- Removing `validateFunctionVersions` from `FUTURE_FLAGS` also removes it from newly generated project configuration.
- Existing configurations containing the legacy key remain accepted because context is an untyped pass-through map; `false` no longer disables validation, as intended.
- Changed tests now exercise production stack-level registration rather than manually attaching the validator.
- Cross-App isolation, interleaved Apps, independent sibling-stack targets, compatible targets, unsupported functions, and malformed target declarations remain covered.
- No workflow files changed, so there are no applicable zizmor findings.
- `git diff --check` passed.
- PR title and breaking-change commit metadata follow repository conventions.

### Configured checks

Executed with the supplied `TMPDIR` and `PATH`:

1. `pnpm exec jest packages/cdktn/test/validations.test.ts --runInBand`
   - **PASS**
   - 1/1 suite
   - 41/41 tests
   - 11/11 snapshots

2. `pnpm exec nx build cdktn`
   - **PASS**
   - jsii: 0 errors, 0 warnings
   - Nx reported the build target as flaky, but the build completed successfully.

3. `pnpm exec nx test cdktn --runInBand`
   - **PASS**
   - 43/43 suites
   - 542/542 tests
   - 301/301 snapshots
   - Two existing CDKTN escape-sequence warnings were emitted; neither is related to this PR.

GitHub’s unit, build, lint, formatting, knip, CodeQL, and analysis checks were also passing when checked. Integration preparation jobs were still pending.

### Test and artifact-value pass

The changed unit tests run through the normal `cdktn` Jest/Nx target and ordinary PR unit CI. The integration fixture is part of the normal TypeScript integration suite. None of the changed tests are skipped, environment-gated, static-string-only, or manual-only.

**Ponytail artifact-value pass: Lean already. Ship.**

jsteinich (approved):

PR #363, opened 2026-08-01
fix(lib): Disallow constructs 10.8 until support can be added

jsteinich:
### Related issue

Fixes https://github.com/open-constructs/cdk-terrain/actions/runs/30545326461/job/90912362214

### Description

constructs 10.8.0 moved to the jsii 6.0.7 compiler and dropped jsii.tsc: {outDir: "lib", rootDir: "src"} from its package.json (replaced by jsii.tsconfig/validateTsconfig).

In jsii/lib/common/symbol-id.js, findPackageInfo reads tscOutDir only from jsii.tsc.outDir in package.json. tscRootDir has an assembly-metadata fallback (asm.metadata.tscRootDir); tscOutDir does not, and no tsconfig.json ships in the tarball. So normalizePath bails, the computed symbolId stays lib/construct:Construct instead of src/construct:Construct, misses the assembly's symbolIdMap, and importedSymbol comes back undefined. assemblies.js:234 swallows this in a bare catch {}/undefined path, so you get silently wrong output instead of an error.
See https://github.com/aws/jsii-compiler/issues/2740 for more details.

Needed to additionally quote the installed dependency in the template so that shell wouldn't mangle the updated version range.

We could skip the peer dependency constraint, but it's clear that cdktn isn't fully compatible with constructs 10.8 at this time. We can revisit when we update to JSII 6.x.

### Checklist

### DIFF
--- a/packages/@cdktn/cli-core/templates/typescript/.hooks.sscaff.js
+++ b/packages/@cdktn/cli-core/templates/typescript/.hooks.sscaff.js
@@ -28,7 +28,13 @@ exports.post = (ctx) => {
     throw new Error(`missing context "npm_cdktf"`);
   }
 
-  installDeps([npm_cdktf, `constructs@10`], false, silent);
+  // Mirrors the `constructs` peer dependency range declared by the cdktn
+  // package. 10.8.0 is excluded because it dropped `jsii.tsc.outDir` from its
+  // package.json, which jsii-rosetta needs to map the shipped .d.ts files back
+  // to the symbol ids in the assembly; without it `cdktn convert` emits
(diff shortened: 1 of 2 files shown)


so0k (approved):

sakul-learning (approved):
Approved. I built the PR head and verified `cdktn init --template java` using the locally packaged `cdktn` artifact; initialization, `./gradlew install`, and `./gradlew test` passed. Local testing briefly hit a `semver@7.8.5` resolution issue in `pack.mjs`, resolved with a clean npm cache/staging directory and apparently unrelated to this PR.

PR #364, opened 2026-08-01
fix(tools): make the workspace build, package, lint and test on Windows

jsteinich:
Restores a working Windows dev loop: `pnpm build`, `pnpm package-windows`,
`pnpm lint:workspace` and `pnpm test` all pass.

The scope grew beyond the initial "fix the build" because the first fix
uncovered the next: with line endings corrected the build reached the
packaging step, packaging reached the test suite, and the test suite exposed
several genuine bugs rather than snapshot drift. Each commit is scoped to one
cause so they can be reviewed — or reverted — independently.

**Not all of this is Windows-only.** Four changes fix defects that affect
every contributor; they are marked ⚠ below.

## Root cause: line endings

`core.autocrlf=true` checked out 1729 of 1770 tracked files as CRLF while every
blob in the repo is stored as LF. That alone broke the build:

- nx splits the two-document `pnpm-lock.yaml` on a literal `"---\n"`. With CRLF
  it fails to split, hands both documents to the YAML parser, and aborts with
  *"expected a single document in the stream"* — no project graph, no build.
- All 25 build scripts are `#!/bin/bash` and fail on CRLF.

`.gitattributes` now pins the working tree to LF. Batch files use `-text`
rather than `text eol=crlf`: 8 of the 9 vendored `gradlew.bat` files are stored
with CRLF blobs, and `eol=crlf` would rewrite those blobs to LF on the next
`git add`, silently changing files that need no change.

This is also why a stray reformat of a JSDoc comment appeared to fix
`cdktn:package` — under CRLF the `\r` survived into a C# `[Obsolete("...")]`
string literal and `dotnet build` failed with `CS1010: Newline in constant`.
With the tree normalized there is nothing to work around, so no source change
was needed.

## Production fixes

| Change | Why it matters |
| --- | --- |
| ⚠ `terraform-cloud-client.ts` composes URLs with `path.posix.join` | It used `path.join`, yielding `https:\app.terraform.io\...` and `TypeError: Invalid URL`. **Every Terraform Cloud/Enterprise call was broken on Windows**, not just the test that caught it. |
| `terraform-module-asset.ts` emits `/`-separated module sources | Terraform local module sources must use forward slashes. cdktn emitted `source = "./assets\...\module"`, which Terraform cannot resolve — the synthesized configuration was unusable. |
| `manifest.ts` records manifest paths with `path.posix.join` | `manifest.json` now reads identically on every platform. Consumers join these onto `outdir`, and Node accepts forward slashes on Windows, so nothing on the reading side changes. |
| ⚠ `constructs-maker.ts` accepts either `npm pack --json` shape | npm 12 changed the output from an array to an object keyed by package name, so binding generation failed with *"Cannot read properties of undefined"*. Affects anyone on npm ≥ 12, on any OS. |

## Build and packaging

- **Package scripts** — cmd.exe does not strip single quotes, so
  `-p 'cdktn*' '@cdktn/*'` matched no projects and nx exited **0** reporting
  "No tasks were run": a silent no-op, not an error. Switched to double quotes,
  which both cmd.exe and POSIX sh strip. `./package.sh` and friends are now
  invoked as `bash <script>` (all are already `#!/bin/bash`, so behaviour on
  Linux/macOS is unchanged).
- **`.pack-staging`** — `pack.mjs` only deleted the staging tree at the *start*
  of a run, so it was always left behind. It holds its own `package.json` and a
  nested `node_modules`; when `npm pack` descended into a stale one it aborted
  with the opaque *"Exit handler never called!"*. Now excluded via `.npmignore`,
  cleaned up after use, and listed in `.nxignore` because the nx daemon's file
  watcher held handles on it and made removal fail with EPERM.
- **npm/pnpm spawning** — Node refuses to spawn `.cmd` shims directly
  (CVE-2024-27980), surfacing as `spawnSync npm.cmd EINVAL`.
- ⚠ **`collect-dist.mjs`** — replaced `pnpm --recursive exec pwd` (no `pwd` on
  Windows, and pnpm is a `.cmd` shim) with a JSON query. This also fixes a
  latent bug: the JSON list includes the workspace root, whose `dist/` is the
  collection destination, so it is now excluded explicitly.

## Lint

- ⚠ **`lint-examples.mjs`** stripped the repo root with
  `p.replace(REPO_ROOT + "/", "")`. On Windows the needle never matched, so
  every example's expected name became `@examples/C:\...\examples\csharp\aws`.
  It also carried a broken regex elsewhere (`/\s*/` matches the empty string and
  has no global flag). This blocked `pnpm lint:examples` *and* any commit
  touching `examples/`, via lint-staged.
- **`@nx/enforce-module-boundaries` disabled for `test/`** — `test/` is a
  standalone pnpm project, so its `node_modules` sits inside its own project
  root. nx maps files to projects by root containment, so every bare import
  resolves back to the importing project and the rule crashes on this repo's
  missing root tsconfig. Adding a root tsconfig is *not* the fix: with one
  present the rule instead reports five false positives telling you to replace
  the npm package `has-ansi` with a relative path. Reproduces for anyone who has
  run the integration tests; CI never installs `test/node_modules`.

## Tests

Unit tests went from **167 failures to 0** on Windows.

- **hcl2cdk (111 of those)** had a single cause: the harness linked a shared
  `node_modules` into each temp project with symlink type `"dir"`, which needs
  Developer Mode on Windows and otherwise fails with EPERM. A **junction** is
  created without elevation and resolves identically for reads, so these tests
  now genuinely run rather than being skipped.
- **Capability gating, not platform gating** — `test/helper/capabilities.ts`
  probes for symlink creation, `unzip` and POSIX permission bits. A Windows
  developer with Developer Mode enabled gets that coverage automatically
  instead of a permanently dead suite. The `unzip` tests are skipped rather than
  ported to fflate on purpose: they exist to validate fflate-written archives
  against an *independent* extractor, so extracting with fflate would make the
  assertion circular.
- **Path assertions** were adjusted rather than the code they assert on. In
  every case the production value is correct — these are real filesystem paths,
  and native separators are what callers want.
- **Two harness bugs** unrelated to paths: a test name containing `*` became a
  directory name (illegal on Windows), and the `convert` tests asserted empty
  stderr while nx sets `FORCE_COLOR` and the CLI sets `NO_COLOR`, making Node
  warn about the conflict on stderr.
- **`parallel-error` fixture** staggers three stacks with `sleep` in a
  `local-exec` provisioner. Terraform runs those through cmd.exe on Windows,
  which has no `sleep`, so all three stacks failed instead of only the one the
  test expects to.

## Windows prerequisites

Now documented in `CONTRIBUTING.md`. Both fail in ways that name neither the
tool nor the cause, which is why they were worth writing down:

- **Git Bash on `PATH`**, ahead of the WSL stub — package scripts invoke
  `bash <script>.sh`. The `bash.exe` in `WindowsApps` is the WSL launcher and
  cannot see the Windows Node toolchain. `C:\Program Files\Git\bin` contains
  only `bash`/`sh`/`git`, so it will not shadow Windows `find`/`sort` the way
  `Git\usr\bin` would.
- **A running Docker daemon** for the `cdktn` matcher tests, which run a real
  `terraform plan` against the docker provider. Without it they fail as a bare
  `expect(received).toBeTruthy()`. Docker was previously mentioned only as an
  optional containerized dev environment, so nothing connected the failure to
  its cause.

The same section records how to renormalize a clone created before
`.gitattributes` pinned the working tree to LF.

## Verification

On Windows: `pnpm build`, `pnpm package-windows` (all five language targets),
`pnpm lint:workspace` (35/35 projects) and `pnpm test` all pass. Test runs were
uncached (`--skip-nx-cache`). No behaviour change on Linux/macOS is intended —
every platform-specific branch is a no-op there, and the `posix.join` switches
reproduce the previous POSIX output exactly.


sakul-learning:
`CONTRIBUTING.md`s Windows line-ending recovery steps tell an existing checkout to run `git reset --hard`. That command discards tracked, uncommitted changes, so a contributor discovering this after local edits can lose work. Please add a clean-tree/stash-or-commit warning, or recommend a fresh clone/worktree before the reset.

CONTRIBUTING.md
```diff
@@ -121,6 +121,35 @@ To build and install `cdk-terrain` locally you need to install:
...
+- An LF working tree. `.gitattributes` pins this, so a fresh clone is correct.
+  A clone made earlier with `core.autocrlf=true` will still have CRLF files,
+  which breaks nx's lockfile parsing and every `#!/bin/bash` script. Renormalize
+  an existing clone with:
+
+  ```shell
+  $ git config core.autocrlf false
+  $ git rm --cached -r .
+  $ git reset --hard
+  ```
```
so0k, line 162: Please add a clean-tree/stash-or-commit warning, or recommend a fresh clone/worktree before the reset.

jsteinich: added

packages/cdktn/test/archive-symlink.test.ts
```diff
@@ -225,20 +239,23 @@ describe("archiveSync zip metadata (#320 follow-ups)", () => {
     fs.rmSync(destDir, { recursive: true, force: true });
   });
 
-  test("preserves the executable bit on regular files", () => {
-    const script = path.join(srcDir, "run.sh");
-    fs.writeFileSync(script, "#!/bin/sh\necho hi\n");
-    fs.chmodSync(script, 0o755);
+  testIfPosixPermissions(
```
sakul-learning: `packages/cdktn/test/archive-symlink.test.ts` registers “preserves the executable bit on regular files” with `testIfPosixPermissions`, but the test calls `extractZip()`, which executes the external `unzip` command. On a POSIX host without `unzip`, this test still runs and fails despite the new capability gating. Please gate it on both POSIX permissions and `hasUnzip`, for example with a combined helper.

so0k (approved):

PR #371, opened 2026-08-06
feat(lib)!: add asset pipeline foundation types and hash utilities

eduardomourar:
### Related issue

Spin-off from #339

### Description
Foundational layer for the asset pipeline, per the #380 scope discussion:
- `IAsset`, `AssetOptions`, `AssetHashType` — identity types. `TerraformAsset` now `implements IAsset` and consumes `AssetHashType`: `SOURCE` (default) hashes the source, `CUSTOM` uses an explicit `assetHash`, `OUTPUT` throws until bundling exists.
- `IAssetPackaging` + `AssetPackaging.FILE/DIRECTORY/ZIP`, with `pack()` taking a `PackOptions` struct (`source`, `target`, `ignoreStrategy?`) so it stays extensible without a breaking change. `TerraformAsset` packages through `AssetPackaging` and derives its on-disk layout and artifact name from `producesDirectory`/`extension` rather than a local switch.
- `StagedAsset` (`assetHash`, `path`, `isDirectory`) — publisher input, replaces the earlier `FileAssetSource`.
- `IIgnoreStrategy` + `ExcludeIgnoreStrategy`, and `AssetHash.of()` for content identity without staging. `ignores()` takes an `IgnoreQuery` (`relativePath`, `isDirectory`) so `.gitignore`/`.dockerignore`-style directory-only patterns are expressible; `pruneExcludedDirectories` lets a strategy opt into descending excluded directories for negation patterns; optional `cacheKey` supports future result caching.
- `archiveSync` gained an exclude predicate to match `copySync`, so packaging and hashing agree on what's excluded.
- `AssetHash.of()` pins to the canonical hash scheme, resolves relative paths against `cdktf.json` (matching `TerraformAsset`), and throws if both `exclude` and `ignoreStrategy` are passed.

Held back per #380: Docker/container asset types (`DockerImageAssetSource`, `DockerImageAssetLocation`, `DockerCacheOption`) and `FileAssetLocation` — these model synth-resolved locations that don't fit the publish-at-apply model this pipeline is built on.

This is the foundation PR (1 of 3) for the asset staging pipeline. Later PRs will add `AssetStaging`, `TerraformAsset` integration, and Docker bundling on top of these primitives.

### Breaking changes
- `TerraformAsset.assetHash` is now `readonly` (required by `implements IAsset`). The generated setter is removed in Java, C#, Python, and Go. It was only ever assigned at construction, so this removes an assignment that would have desynced `assetHash` from `path`.

### Checklist

### DIFF
--- a/packages/cdktn/src/asset-hash.ts
+++ b/packages/cdktn/src/asset-hash.ts
@@ -1,0 +1,131 @@
+// Copyright (c) HashiCorp, Inc
+// SPDX-License-Identifier: MPL-2.0
+import * as crypto from "crypto";
+import * as path from "path";
+import { hashPath, findFileAboveCwd } from "./private/fs";
+import { ExcludeIgnoreStrategy, IIgnoreStrategy } from "./ignore-strategy";
+import {
+  assetHashConflictingExcludeOptions,
+  assetHashOutOfScopeCdktnJson,
+} from "./errors";
+
+/**
+ * Options for {@link AssetHash.of}.
+ */
+export interface AssetHashOptions {
+  /**
+   * Paths to exclude, relative to the hashed path. Cannot be combined with
+   * `ignoreStrategy`, which replaces this matcher rather than layering on
+   * top of it.
+   *
+   * @default - nothing is excluded
+   */
+  readonly exclude?: string[];
+
+  /**
+   * Extra information to fold into the hash.
+   *
+   * @default - no extra data
+   */
+  readonly extraHash?: string;
+
+  /**
+   * Exclusion matching, for callers that need `.gitignore` / `.dockerignore`
+   * parity rather than the built-in exact-path / suffix / directory matcher.
+   *
+   * @default - `exclude` is used with the built-in matcher
+   */
+  readonly ignoreStrategy?: IIgnoreStrategy;
+}
+
+/**
+ * Computes a content hash of a file or directory without staging it.
+ *
+ * Providers that read a local path directly — a Docker build `context`, for
+ * example — need a content hash to drive `triggers`, but have no use for a
+ * staged copy of the source. `Asset` and `TerraformAsset` always produce a
+ * staged copy; this is the identity half without the staging half.
+ */
+export class AssetHash {
+  /**
+   * Content hash of a file or directory, without staging it.
+   *
+   * This is a hash of the source tree, and it does not depend on how the
+   * source is later packaged. `hashPath` is always called with `archive`
+   * unset, so directory records are part of the digest; the hash of a given
+   * tree is therefore the same whether it is later copied, zipped, or packed
+   * by a custom `IAssetPackaging` such as `tar.bz2`.
+   *
+   * A consequence worth stating: this does not equal
+   * `TerraformAsset(dir, { type: ARCHIVE }).assetHash` for a directory with
+   * subdirectories. That asset frames its hash to the emitted ZIP, which has
+   * no directory entries (see #323), so directory-only changes move it and
+   * not this. It equals a `FILE` / `DIRECTORY` `TerraformAsset` hash only
+   * when the `canonicalAssetHashes` flag is enabled, since that flag is what
+   * puts the asset on this same canonical scheme.
+   *
+   * A relative `filePath` is resolved against the directory containing
+   * `cdktf.json`, the same base `TerraformAsset` uses, so both hash the same
+   * source regardless of the process working directory. Absolute paths are
+   * used as-is. Throws if `filePath` is relative and no `cdktf.json` is found
+   * above the working directory.
+   * @param filePath - path to a file or directory to hash
+   * @param options - see {@link AssetHashOptions}
+   */
+  public static of(filePath: string, options: AssetHashOptions = {}): string {
+    if (options.exclude?.length && options.ignoreStrategy) {
+      throw assetHashConflictingExcludeOptions();
+    }
+
+    const resolved = AssetHash.resolvePath(filePath);
+    const strategy =
+      options.ignoreStrategy ??
+      new ExcludeIgnoreStrategy(options.exclude ?? []);
+
+    // Pinned to the canonical scheme: this is a brand-new API with no
+    // existing hashes to preserve, so it has no reason to start on the
+    // legacy scheme that `canonicalAssetHashes` exists to move away from.
+    const baseHash = hashPath(resolved, {
+      canonical: true,
+      shouldExclude: (relativePath, isDirectory) =>
+        strategy.ignores({ relativePath, isDirectory }),
+      descendIntoExcludedDirectories:
+        strategy.pruneExcludedDirectories === false,
+    });
+
+    if (!options.extraHash) {
+      return baseHash;
+    }
+
+    return crypto
+      .createHash("md5")
+      .update(baseHash)
+      .update(options.extraHash)
+      .digest("hex")
+      .slice(0, 32)
+      .toUpperCase();
+  }
+
+  /**
+   * Resolve `filePath` to an absolute path using the same base as
+   * `TerraformAsset`: relative paths are anchored to the directory holding
+   * `cdktf.json`, not `process.cwd()`, so a hash taken here matches the one
+   * `TerraformAsset` computes for the same source even when the app is run
+   * from a subdirectory or a test runner. Unlike `TerraformAsset`, there is
+   * no construct scope to read the `cdktfJsonPath` context from, so this uses
+   * the `findFileAboveCwd` fallback directly.
+   * @param filePath - path passed to {@link AssetHash.of}
+   */
+  private static resolvePath(filePath: string): string {
+    if (path.isAbsolute(filePath)) {
+      return filePath;
+    }
+    const cdktfJsonPath = findFileAboveCwd("cdktf.json");
+    if (!cdktfJsonPath) {
+      throw assetHashOutOfScopeCdktnJson(filePath);
+    }
+    return path.resolve(path.dirname(cdktfJsonPath), filePath);
+  }
+
+  private constructor() {}
+}
--- a/packages/cdktn/src/assets.ts
+++ b/packages/cdktn/src/assets.ts
@@ -1,0 +1,240 @@
+// Copyright (c) HashiCorp, Inc
+// SPDX-License-Identifier: MPL-2.0
+
+import * as fs from "fs";
+import { archiveSync, copySync } from "./private/fs";
+import { IIgnoreStrategy } from "./ignore-strategy";
+
+/**
+ * Common interface for all assets.
+ */
+export interface IAsset {
+  /**
+   * A hash of this asset, which is available at construction time. As this is a plain string, it
+   * can be used in construct IDs in order to enforce creation of a new resource when the content
+   * hash has changed.
+   */
+  readonly assetHash: string;
+}
+
+/**
+ * Asset hash options
+ */
+export interface AssetOptions {
+  /**
+   * Specify a custom hash for this asset. If `assetHashType` is set it must
+   * be set to `AssetHashType.CUSTOM`. The value is used verbatim as the asset
+   * hash, and because it names the staged asset file it may only contain
+   * letters, digits, `_`, `.` and `-`.
+   *
+   * NOTE: the hash is used in order to identify a specific revision of the asset, and
+   * used for optimizing and caching deployment activities related to this asset such as
+   * packaging, uploading to cloud storage, etc. If you chose to customize the hash, you will
+   * need to make sure it is updated every time the asset changes, or otherwise it is
+   * possible that some deployments will not be invalidated.
+   *
+   * @default - based on `assetHashType`
+   */
+  readonly assetHash?: string;
+
+  /**
+   * Specifies the type of hash to calculate for this asset.
+   *
+   * If `assetHash` is configured, this option must be `undefined` or
+   * `AssetHashType.CUSTOM`.
+   *
+   * @default - the default is `AssetHashType.SOURCE`, but if `assetHash` is
+   * explicitly specified this value defaults to `AssetHashType.CUSTOM`.
+   */
+  readonly assetHashType?: AssetHashType;
+}
+
+/**
+ * The type of asset hash
+ *
+ * NOTE: the hash is used in order to identify a specific revision of the asset, and
+ * used for optimizing and caching deployment activities related to this asset such as
+ * packaging, uploading to cloud storage, etc.
+ */
+export enum AssetHashType {
+  /**
+   * Based on the content of the source path
+   *
+   * Use `SOURCE` when the content of the asset changes frequently or when
+   * you want to track changes to the source files directly.
+   */
+  SOURCE = "source",
+
+  /**
+   * Based on the content of the bundling output
+   *
+   * Use `OUTPUT` when the source of the asset is a top level folder containing
+   * code and/or dependencies that are not directly linked to the asset.
+   */
+  OUTPUT = "output",
+
+  /**
+   * Use a custom hash
+   */
+  CUSTOM = "custom",
+}
+
+/**
+ * How a staged asset is produced and what shape it takes on disk.
+ *
+ * Packaging answers two independent questions: how the artifact is produced
+ * (copy, zip, tar.gz, ...) and whether the result is a directory or a single
+ * file. A closed enum can only ever answer the second one, so it is an
+ * interface rather than an enum — custom formats (e.g. `tar.bz2`) need no
+ * core change.
+ */
+export interface IAssetPackaging {
+  /**
+   * Appended to the staged artifact name, e.g. ".zip", "", ".tar.bz2".
+   */
+  readonly extension: string;
+
+  /**
+   * Whether the staged result is a directory rather than a single file.
+   *
+   * Publishers branch on this to decide whether they upload one object or
+   * sync a tree.
+   */
+  readonly producesDirectory: boolean;
+
+  /**
+   * Perform the staging transformation, writing the packaged result to
+   * `options.target`.
+   * @param options - see {@link PackOptions}
+   */
+  pack(options: PackOptions): void;
+}
+
+/**
+ * Options for {@link IAssetPackaging.pack}.
+ *
+ * A struct rather than positional parameters: adding a struct field is
+ * additive, adding a method parameter is not, and `pack` is called through
+ * JSII where that distinction is a breaking-change boundary.
+ */
+export interface PackOptions {
+  /**
+   * Path to the resolved (already bundled, if applicable) source.
+   */
+  readonly source: string;
+
+  /**
+   * Path the packaged result should be written to.
+   */
+  readonly target: string;
+
+  /**
+   * Entries to omit from the packaged result. Must match the strategy used
+   * to hash the same source, or the hash and the artifact describe
+   * different sets of files.
+   *
+   * @default - nothing is excluded
+   */
+  readonly ignoreStrategy?: IIgnoreStrategy;
+}
+
+/**
+ * Copies a single file verbatim.
+ */
+class FilePackaging implements IAssetPackaging {
+  public readonly extension = "";
+  public readonly producesDirectory = false;
+  public pack(options: PackOptions): void {
+    fs.copyFileSync(options.source, options.target);
+  }
+}
+
+/**
+ * Copies a directory tree verbatim, without archiving it.
+ */
+class DirectoryPackaging implements IAssetPackaging {
+  public readonly extension = "";
+  public readonly producesDirectory = true;
+  public pack(options: PackOptions): void {
+    copySync(options.source, options.target, {
+      shouldExclude: options.ignoreStrategy
+        ? (relativePath, isDirectory) =>
+            options.ignoreStrategy!.ignores({ relativePath, isDirectory })
+        : undefined,
+      descendIntoExcludedDirectories:
+        options.ignoreStrategy?.pruneExcludedDirectories === false,
+    });
+  }
+}
+
+/**
+ * Archives a directory tree into a single zip file.
+ */
+class ZipPackaging implements IAssetPackaging {
+  public readonly extension = ".zip";
+  public readonly producesDirectory = false;
+  public pack(options: PackOptions): void {
+    archiveSync(
+      options.source,
+      options.target,
+      options.ignoreStrategy
+        ? (relativePath, isDirectory) =>
+            options.ignoreStrategy!.ignores({ relativePath, isDirectory })
+        : undefined,
+      options.ignoreStrategy?.pruneExcludedDirectories === false,
+    );
+  }
+}
+
+/**
+ * Built-in packaging strategies. Custom formats implement `IAssetPackaging`
+ * directly rather than extending this class.
+ */
+export class AssetPackaging {
+  /**
+   * Copy a single file as-is.
+   */
+  public static readonly FILE: IAssetPackaging = new FilePackaging();
+
+  /**
+   * Copy a directory tree as-is, without archiving.
+   */
+  public static readonly DIRECTORY: IAssetPackaging = new DirectoryPackaging();
+
+  /**
+   * Archive a directory tree into a single zip file.
+   */
+  public static readonly ZIP: IAssetPackaging = new ZipPackaging();
+
+  private constructor() {}
+}
+
+/**
+ * A staged artifact, ready to hand to an `IAssetPublisher`.
+ *
+ * Deliberately narrower than a location: `path` and `isDirectory` are known
+ * once staging runs, at synth time, before anything is published. Where an
+ * asset ends up — a bucket name, an object key, a URL — is resolved at
+ * apply time and belongs on the publisher's own reference type instead.
+ */
+export interface StagedAsset {
+  /**
+   * A hash on the content source. This hash is used to uniquely identify this
+   * asset throughout the system. If this value doesn't change, the asset will
+   * not be rebuilt or republished.
+   */
+  readonly assetHash: string;
+
+  /**
+   * The path to the staged artifact, relative to the stack directory.
+   */
+  readonly path: string;
+
+  /**
+   * Whether the staged artifact is a directory rather than a single file.
+   *
+   * Publishers branch on this to decide whether they upload one object or
+   * sync a tree; see `IAssetPackaging.producesDirectory`.
+   */
+  readonly isDirectory: boolean;
+}
--- a/packages/cdktn/src/ignore-strategy.ts
+++ b/packages/cdktn/src/ignore-strategy.ts
@@ -1,0 +1,103 @@
+// Copyright (c) HashiCorp, Inc
+// SPDX-License-Identifier: MPL-2.0
+import { excludeMatcher } from "./private/fs";
+
+/**
+ * A single entry presented to an {@link IIgnoreStrategy}.
+ *
+ * A struct rather than a positional `relativePath` string so the shape can
+ * evolve. `isDirectory` is carried from the outset because `.gitignore` /
+ * `.dockerignore` semantics turn on it: a `foo/` pattern matches a
+ * directory but not a file named `foo`, and a matcher cannot recover
+ * that distinction from the path text alone.
+ */
+export interface IgnoreQuery {
+  /**
+   * `/`-separated path relative to the asset root.
+   */
+  readonly relativePath: string;
+
+  /**
+   * Whether this entry is a directory rather than a file or symlink.
+   *
+   * The tree walkers set this from `lstat`, so a strategy can honor
+   * directory-only patterns without stat-ing the path itself (which it has
+   * no root to resolve against). Excluding a directory also excludes
+   * everything below it, since the walkers stop descending once a directory
+   * is excluded.
+   */
+  readonly isDirectory: boolean;
+}
+
+/**
+ * Decides whether a path relative to an asset root is excluded from staging
+ * and hashing.
+ *
+ * Core ships only the exact-path / `*.ext` / directory matcher used by
+ * `exclude` today (`ExcludeIgnoreStrategy`). Full glob, `.gitignore`, and
+ * `.dockerignore` parity can be implemented against this interface without
+ * core taking on a glob parser.
+ */
+export interface IIgnoreStrategy {
+  /**
+   * Whether the given entry should be excluded.
+   * @param query - the entry under consideration, see {@link IgnoreQuery}
+   */
+  ignores(query: IgnoreQuery): boolean;
+
+  /**
+   * Whether excluding a directory also excludes everything beneath it.
+   *
+   * When `true` (the default), the walkers stop descending as soon as a
+   * directory is excluded — cheaper, and correct for a strategy whose
+   * patterns never re-include a path below an excluded parent.
+   *
+   * A strategy with negation patterns must set this to `false`:
+   * `.gitignore` / `.dockerignore` allow `node_modules` followed by
+   * `!node_modules/keep`, which is only reachable if the walk descends into
+   * the excluded `node_modules` and asks about `node_modules/keep`. The
+   * excluded directory entry itself is still omitted; only the descent
+   * changes. Opting out costs a full walk of excluded subtrees.
+   *
+   * @default true
+   */
+  readonly pruneExcludedDirectories?: boolean;
+
+  /**
+   * A value identifying this strategy's exclusion behavior, suitable for
+   * folding into a cache key. Two strategies that return the same
+   * `cacheKey` must exclude the same paths.
+   *
+   * Callers such as `AssetStaging`'s result cache key on a JSON-serializable
+   * representation of their inputs, which a strategy instance is not. Omit
+   * this when the strategy's behavior can't be summarized this way; the
+   * caller then has to treat every call as uncacheable.
+   *
+   * @default - this strategy cannot be represented in a cache key
+   */
+  readonly cacheKey?: string;
+}
+
+/**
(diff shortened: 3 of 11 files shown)


so0k:
@sakul-learning — picking up your review threads from https://github.com/open-constructs/cdk-terrain/pull/339#issuecomment-5016343152 here, since the shapes have moved into this slice.

### On temporarily removing `extraHash` / `bundling` / `assetHashType` and their `index.ts` exports

I follow the reasoning, but I'm not sure I agree with pulling them.

The split into stacked PRs is a review-ergonomics device — it exists to keep each diff small enough to review properly, not to create independent release boundaries. The slices are intended to land together, so the enums and option shapes won't reach consumers ahead of the implementation that gives them meaning. The cross-language commitment you describe is real, but it's incurred at release, and there's no release in between.

Worth noting too that your recommendation was explicitly conditional on finishing the unification being too large for #339's scope — and the unified fingerprinting has since landed in `d212d56`, standardising on the existing MD5-derived identity rather than introducing the second SHA-256 walker. So the dual-path compatibility migration you were guarding against shouldn't materialise.

Stating the assumption explicitly so it's on the record rather than implied: if this foundation slice would ever ship on its own, ahead of the slices that implement it, your concern applies in full and we should revisit.

### On symlinks

Your catch on #339 was a good one, and it's worth noting that this slice is on the right side of it. `private/fs.ts` keeps the `lstatSync()`-based walkers, and the behaviour is now documented rather than incidental:

> Symlinks are recreated as symlinks rather than dereferenced, which keeps the copy consistent with `hashPath` (it hashes links by their target) and makes dangling links and link cycles harmless.

The new `shouldExclude` predicate threads through both the legacy and canonical schemes without touching symlink handling, so the #321 baseline is preserved here.

One gap in the same area: the new fixtures under `test/fs/fixtures/test1/` (a local symlink and a dangling absolute one) aren't referenced by anything — `test/fs/` contains no test file at all. That looks like exactly where the symlink coverage should land, and having the fixtures pre-staged makes it cheap.

### On test quality

You already made this point on #339 — that the `SOURCE` vs `OUTPUT` tests asserted only `toBeDefined()` or hash length rather than the advertised contract — so treat this as an ack rather than a new finding. It carries directly into this slice: `assets-types.test.ts` is 280 lines asserting enum string values and that struct literals compile, which would pass against any implementation. Meanwhile `excludeMatcher()` and the `shouldExclude` threading are the genuinely new logic in this PR and currently have no tests or callers.

That's a very fixable shape of problem, and the groundwork is already there. The matcher's contract is well documented — I checked its stated limitations against 22 edge cases locally (bare `*`, `**`, `!` negation, leading `/`, `./` prefixes, `dir` vs `dirty/file`, case sensitivity) and it behaves exactly as the docstring says. A behavioural suite over those cases plus the staged fixtures would lock it in, and it's a better use of the slice than the type-shape assertions currently there.

eduardomourar:
Agreed. No release boundary between these slices, so keeping extraHash / bundling / assetHashType in place makes sense.

To back that up with real implementation, I'll fold in AssetStaging Core with no bundling: the AssetStaging class doing SOURCE/OUTPUT/CUSTOM hash resolution through the shared fingerprinting path (gated on canonicalAssetHashes, per the unification in `d212d56`), just without Docker bundling. That gives the types actual behavior instead of shipping ahead of it.

To keep that from bloating the PR, I'll move the Docker-specific bits — the docker* fields on DockerImageAssetSource and DockerCacheOption — into the Docker Bundling PR, since they've got no caller until bundling lands anyway.

So this PR ends up as: foundation types (minus Docker interfaces) + a working AssetStaging core (SOURCE/OUTPUT/CUSTOM, no bundling). Docker bundling PR gets the Docker interfaces plus the actual DockerImage/bind-mount/volume-copy execution.

@so0k and @jsteinich, could you both confirm if that is the approach we want to take? I know that is not the small API surface we had in mind, but, at least, we can release those slices separately if needed be.

eduardomourar:
@jsteinich and @so0k, this is ready for review now fully based on the discussed from #380.

packages/cdktn/src/assets.ts
```diff
@@ -0,0 +1,197 @@
...
+   */
+  readonly producesDirectory: boolean;
+
+  /**
+   * Perform the staging transformation, writing the packaged result to
+   * `target`.
+   * @param source - path to the resolved (already bundled, if applicable) source
+   * @param target - path the packaged result should be written to
+   */
+  pack(source: string, target: string): void;
```
jsteinich: `pack()` has no way to receive an exclusion, and both implementations silently drop one:

```ts
class DirectoryPackaging { pack(source, target) { copySync(source, target); } }    // shouldExclude never passed
class ZipPackaging       { pack(source, target) { archiveSync(source, target); } } // no exclusion support at all
```

`copySync` gained a `shouldExclude` option in this PR and it is never passed; `archiveSync(src, dest)` has no exclusion parameter to pass.

This is latent today because `TerraformAsset` has no `exclude`. But `AssetHash.of()` in this same PR does exclude, and `Asset` / `AssetStaging` will. Once they meet, the hash omits files the staged copy still contains, which is exactly the hash/artifact drift this design set out to prevent.

Worth settling before this ships, because changing a public JSII method signature afterwards is breaking. A struct parameter is the future-proof shape here, since adding struct fields is additive while adding method parameters is not:

```ts
export interface PackOptions {
  readonly source: string;
  readonly target: string;
  readonly ignoreStrategy?: IIgnoreStrategy;
}

export interface IAssetPackaging {
  readonly extension: string;
  readonly producesDirectory: boolean;
  pack(options: PackOptions): void;
}
```

`archiveSync` needs an exclusion parameter as well, but that one is internal and cheap.

jsteinich (commented):
The trimmed `assets.ts` and the behavioural `asset-hash.test.ts` are big improvements. Wiring `TerraformAsset` through `PACKAGING_BY_TYPE` also gives `IAssetPackaging` a real consumer rather than shipping it ahead of one, which is better than what the delta asked for, and removing the unreferenced `test/fs/fixtures/` closes an earlier thread.

Four notes below. The `pack()` signature is the one I'd want settled before merge, since it is public API. The canonical pinning is a decision worth making explicitly. The last two are cheap now and expensive later.

packages/cdktn/src/asset-hash.ts
```diff
@@ -0,0 +1,74 @@
...
+   * @param filePath - path to a file or directory to hash
+   * @param options - see {@link AssetHashOptions}
+   */
+  public static of(filePath: string, options: AssetHashOptions = {}): string {
+    const resolved = path.resolve(filePath);
+    const strategy =
+      options.ignoreStrategy ??
+      new ExcludeIgnoreStrategy(options.exclude ?? []);
+
+    const baseHash = hashPath(resolved, {
```
jsteinich, line 65: Neither `canonical` nor `archive` is passed here, so `hashPath` falls back to the legacy walker. `TerraformAsset` passes `canonical: !!this.node.tryGetContext(CANONICAL_ASSET_HASHES)`, so with the flag enabled the same directory hashes differently through the two paths.

**This should pin to canonical:**

```ts
const baseHash = hashPath(resolved, {
  canonical: true,
  shouldExclude: (relativePath) => strategy.ignores(relativePath),
});
```

The legacy scheme exists to preserve hashes already in the wild, and `canonicalAssetHashes` is a future flag on its way to becoming the default. A brand-new public API has no existing hashes to preserve, so it should start on the scheme that is staying rather than the one being retired.

Deliberately *not* adding a `canonical?: boolean` option: there is no reason to expand the public API surface just to expose a mode that is going away.

Leaving `archive` unset is correct here. There is no emitted archive to match framing against, so hashing the tree as a tree is the right choice.

packages/cdktn/src/asset-hash.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Options for {@link AssetHash.of}.
+ */
+export interface AssetHashOptions {
+  /**
+   * Paths to exclude, relative to the hashed path. Ignored if
+   * `ignoreStrategy` is given.
+   *
+   * @default - nothing is excluded
+   */
+  readonly exclude?: string[];
```
jsteinich, line 20: `exclude` is silently ignored when `ignoreStrategy` is supplied. The doc comment says so, but silently discarding a caller's input tends to be discovered the hard way rather than by reading the docs. Throwing when both are provided would be kinder, and it is a cheap guard.

packages/cdktn/src/ignore-strategy.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Decides whether a path relative to an asset root is excluded from staging
+ * and hashing.
+ *
+ * Core ships only the exact-path / `*.ext` / directory matcher used by
+ * `exclude` today (`ExcludeIgnoreStrategy`). Full glob, `.gitignore`, and
+ * `.dockerignore` parity can be implemented against this interface without
+ * core taking on a glob parser.
+ */
+export interface IIgnoreStrategy {
```
jsteinich, line 14: Worth considering an optional identity member on this interface now:

```ts
readonly cacheKey?: string;
```

`AssetStaging`'s result cache keys on `JSON.stringify` of its inputs, and a custom strategy object cannot be serialised into that key, so supplying one skips the cache entirely. That is correct behaviour, but it means reaching for a `.dockerignore` strategy silently costs all caching.

The cache itself lives in #339, but the interface lives here, and adding a member to a released JSII interface is considerably more expensive than adding one now.

packages/cdktn/src/assets.ts
eduardomourar: Fixed — `pack()` now takes a `PackOptions` struct (source, target, ignoreStrategy?) instead of positional args, exactly as you sketched. `archiveSync` picked up an optional exclude predicate to match. Threaded through `DirectoryPackaging/ZipPackaging`, added a test that both actually drop excluded paths.

packages/cdktn/src/asset-hash.ts
eduardomourar: Yep, pinned it to `canonical: true` unconditionally. Didn't add a public `canonical` option since there's nothing to preserve here yet. Added a test that catches drift (retargeted symlink case).

packages/cdktn/src/asset-hash.ts
eduardomourar: Agreed, now throws if both `exclude` and `ignoreStrategy` are passed instead of quietly ignoring one.

packages/cdktn/src/ignore-strategy.ts
eduardomourar: Added `cacheKey?: string` to `IIgnoreStrategy`, `ExcludeIgnoreStrategy` sets it from the exclude list so the default path stays cacheable. Custom strategies without one just keep missing the cache, as before.

Side note: making it a required prop on `ExcludeIgnoreStrategy` fails jsii build (JSII5009, can't narrow an optional interface member to required in the impl) — worth keeping in mind for future interface members like this.

packages/cdktn/test/asset-hash.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    expect(() =>
+      AssetHash.of(tempDir, {
+        exclude: ["*.log"],
+        ignoreStrategy: new ExcludeIgnoreStrategy(["*.tmp"]),
+      }),
+    ).toThrow(/exclude.*ignoreStrategy|ignoreStrategy.*exclude/i);
+  });
+
+  test("hashes on the canonical scheme regardless of the feature flag", () => {
```
jsteinich, line 116: Two problems with this test, both fixed by the same rewrite.

**1. It doesn't discriminate between the two schemes.**

The comment says the legacy scheme "would not reliably tell them apart", but `legacyHashPath` hashes symlink targets explicitly:

```ts
links.update(`${relPath}\0${fs.readlinkSync(p)}\0`);
```

So retargeting the symlink changes the legacy digest too. This test passes with or without `canonical: true`, which means it doesn't protect the pin it's named for — remove the `canonical: true` line and it stays green.

What actually separates the two schemes: legacy hashes only file *bytes* (`content.update(fs.readFileSync(p))`) with no path, and never records directory entries. Canonical frames every entry with its path. So identical bytes under different names collide on legacy and diverge on canonical:

```ts
// dirA/a.txt = "hello"   vs   dirB/b.txt = "hello"
expect(AssetHash.of(dirA)).not.toBe(AssetHash.of(dirB));
```

That fails if the pin is ever removed, which is the property worth locking in.

**2. `fs.symlinkSync` is ungated.**

`test/helper/capabilities.ts` on this branch exports `testIfSymlinks` for exactly this — on Windows without Developer Mode `symlinkSync` throws `EPERM` and the test fails rather than skipping. Relevant to #364.

The rewrite above needs no symlinks at all, so it resolves this at the same time. If symlink coverage is still wanted separately, `testIfSymlinks` is the way to add it.

jsteinich (commented):
Thanks — all four items from the previous pass look good. `PackOptions` threads the strategy through to both `copySync` and `archiveSync`, and putting the `shouldExclude` check ahead of the `lstatSync` in the archive walk means excluding a directory correctly skips its whole subtree. The canonical pin, the conflicting-options throw, and `cacheKey` all read the way I'd hoped.

Two follow-ups on the new test, below. Both are fixed by the same rewrite.

packages/cdktn/test/asset-hash.test.ts
eduardomourar: Done.

jsteinich (approved):
Looks good to me.
What do you think @so0k ?

packages/cdktn/src/ignore-strategy.ts
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Whether the given path should be excluded.
+   * @param relativePath - `/`-separated path relative to the asset root
+   */
+  ignores(relativePath: string): boolean;
```
so0k: I don't think this signature can reach the `.gitignore` / `.dockerignore` parity that #380 gives as the reason for the interface:

- **Directories have no marker.** The walkers call `ignores("foo")` for a directory, never `"foo/"`. Git-style `foo/` patterns match directories only. As far as I recall, node-ignore expects directory paths to end in `/`, so an implementation can't tell the two cases apart without `stat`-ing, and it doesn't know the root to `stat` against.
- **Excluded subtrees are pruned.** `copySync`, `archiveSync` and both hash walkers stop descending once a directory is excluded. That makes Docker's `node_modules` + `!node_modules/keep` inexpressible, and so is `*` + `!src/`.

This is inherited from AWS CDK. TerraConstructs' port has the same limitation: `ignores(absoluteFilePath)` with no directory flag, and pruning at `fs/copy.ts:34` / `fs/fingerprint.ts:76`. It matters more here because of [the container-context case](https://github.com/opentofu/opentofu/issues/3309#issuecomment-5637974082). BuildKit applies `.dockerignore` itself, so a strategy that disagrees with it produces a trigger hash that misses real context changes.

Changing an interface method's signature after release breaks Java, C# and Python implementers, so it's worth settling before merge. For example, pass a struct argument for extensibility:

```ts
export interface IgnoreQuery {
  readonly relativePath: string;
  readonly isDirectory: boolean;
}

export interface IIgnoreStrategy {
  ignores(query: IgnoreQuery): boolean;
  readonly cacheKey?: string;
}
```

It's also worth deciding whether a strategy can opt out of subtree pruning, e.g. when it has negation patterns.

packages/cdktn/src/asset-hash.ts
```diff
@@ -0,0 +1,84 @@
...
+    const resolved = path.resolve(filePath);
+    const strategy =
+      options.ignoreStrategy ??
+      new ExcludeIgnoreStrategy(options.exclude ?? []);
+
+    // Pinned to the canonical scheme: this is a brand-new API with no
+    // existing hashes to preserve, so it has no reason to start on the
+    // legacy scheme that `canonicalAssetHashes` exists to move away from.
+    const baseHash = hashPath(resolved, {
+      canonical: true,
```
so0k, line 89: Pinning canonical makes sense. The remaining problem is that the framing of the hash depends on packaging, and that dependency isn't visible anywhere:

- `AssetHash.of()` never sets `archive`, so directory records are always included.
- `TerraformAsset` passes `archive: true` for `AssetType.ARCHIVE`, which omits directory records so the hash tracks the zip bytes (#323).

So for any directory with subdirectories, `AssetHash.of(dir) !== new TerraformAsset(..., { path: dir, type: ARCHIVE }).assetHash`, even with `canonicalAssetHashes` enabled. For `FILE` / `DIRECTORY` the two only agree when the flag is on.

Leaving `archive` unset here is deliberate ([review comment](https://github.com/open-constructs/cdk-terrain/pull/371#discussion_r3940542837)), and it matches #380's split: the source hash names the artifact, and `Fn.filebase64sha256(asset.path)` checks the artifact itself. The consequence isn't written down anywhere, though: `AssetHash.of(dir)` won't equal `TerraformAsset(dir, ARCHIVE).assetHash`, because #323 makes that one track the zip bytes. It also means a custom `IAssetPackaging` (e.g. `tar.bz2`) doesn't change the identity. That's fine, but it should be stated.

Could the `AssetHash.of` doc say it is a source-tree identity that doesn't depend on packaging, and could a test pin its relationship to `TerraformAsset` for `DIRECTORY` and `ARCHIVE` so it can't change unnoticed?

packages/cdktn/src/asset-hash.ts
```diff
@@ -0,0 +1,84 @@
...
+ * staged copy; this is the identity half without the staging half.
+ */
+export class AssetHash {
+  /**
+   * Content hash of a file or directory, without staging it.
+   * @param filePath - path to a file or directory to hash
+   * @param options - see {@link AssetHashOptions}
+   */
+  public static of(filePath: string, options: AssetHashOptions = {}): string {
+    if (options.exclude && options.ignoreStrategy) {
```
so0k: Nit: `[]` is truthy, so a caller that spreads defaults, e.g. `{ exclude: [], ...props }` together with an `ignoreStrategy`, will throw here even though nothing is being excluded.

```ts
if (options.exclude?.length && options.ignoreStrategy) {
```

packages/cdktn/src/asset-hash.ts
```diff
@@ -0,0 +1,84 @@
...
+      throw assetHashConflictingExcludeOptions();
+    }
+
+    const resolved = path.resolve(filePath);
```
so0k: `path.resolve(filePath)` resolves relative paths against `process.cwd()`. `TerraformAsset` resolves relative `path` against the directory containing `cdktf.json` (`cdktfJsonPath` context / `findFileAboveCwd`). Both work when the CLI runs the app from the project root, but they diverge when a test runner or script runs from elsewhere. Should they be aligned, or should this doc comment state which base it uses?

packages/cdktn/src/assets.ts
```diff
@@ -0,0 +1,235 @@
...
+}
+
+/**
+ * The type of asset hash
+ *
+ * NOTE: the hash is used in order to identify a specific revision of the asset, and
+ * used for optimizing and caching deployment activities related to this asset such as
+ * packaging, uploading to cloud storage, etc.
+ */
+export enum AssetHashType {
```
so0k, line 59: [DRAFT] These types have no consumer in this PR:

- `AssetHashType`: `OUTPUT` is documented as "based on the content of the bundling output", but there's no bundler interface yet. #380 also argues `OUTPUT` should be documented as forcing an eager, non-skippable build.
- `AssetOptions.assetHash`: documents "may only contain letters, digits, `_`, `.` and `-`", but nothing validates it.
- `IAsset`: not implemented anywhere. `TerraformAsset` has `assetHash` but doesn't declare `implements IAsset`.
- `StagedAsset`: nothing produces one.

Following up on [the 08-12 thread](https://github.com/open-constructs/cdk-terrain/pull/371#issuecomment-5270597204) and [the reply](https://github.com/open-constructs/cdk-terrain/pull/371#issuecomment-5273976171): that agreement assumed these slices land together and that the `AssetStaging` core would be folded in here. At this head the fold-in isn't present, the middle slice from the [split plan](https://github.com/open-constructs/cdk-terrain/pull/339#issuecomment-5208734511) hasn't been opened, and #339 still targets `main` without being rebased onto this branch. So nothing in git keeps this slice from reaching the next release-please cut on its own. Either merge it together with the `AssetStaging` slice (or into a feature branch), or trim to the parts that have callers.

packages/cdktn/src/assets.ts
```diff
@@ -0,0 +1,235 @@
...
+   */
+  readonly extension: string;
+
+  /**
+   * Whether the staged result is a directory rather than a single file.
+   *
+   * Publishers branch on this to decide whether they upload one object or
+   * sync a tree.
+   */
+  readonly producesDirectory: boolean;
```
so0k, line 103: [DRAFT] `producesDirectory` and `extension` aren't read by anything yet. `TerraformAsset` still decides the directory-vs-file layout from `this.type` and names archives `archive.zip` (`ARCHIVE_NAME`) rather than using `extension`. So these members of a public interface haven't been validated by any caller. Could `TerraformAsset._onSynthesize` / `path` derive from them, or can they wait for the slice that reads them?

packages/cdktn/src/private/fs.ts
```diff
@@ -259,6 +314,34 @@ function canonicalHashPath(src: string, includeDirectories: boolean): string {
   return hash.digest("hex");
 }
 
+/**
+ * Build a predicate matching the exclusion forms documented on
+ * `TerraformAssetConfig.exclude`: an exact relative path, a `*.ext` suffix, or a
```
so0k: `TerraformAssetConfig.exclude` doesn't exist in this PR HEAD. There's no `exclude` in `terraform-asset.ts` on this branch or on `main`, so this points at an option users can't find.

 @jsteinich's delta proposes exclude on AssetHashOptions and AssetProps, not on TerraformAssetConfig.

Relatedly, no caller in core passes `PackOptions.ignoreStrategy`: `TerraformAsset` calls `packaging.pack({ source, target })`. The `shouldExclude` threading in `legacyHashPath` also has no caller, since `AssetHash.of` pins canonical. Either add `exclude` to `TerraformAssetConfig` here, which would give the pack/hash exclusion path a real end-to-end consumer, or reword this to reference `AssetHashOptions.exclude`.

packages/cdktn/test/assets-types.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        });
+
+        const zipped = fs.readFileSync(target);
+        expect(zipped.toString("latin1")).toContain("a.txt");
+        expect(zipped.toString("latin1")).not.toContain("b.log");
+      });
+    });
+  });
+
+  describe("StagedAsset", () => {
```
so0k, line 138: [DRAFT] The `StagedAsset` and `AssetOptions` blocks only check that struct literals compile and read back, so they would pass against any implementation.

What's missing is a test for the property this PR claims: packing and hashing agree on the file set. For example:

```ts
const strategy = new ExcludeIgnoreStrategy(["*.log", "node_modules"]);
AssetPackaging.DIRECTORY.pack({ source, target, ignoreStrategy: strategy });
expect(AssetHash.of(target)).toBe(AssetHash.of(source, { ignoreStrategy: strategy }));
```

Also, `excludeMatcher` has only three assertions, via `ExcludeIgnoreStrategy`. The edge cases checked by hand earlier in this thread (bare `*`, `**`, `!`, leading `/`, `./`, `dir` vs `dirty/file`, case sensitivity) would lock in its documented limits.

Minor: the ZIP test searches the compressed bytes with `toString("latin1")`. Listing entries with fflate's `unzipSync` would be more robust.

so0k (approved):
nitpicks, didn't get through it all (left draft comments where I haven't personally reviewed it again) - don't want to block this

packages/cdktn/src/ignore-strategy.ts
so0k: see also https://github.com/TerraConstructs/base/pull/165

eduardomourar: Fixed both. `ignores` now takes a struct with `isDirectory`, so `foo/` directory-only patterns work.

For pruning: added `pruneExcludedDirectories` (default `true`). Set it `false` and the walk still drops the excluded dir but keeps descending, so `node_modules` + `!node_modules/keep` and `*` + `!src/` work. Added a test that `AssetHash.of(stagedCopy) === AssetHash.of(source, { ignoreStrategy })` under a negation strategy, so pack and hash stay in sync.

packages/cdktn/src/asset-hash.ts
eduardomourar: Good point. Added both.

Doc on `AssetHash.of` now says it's a source-tree identity that doesn't depend on packaging.

Pinned it with tests in `canonical-asset-hash.test.ts` for a dir with subdirs:
- equals `DIRECTORY` asset hash with the flag on
- not equal to `ARCHIVE` (archive omits directory records)
- not equal to `DIRECTORY` on the legacy scheme

So the relationship can't drift unnoticed.

packages/cdktn/src/asset-hash.ts
eduardomourar: Done

packages/cdktn/src/asset-hash.ts
eduardomourar: When adding CDKTN to projen, I ran into similar issue. Let's be consistent and follow the same approach used by `TerraformAsset`.

packages/cdktn/src/private/fs.ts
eduardomourar: Good catch. The doc referenced `TerraformAssetConfig.exclude`, which doesn't exist. Probably something I forgot to modify when following the discussion from #380.

So I went with the reword: the `excludeMatcher` doc now points at `AssetHashOptions.exclude` instead. When a real need for asset-level exclusion on `TerraformAsset` shows up, we can add the option then with an actual consumer behind it.

packages/cdktn/src/assets.ts
eduardomourar: Wired the two that can have a real consumer now:

- `TerraformAsset implements IAsset`. `assetHash` became `readonly` for jsii (set once in the constructor anyway); worth a changelog note as a pre-release narrowing.
- Added `assetHashType` to `TerraformAssetConfig`: `SOURCE`/unset hashes the source, `CUSTOM` uses `assetHash` (throws if missing), `OUTPUT` throws "not supported yet". So `OUTPUT` can't ship as a silent no-op.

`StagedAsset` still has no producer. That belongs with the `AssetStaging` slice, so I'd rather not fake a consumer here. I do believe it is best to avoid cutting a new release until we have those consumers in place.

packages/cdktn/src/assets.ts
eduardomourar: Done. `path`/`_onSynthesize` now use `packaging.producesDirectory`, and `fileName` derives from `packaging.extension` (`archive` + ext), replacing the hardcoded `archive.zip`. Output is identical for all three types (verified against the old constants; existing tests pass), so no behavior change. Added tests pinning the layout to the packaging.

packages/cdktn/src/terraform-asset.ts
```diff
@@ -99,13 +125,63 @@ export class TerraformAsset extends Construct {
...
+      if (
+        assetHashType !== undefined &&
+        assetHashType !== AssetHashType.CUSTOM
+      ) {
+        throw assetHashConflictingHashType(id);
+      }
+      return assetHash;
+    }
+
+    switch (assetHashType) {
```
jsteinich, line 152: This switch has no `default`, so an enum value outside the three members falls through and returns `undefined` — `assetHash` then ends up undefined rather than raising.

TypeScript is satisfied because `CUSTOM` / `OUTPUT` / `SOURCE` / `undefined` are exhaustive over the declared type, but that guarantee stops at the JSII boundary: a caller from Python, Go, Java, or C# passing an out-of-range value reaches this at runtime. The risk is low since jsii validates enums at the kernel boundary, but a `default` that throws closes it for free and keeps the failure loud rather than silent.

packages/cdktn/src/ignore-strategy.ts
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * The default ignore strategy: exact paths, `*.ext` suffixes, and
+ * directories (with everything inside them).
+ */
+export class ExcludeIgnoreStrategy implements IIgnoreStrategy {
+  private readonly matcher: (relativePath: string) => boolean;
+
+  public readonly cacheKey?: string;
+
+  // The built-in patterns have no negation form, so nothing below an excluded
```
jsteinich: Worth trimming the inline rationale in this round back toward the house style — implementation reasoning belongs in the commit message or PR body rather than at the declaration site.

The public JSDoc is a different case and should stay: `IgnoreQuery`, `pruneExcludedDirectories`, and the `AssetHash.of` block all ship as API documentation into five languages, and they earn their length there.

It's the inline ones I'd shorten — this JSII5009 note, the descend-into-excluded explanations in `copySync` / `legacyHashPath` / `canonicalHashPath`, and the repackaging note on `TerraformAsset.fileName`. Each explains *why the change was made*, which is what the commit message is for; a reader of the code a year from now mostly needs *what it does*.

jsteinich (commented):
Went through the latest round. The `IgnoreQuery` struct and the `pruneExcludedDirectories` opt-out both look right — carrying `isDirectory` from the walker's `lstat` is the only way a `.gitignore`-style strategy can honour directory-only patterns, and negation genuinely does require descending into an excluded parent.

Also confirmed the `TerraformAsset` refactor is behaviour-preserving for all three existing types: ARCHIVE resolves to `"archive" + ".zip"` exactly as the old `ARCHIVE_NAME` did, FILE and DIRECTORY keep `basename`, and `producesDirectory` maps onto the old `type === DIRECTORY` checks in both the `path` getter and `_onSynthesize`. `canonical-asset-hash.test.ts` is purely additive — nothing existing was rewritten.

Two notes below.

packages/cdktn/src/terraform-asset.ts
```diff
@@ -34,15 +51,29 @@ export enum AssetType {
...
+ */
+const PACKAGING_BY_TYPE: Record<AssetType, IAssetPackaging> = {
+  [AssetType.FILE]: AssetPackaging.FILE,
+  [AssetType.DIRECTORY]: AssetPackaging.DIRECTORY,
+  [AssetType.ARCHIVE]: AssetPackaging.ZIP,
+};
+
 // eslint-disable-next-line jsdoc/require-jsdoc
-export class TerraformAsset extends Construct {
+export class TerraformAsset extends Construct implements IAsset {
```
jsteinich, line 73: `assetHash` changes from mutable to `readonly` here, and that is a change against `main`, not just against an earlier revision of this PR:

```
main       ->  public assetHash: string;
this PR    ->  public readonly assetHash: string;
```

For an already-published class that removes the generated setter in Python, Java, C# and Go. The commit carrying it is `feat(lib): consume asset types`, with no `!` and no `BREAKING CHANGE:` footer, so release-please will not surface it — it would ship silently.

It is not a gratuitous tightening, though. `implements IAsset` forces it; jsii rejects the mutable form outright:

```
error JSII5010: "TerraformAsset#assetHash" turns mutable when implementing IAsset.
Make it readonly [language-compatibility/override-changes-mutability]
```

So the two are a package: keeping `implements IAsset` means keeping `readonly`, and undoing `readonly` means dropping `implements IAsset` as well.

Worth noting what the `implements` clause buys today: nothing consumes `IAsset` yet — no parameter, no return type, nothing but the declaration and this clause. TypeScript's structural typing means `TerraformAsset` already satisfies `IAsset` without it. The declaration only starts to matter in Java, C# and Go, where jsii generates nominal interfaces — and only once something actually accepts an `IAsset`.

Preference is to undo both for now. That keeps this PR free of an API break, loses nothing in practice, and lets `implements IAsset` land together with the first thing that takes an `IAsset` — a publisher, most likely — where the break is paid for by something concrete.

If there is a reason to keep it in this PR that I am missing, that is fine too, but then the commit needs a `!` or a `BREAKING CHANGE:` footer so it reaches the changelog.

jsteinich (commented):
One more, on a change against `main` rather than against an earlier state of this PR.

packages/cdktn/src/terraform-asset.ts
jsteinich: Claude got a bit ahead of me here. I suspect that we'll want that interface implemented for the future work and this PR is about establishing the interface. I'm ok with it, but we do need to note that it's a breaking change.

packages/cdktn/src/terraform-asset.ts
eduardomourar: Added a `default` to the `resolveAssetHash` switch that throws `assetHashTypeUnknown`, so an out-of-range value from another jsii language fails loudly instead of returning undefined. Added a test for it.

packages/cdktn/src/ignore-strategy.ts
eduardomourar: Trimmed the inline rationale comments (JSII5009 note, the descend-into-excluded explanations, the `fileName` repackaging note) down to what-it-does. Left the public JSDoc on `IgnoreQuery`, `pruneExcludedDirectories`, and `AssetHash.of` as-is since those ship as API docs.

packages/cdktn/src/terraform-asset.ts
eduardomourar: You're right, and I confirmed it: `TerraformAsset` ships on `main` with a mutable `public assetHash`, so jsii generates a setter in Java/C#/Python/Go. Making it `readonly` removes that setter — a real break for those bindings, even though it was only ever assigned once at construction and setting it would just desync from `path`.

Keeping `implements IAsset` and `readonly` as you suggested (the interface is what we're establishing here for the future publisher work). I'll mark the commit as breaking so release-please surfaces it.

Issue #361, opened 2026-07-29
cli: Crash on failed deploy


jsteinich:
### Expected Behavior

No crash on failed deploy with just the error displayed

### Actual Behavior

Crash on deploy failure after displaying the error.

```
187  docker_image.nginxImage: Creating...
187  ╷
│ Error: failed to create Docker client: Error pinging Docker server, please make sure that npipe:////./pipe/docker_engine is reachable and has a  '_ping' endpoint. Error: error during connect: in the default daemon configuration on Windows, the docker client must be run with elevated privileges to connect: Head "http://%2F%2F.%2Fpipe%2Fdocker_engine/_ping": open //./pipe/docker_engine: The system cannot find the file specified.
│
│   with docker_image.nginxImage (nginxImage),
│   on cdk.tf.json line 46, in resource.docker_image.nginxImage (nginxImage):
│   46:       }
│
╵
undefined
undefined
Collecting Debug Information...
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
Invoking Terraform CLI failed with exit code 1
(node:33900) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 11)
(Use `node --trace-warnings ...` to show where the warning was created)
Debug Information:
language: typescript
cdktn-cli: 0.24.0-pre.91
node: v22.22.2
cdktn: 0.23.1
cdktf: null
constructs: 10.6.0
jsii: null
terraform: 1.14.3
arch: x64
os: win32 10.0.26200
```

### Steps to Reproduce

1. Setup a project from https://github.com/open-constructs/cdk-terrain/issues/187
2. Run cdktn deploy while docker is not running
3. Observe the issue

### Versions

language: typescript
cdktn-cli: 0.24.0-pre.91
node: v22.22.2
cdktn: 0.24.0-pre.91
cdktf: null
constructs: 10.6.0
jsii: null
terraform: 1.14.3
arch: x64
os: win32 10.0.26200
providers
aws@~> 6.45 (LOCAL)
        terraform provider version: 6.56.0

### Migration Context

Not sure / don't remember

### References

- #187 
- #360 

### Help Wanted

- [ ] I'm interested in contributing a fix myself

### Community Note

[...]

PR #264, opened 2026-06-10
chore(cli): replace Ink + React with smaller-tree CLI libraries

X-Guardian:
### Related issue

Fixes #263

### Description

Implements the Ink + React replacement proposed in the linked issue: swaps the Ink/React UI layer for `@inquirer/prompts` + `cli-spinners` + `cli-table3` + `ansi-escapes` + `cli-cursor`. Same commands, flags, approval prompts, and `--auto-approve` semantics.

### Improvements

- **Detached stdin exits cleanly** instead of crashing with Ink's "Raw mode is not supported" stack trace.
- **Non-interactive output is ANSI-free** — bar/spinner suppressed whenever stdout isn't a TTY.

**Measured result:**

| Metric                                              | Before  | After   |
| --------------------------------------------------- | ------- | ------- |
| Shipped bundle (`bundle/bin/cmds/handlers.js`)      | 3.7 MB  | 2.9 MB  |
| Transitive install size of the affected dep trees   | 22.6 MB | ~6 MB   |

**What changes:**

- **New helpers** in `packages/cdktn-cli/src/bin/cmds/helper/`:
  - `tty-stream.ts` — `StreamRenderer` class. Log-above / bar-below pattern with a dots spinner (`cli-spinners`), wrap-aware row counting (via `strip-ansi` + `out.columns` so the erase/repaint doesn't cannibalise scrollback when the bar wraps), paused-state log buffering during interactive prompts, and a leading blank row above the bar that only paints once logs are streaming (so spinner-only commands stay single-spaced). Painting is gated on `stdout.isTTY && !isCI` (`is-ci`), so the bar suppresses itself for piped/redirected output and for CI runners whether or not they allocate a TTY.
  - `project-runner.ts` — `runCdktfProject`, replaces the `useCdktfProject` React hook by subscribing to `CdktfProject`'s existing event callbacks. Maps cli-core's `ProjectUpdate` events to a discriminated `Status` union the UI consumes.
  - `prompts.ts` — `promptApprove`, `promptOverride`, and the `requireTty` / `isNonTtyError` helpers that let callers fall back to `status.stop()` cleanly when no TTY is present.
  - `format.ts` — pure render functions for outputs, stack list (Stack-name column sized to 40% of terminal width for stable column positions across runs), provider table, execution counter, watch state.
- **All UI commands rewritten as `.ts` (no JSX):** `get`, `provider-list`, `synth`, `list`, `output`, `deploy`, `destroy`, `diff`, `watch`.
[...]

PR #375, opened 2026-08-07
fix(cli): don't crash after a successful deploy when an output is missing

so0k:
### Related issue

No GitHub issue — this was reported by a user through **Sentry (CDKTN-5)** on `cdktn-cli 0.24.0`, not filed as an issue. Happy to open one retroactively if you'd prefer the paper trail.

Related: #361 — the top-level `.fail()` handler is the common exit path that turned this throw into an unhandled rejection with a stack trace. Deliberately **not** fixed here; it deserves its own PR and a regression test.

### Description

A user's `cdktn deploy` applied both stacks successfully and then crashed:

```
TypeError: Cannot convert undefined or null to object
    at Object.entries (<anonymous>)
    at renderNested (helper/format.ts:126)   <- twice, nested
    at renderOutputs (helper/format.ts:142)
    at runDeploy (ui/deploy.ts:164)
```

#### Root cause

`getConstructIdsForOutputs` maps synth metadata (`cdk.tf.json` → `"//".outputs`) onto the real
`terraform output -json` result with an unchecked lookup:

```ts
return { ...acc, [key]: outputs[value] };   // undefined when terraform didn't return it
```

An output declared in metadata but **absent** from terraform's result was retained with value `undefined`.
`isObjectEmpty` only drops a group where *every* output is missing, so a **partial** miss survived. `renderNested`
then reached that node — `isTerraformOutput(undefined)` is `false`, control fell through, and
`Object.entries(undefined)` threw.

The two nested `renderNested` frames put the bad value at `outputsByConstructId[stack][constructId]`, depth 1,
which matches. The reporter's `network` stack mixed user outputs with generated `cross-stack-output-*` entries
feeding a dependent `db` stack — a realistic way to get a partial miss (state drift, `--skip-synth`, an output
renamed or removed between applies).

#### Why the debug log looked healthy

The logged `OutputsByConstructId` appeared complete, which initially looked like it contradicted the theory. It
doesn't: those lines log via `JSON.stringify`, which silently omits `undefined`-valued keys. The poison map and a
clean map serialize **byte-identically** — verified — while `Object.keys` reveals the extra key. The defect was
structurally invisible in the logs.

#### Why it surfaced in 0.24.0

The defective data path is unchanged since before v0.23 (`git diff v0.23.4..v0.24.0` on `output.ts` is a single
unrelated zod line). What changed is #264, which removed `renderInk` — whose `try/catch` used to swallow this
throw as a one-line message and `exit(1)`. Without it the throw escapes unhandled, so a long-latent bug now
produces a stack trace, telemetry and a Sentry report. **An exposure regression, not a new defect.**

#### The fix — producer first, with defence in depth

- **`getConstructIdsForOutputs`** omits keys whose terraform output is absent, and logs which one was dropped.
  The check is an explicit `=== undefined`, *not* a truthiness test, so an output whose legitimate value is
  `""`, `0` or `false` still renders.
- **`renderNested` and `unpackTerraformOutput`** skip non-object nodes instead of recursing into
  `Object.entries`. These are deliberately defensive: with the producer fixed there is no *known* input that
  reaches them, but both walk externally-sourced data and neither should be able to take down the CLI after a
  successful deploy. `unpackTerraformOutput` is the one that runs first on the `--outputs-file` path.
- **`isTerraformOutput`** excludes `null`, which previously threw on `null.sensitive` rather than returning `false`.
- **Rendering the outputs table is now non-fatal** in `runDeploy`. The deploy has already succeeded by that
  point, so a presentation failure is logged instead of failing the deploy — restoring the safety net that
  deleting `renderInk` removed.

#### Behaviour note

`renderNested` now renders nothing for a group whose children all drop, and for a legitimately empty group,
where it previously printed a bare header.

#### Verification

The new tests were checked against the **unfixed** code, not just the fixed code: reverting the source changes
while keeping the tests reproduces the exact reported `TypeError`, with the same nested `renderNested` frames,
end-to-end through the real `runDeploy`. Assertions use `not.toHaveProperty` rather than `toEqual`, since
`toEqual` treats a missing key and an `undefined` value as equal and would pass on the buggy code.

---

#### Stacked PR

#376 builds on this branch and fixes a **separate** crash on the `--outputs-file` path found during review of
this change. It is deliberately split out — the reporter never passed `--outputs-file`, so none of it is on this
crash's path.

### Checklist

- [ ] I have made corresponding changes to the [documentation](https://github.com/open-constructs/cdk-terrain-docs/tree/main/content) if applicable — n/a, no documented behaviour changes

### DIFF
--- a/packages/@cdktn/cli-core/src/lib/output.ts
+++ b/packages/@cdktn/cli-core/src/lib/output.ts
@@ -24,17 +24,26 @@ function unpackTerraformOutput(
   outputs: NestedTerraformOutputs,
   includeSensitiveOutputs: boolean,
 ): Record<string, string> {
-  return Object.entries(outputs).reduce(
-    (acc, [key, entry]) => ({
+  return Object.entries(outputs).reduce((acc, [key, entry]) => {
+    if (isTerraformOutput(entry)) {
+      return {
+        ...acc,
+        [key]:
+          !entry.sensitive || includeSensitiveOutputs ? entry.value : undefined,
+      };
+    }
+    // Defensive hardening: `entry` comes from externally-sourced data (terraform output metadata
+    // mapped by getConstructIdsForOutputs), and Object.entries() below would throw on a null or
+    // undefined entry. Omit the key rather than recurse into it or retain an undefined/null value -
+    // consistent with how getConstructIdsForOutputs omits unresolved outputs.
+    if (entry === null || typeof entry !== "object") {
+      return acc;
+    }
+    return {
       ...acc,
-      [key]: isTerraformOutput(entry)
-        ? !entry.sensitive || includeSensitiveOutputs
-          ? entry.value
-          : undefined
-        : unpackTerraformOutput(entry, includeSensitiveOutputs),
-    }),
-    {},
-  );
+      [key]: unpackTerraformOutput(entry, includeSensitiveOutputs),
+    };
+  }, {});
 }
 
 export async function saveOutputs(
@@ -227,7 +236,17 @@ export const getConstructIdsForOutputs = (
   const mapOutputs = (value: OutputIdMap): NestedTerraformOutputs => {
     return Object.entries(value).reduce((acc, [key, value]) => {
       if (typeof value === "string") {
-        return { ...acc, [key]: outputs[value] };
+        const resolved = outputs[value];
+        if (resolved === undefined) {
+          // The metadata declares an output that terraform did not return (state drift,
+          // --skip-synth, a renamed/removed output, ...). Omit the key entirely rather than
+          // retaining it with an undefined value, which would later crash the renderer.
+          logger.debug(
+            `Output "${value}" (construct id "${key}") declared in stack metadata but absent from terraform output; omitting it.`,
+          );
+          return acc;
+        }
+        return { ...acc, [key]: resolved };
       }
 
       const mapped = mapOutputs(value);
--- a/packages/@cdktn/cli-core/src/lib/output.ts
+++ b/packages/@cdktn/cli-core/src/lib/output.ts
@@ -211,6 +211,12 @@ export const parseOutput = (str: string): DeployingResource[] => {
   }, new Array());
 };
 
+// Matches the prefix TerraformStack.registerOutgoingCrossStackReference() uses when it
+// synthesizes an output for a cross-stack dependency (packages/cdktn/src/terraform-stack.ts).
+// There is no shared constant to import for it - the identifier is not exported anywhere - so the
+// prefix is duplicated here.
+const CROSS_STACK_OUTPUT_PREFIX = "cross-stack-output-";
+
 const isObjectEmpty = (obj: Record<string, any>): boolean => {
   if (typeof obj !== "object") {
     return false;
@@ -241,9 +247,18 @@ export const getConstructIdsForOutputs = (
           // The metadata declares an output that terraform did not return (state drift,
           // --skip-synth, a renamed/removed output, ...). Omit the key entirely rather than
           // retaining it with an undefined value, which would later crash the renderer.
-          logger.debug(
-            `Output "${value}" (construct id "${key}") declared in stack metadata but absent from terraform output; omitting it.`,
-          );
+          //
+          // Cross-stack outputs are generated plumbing for stack dependencies rather than
+          // something the user declared directly, and a dependent stack legitimately produces
+          // many of them - warning on each missing one would be noise the user cannot act on.
+          // A directly user-declared output silently going missing is exactly the kind of
+          // condition that made the original crash hard to diagnose, so it gets a warning.
+          const message = `Output "${value}" (construct id "${key}") declared in stack metadata but absent from terraform output; omitting it.`;
+          if (value.startsWith(CROSS_STACK_OUTPUT_PREFIX)) {
+            logger.debug(message);
+          } else {
+            logger.warn(message);
+          }
           return acc;
         }
         return { ...acc, [key]: resolved };
--- a/packages/cdktn-cli/src/bin/cmds/ui/deploy.ts
+++ b/packages/cdktn-cli/src/bin/cmds/ui/deploy.ts
@@ -162,23 +162,35 @@ export async function runDeploy({
 
     onOutputsRetrieved(outputs);
 
-    if (outputs && Object.keys(outputs).length > 0) {
-      try {
-        console.log(renderOutputs(outputs));
-      } catch (e) {
-        // The deploy already succeeded at this point; a failure rendering the outputs table is
-        // purely cosmetic and must not fail the deploy. Log it so the user still learns something
-        // went wrong, but do not rethrow.
-        console.error(
-          `\nDeploy succeeded, but rendering the outputs failed: ${
-            e instanceof Error ? e.message : e
-          }`,
-        );
-      }
+    let rendered = "";
+    let renderFailed = false;
+    try {
+      rendered = outputs ? renderOutputs(outputs) : "";
+    } catch (e) {
+      // The deploy already succeeded at this point; a failure rendering the outputs table is
+      // purely cosmetic and must not fail the deploy. Log it so the user still learns something
+      // went wrong, but do not rethrow.
+      console.error(
+        `\nDeploy succeeded, but rendering the outputs failed: ${
+          e instanceof Error ? e.message : e
+        }`,
+      );
+      renderFailed = true;
+    }
+
+    if (rendered) {
+      console.log(rendered);
+    }
+
+    if (rendered || renderFailed) {
       if (outputsPath) {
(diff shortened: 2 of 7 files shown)


sakul-learning:
## Review: PR #375 — `fix(cli): don't crash after a successful deploy when an output is missing`

**Verdict: APPROVE** ✅

### Summary

This PR fixes a crash where the CLI would throw `TypeError` after a successful `terraform apply` when a metadata-declared output was absent from `terraform output -json` (e.g., state drift, renamed output, `--skip-synth`). The fix adds defense-in-depth at every layer of the output rendering pipeline.

### Changes (7 files, +362/−18)

| Layer | File | Change |
|---|---|---|
| Type guard | `packages/@cdktn/cli-core/src/lib/models/terraform.ts` | +null guard in `isTerraformOutput` (`typeof null === "object"` in JS) |
| Output mapping | `packages/@cdktn/cli-core/src/lib/output.ts` | `getConstructIdsForOutputs` omits absent outputs instead of retaining `undefined`; `unpackTerraformOutput` guards null/non-object before `Object.entries()` |
| CLI rendering | `packages/cdktn-cli/src/bin/cmds/helper/format.ts` | `renderNested`/`renderOutputs` guard null/non-object values + filter empty-node results |
| Deploy command | `packages/cdktn-cli/src/bin/cmds/ui/deploy.ts` | Wraps `renderOutputs()` in try/catch — render failure is now non-fatal |
| Tests | `output.test.ts`, `format.test.ts`, `deploy.test.ts` | 10 new test cases covering the regression path and edge cases |

### Configured Checks

| Check | Result |
|---|---|
| `pnpm exec nx build cdktn` | ✅ Pass |
| `pnpm exec nx test cdktn --runInBand` | ✅ 542 passed, 43 suites, 301 snapshots |

### Correctness

- **`isTerraformOutput` null guard:** Necessary and sufficient — `typeof null === "object"` is the only falsy value that passes the `typeof` gate.
- **`getConstructIdsForOutputs` missing-output handling:** Correctly omits absent keys. The `isObjectEmpty` interaction correctly drops groups where all children were omitted.
- **`unpackTerraformOutput` recursion:** Null/typeof guard precedes `Object.entries()`, preventing crashes at any nesting depth.
- **`renderNested` empty-group behavior:** Legitimately empty `{}` groups now print nothing instead of a bare header — safe because `getConstructIdsForOutputs` already strips empty groups upstream.
- **`runDeploy` try/catch:** `stream.stop()` in `finally` is guaranteed on all paths (success, render-failure, project-failure). Tests validate: no rejection, callback fires, error logged, `stop()` called.

### One adjacent note (not a blocker)

`packages/cdktn-cli/src/bin/cmds/ui/output.ts` (the `cdktn output` command) calls `renderOutputs(returnValue)` without the same try/catch this PR adds to `runDeploy`. The same missing-output scenario could crash the `output` command. Pre-existing and out of scope here — worth a follow-up hardening pass.

### Artifact-value

Lean, proportional fix. Every line defends against a real crash. No new abstractions, no speculative code.

Ship it. 🚢

so0k:
Thanks for the review.

On the adjacent note about `cdktn output` — good catch, and it is already handled in the stacked #376, which
applies the same treatment to `runOutput` (it had the identical fire-and-forget `onOutputsRetrieved` defect as
well as the missing render guard). It is not visible from this PR's diff since #376 sits on top of this branch.

Kept out of this PR deliberately: the reporter of the original crash never passed `--outputs-file`, so none of
that work is on this crash's path, and #376's change to the exit behaviour deserves its own review.

jsteinich:
Traced the analysis against `main` — the root cause and the producer-first fix both hold up. One question and two small notes.

**Should the drop be visible?**

https://github.com/open-constructs/cdk-terrain/blob/6e8226dad2a20d7e86c6fd25bfc63591f7131cca/packages/%40cdktn/cli-core/src/lib/output.ts#L244-L246

`logger` is log4js at default level, so this is invisible without `CDKTF_LOG_LEVEL=debug`. Net effect for the reported scenario: instead of crashing, we print a table quietly missing an output the user declared. Silent is consistent with the existing `isObjectEmpty` full-miss behaviour, but this is exactly the condition that was invisible in the Sentry report.

Would you consider `logger.warn` for outputs that aren't `cross-stack-output-*`? Your call — you've looked at the real data path more closely than I have, and I'll approve either way.

Two smaller things:

1. The non-fatal render guard lands in `runDeploy` but not `runOutput`, which has the identical unguarded `console.log(renderOutputs(...))`. #376 covers it — only matters if this ever lands alone.
2. When every group drops, `renderOutputs` returns `""` but `runDeploy`'s `Object.keys(outputs).length > 0` gate still passes, so we `console.log("")` and emit a stray blank line.

so0k:
Both good catches, and I'll take the suggestion on the first.

**Visibility of the drop — yes, `logger.warn`, split the way you describe.**

You've put your finger on the uncomfortable part: the failure mode this PR produces is "quietly missing an
output the user declared", and that is the same class of silence that made the original crash so hard to
diagnose from the Sentry report. Trading a crash for a silent omission is a poor trade for a *user-declared*
output.

The `cross-stack-output-*` split is the right seam — those are generated plumbing for stack dependencies, so a
warning per missing one would be noise the user cannot act on. A user-declared output vanishing is actionable:
it means their state and their synthesized metadata disagree. So: `logger.warn` for user-declared,
`logger.debug` for `cross-stack-output-*`.

**The stray blank line — real bug, mine.**

`runDeploy` gates on `Object.keys(outputs).length > 0`, which counts keys whose *values* were all dropped, so
`renderOutputs` returns `""` and we `console.log("")`. Introduced by the empty-group change in this PR. Fixing
by gating on the rendered string rather than the key count, with a test.

**`runOutput`** — agreed, and covered in #376. This PR is not intended to land alone, but you're right that it
would be a gap if it did.

Both changes incoming; I'll push and re-request.

so0k:
Pushed `b36961754` with both changes.

- **`logger.warn` for user-declared outputs**, `logger.debug` kept for `cross-stack-output-*`. No shared
  constant for that prefix exists anywhere (it's an inline template literal in
  `terraform-stack.ts:535`, used once), so this defines a local `CROSS_STACK_OUTPUT_PREFIX` with a comment
  pointing at the origin rather than importing something that isn't there. Tests cover both branches.
- **Stray blank line gone.** The gate is now the rendered string rather than the key count.

One judgement call worth flagging, since it's user-visible: when the keys survive but every output under them
drops, this prints `No outputs found.` rather than nothing — from the user's point of view there is nothing to
show, so it should read the same as the genuinely-empty case.

Note that #376 then changes this case slightly: because the `written to` line is keyed off write success rather
than render state, `--outputs-file` with all-dropped outputs prints `No outputs found.` *and* the `written to`
line. Both are defensible; calling it out so the two PRs read consistently.

packages/@cdktn/cli-core/src/lib/output.ts
```diff
@@ -227,7 +242,26 @@ export const getConstructIdsForOutputs = (
...
-        return { ...acc, [key]: outputs[value] };
+        const resolved = outputs[value];
+        if (resolved === undefined) {
+          // The metadata declares an output that terraform did not return (state drift,
+          // --skip-synth, a renamed/removed output, ...). Omit the key entirely rather than
+          // retaining it with an undefined value, which would later crash the renderer.
+          //
+          // Cross-stack outputs are generated plumbing for stack dependencies rather than
+          // something the user declared directly, and a dependent stack legitimately produces
+          // many of them - warning on each missing one would be noise the user cannot act on.
```
jsteinich, line 253: I don't really think something being a cross stack output is all that relevant to whether or not a message is logged.
A missing reference could cause incorrect incorrect plans (though that might just result in a Terraform error), or it could be completely benign. The context of what the user is doing matters.
The same seems true for user defined outputs.

packages/cdktn-cli/src/bin/cmds/ui/deploy.ts
```diff
@@ -158,14 +159,38 @@ export async function runDeploy({
     );
 
     const outputs = project.outputsByConstructId;
+
     onOutputsRetrieved(outputs);
 
-    if (outputs && Object.keys(outputs).length > 0) {
```
jsteinich, line 163: Seems like omitting this check is going to generate a warning that could easily be avoided.

packages/cdktn-cli/src/bin/cmds/ui/deploy.ts
```diff
@@ -158,14 +159,38 @@ export async function runDeploy({
...
+      renderFailed = true;
+    }
+
+    if (rendered) {
+      console.log(rendered);
+    }
+
+    if (rendered || renderFailed) {
       if (outputsPath) {
         console.log(`The outputs have been written to ${outputsPath}`);
```
jsteinich, line 187: If `renderFailed`, is this actually true?

Issue #78, opened 2026-03-28
Use TypeScript 6


jsteinich:
### Description

TypeScript 6 has recently been released, but the project doesn't currently use or fully support it.
Some aspects to fix / look into:

- Support in init template
  - This was [recently disabled](https://github.com/open-constructs/cdk-terrain/pull/74) due to an upstream dependency conflict.
- Use in the library itself
  - Need to solve the same problems as the init template due to the same dependency being used.
- May need changes in JSII for support
- Handle [breaking changes](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#breaking-changes-and-deprecations-in-typescript-6.0)

### Help Wanted

- [x] I'm interested in contributing a fix myself

### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #395, opened 2026-09-05
feat(deps): adopt jsii 6.0 and TypeScript 6

### LINKED ISSUE #373: Adopt jsii 6.0 and constructs >=10.8.1 to remove the <10.8.0 peer dep cap
## Context

`cdktn@0.24.0` caps its `peerDependencies.constructs` at `<10.8.0` because `constructs@10.8.0` upgraded from jsii 5.9.x → 6.0.x (also TypeScript ~5.9 → ~6.0). The `<10.8.0` cap is a deliberate guardrail until the cdktn toolchain migrates to jsii 6.0.

`constructs@10.8.1` exists (not deprecated, on jsii 6.0.x), fixing the bug that caused npm to deprecate 10.8.0.

### What needs to happen

1. cdktn itself and `@cdktn/provider-project` upgrade devDependencies:
   - `jsii` from `~5.9.0` → `~6.0.0`
   - `jsii-pacmak` from `1.x` → `1.139.0+`
   - `typescript` from `~5.9` → `~6.0`
2. `@cdktn/provider-generator` updates:
   - `jsii` from `~5.9.0` → `~6.0.0`
   - `jsii-pacmak` from `1.128.0` → `1.139.0+`
   - `constructs` hard-pin from `10.6.0` → `>=10.8.1`
3. cdktn relaxes its `peerDependencies.constructs` from `>=10.6.0 <10.8.0` to `>=10.6.0` (or `^10.8.1`)
4. `cdktn-repository-manager` upgrade workflow propagates the new range to all pre-built providers

### References

- `constructs@10.8.0` removed `jsii.tsc` config, replaced by `jsii.tsconfig` + `jsii.validateTsconfig` (jsii 6.0 feature)
- `constructs@10.8.0` is deprecated on npm: "This version has a bug that breaks downstream jsii consumers"
- `constructs@10.8.1` is not deprecated (same jsii 6.0.x)
- `.jsii` assembly schema is identical (`jsii/0.10.0`) — no format migration needed
- Published JS output is identical CJS — no ESM migration occurred

### Related

- PR that added the cap: https://github.com/open-constructs/cdk-terrain/pull/363
- Repo-manager upgrade PR: https://github.com/cdktn-io/cdktn-repository-manager/pull/63

jsteinich:
### Related issue

Fixes https://github.com/open-constructs/cdk-terrain/issues/373
Part of https://github.com/open-constructs/cdk-terrain/issues/78 (library half; the init-template half is #396)

### Description

Moves the jsii toolchain to 6.0. The `cdktn` library is compiled with TypeScript 6; every other package stays on TypeScript 5.x, since jsii bundles its own compiler and nothing forces them to move.

| package | before | after |
| --- | --- | --- |
| jsii | `~5.9.0` | `~6.0.0` |
| jsii-rosetta | `~5.9.0` | `~6.0.0` |
| jsii-pacmak / codemaker | `1.128.0` | `1.140.0` |
| jsii-docgen | `^10.4.20` | `^10.12.6` |
| constructs (peer) | `>=10.6.0 <10.8.0` | `^10.6.0` |
| TypeScript (jsii packages) | `^5.0.4` | `~6.0.0` |

#### Dropping the `constructs` cap instead of widening it

The `<10.8.0` cap from #363 is removed rather than replaced with a disjunction, because only 10.8.0 itself was ever broken:

| version | `pkg.jsii.tsc` | assembly `metadata.tscOutDir` |
| --- | --- | --- |
| 10.6.0 | `{outDir, rootDir}` | absent |
| **10.8.0** | **absent** | **absent** |
| 10.8.1 | `{outDir, rootDir}` | `"lib"` |

10.8.0 satisfies neither lookup path, which is why `cdktn convert` emitted `constructs.Construct` instead of `software.constructs.Construct`. 10.8.1 restores both, is not deprecated on npm (10.8.0 is, with "This version has a bug that breaks downstream jsii consumers"), and is the higher version — so no range resolver can select 10.8.0 without an explicit pin. Keeping the range a simple caret avoids propagating a disjunction into pre-built providers' peer ranges.

The `cdktn` devDependency stays at the range floor (`10.6.0`) so the assembly is compiled against the oldest `constructs` it claims to support. That is also what jsii's `metadata/missing-dev-dependency` check wants, so the build is warning-free.

Separately, cdktn's own assembly now emits `metadata.tscOutDir` under jsii 6, so it is itself immune to this class of bug for its downstream consumers regardless of package.json shape.

#### Class fields now use define semantics

jsii 6 emits ES2023, so `useDefineForClassFields` is on. Impact was checked against a build of the same commit with the old toolchain:

- The emitted `.d.ts` files are **byte-identical**; the public type surface does not move, and TypeScript 5.4.5 / 5.9.3 consumers still type-check against it cleanly with `skipLibCheck: false`.
- `TS2612` (field overwriting a base property) only exists under define semantics — a check jsii 6 newly turns on. It passes across the library, the generated provider bindings, and the cross-package chain into `constructs`.
- The compiler's blind spot — dynamically installed base properties — is empty here: all 12 such sites in the library are `Object.defineProperty(this, SOME_SYMBOL, ...)`, and symbol keys cannot collide with the string-named fields define semantics emits.
- Generated provider attributes are emitted as `private _x?` backing fields plus `get`/`set` accessors, so the surface that scales to thousands of classes is unaffected by construction.

#### Not included

The init template stays on TypeScript 5.x here. Moving it to 6 requires replacing ts-node first, which is #396.

### Checklist

`jsii` reports `Errors: 0 | Warnings: 0`. Integration suites have not been run locally — they need Terraform plus the Java/Python/.NET toolchains — so CI is the first full exercise of the pacmak path for those languages.


so0k (approved):
Approving — tested and validated against the job this PR needs to fix, plus one commit pushed on top.

### What was tested

#398 (Terraform test ceiling → 1.16.1) rebased onto this branch, run locally (macOS, Node 22.22.3, Python 3.11, pipenv 2026.8.0) through the same three phases CI's `Examples Integration Tests / @examples/python-documentation @ tf1.16.1` runs — `reinstall` → `cdktn get` → `cdktn synth` — with `TERRAFORM_BINARY_NAME=terraform1.16.1` and the CLI installed from `dist/js` via `test/run-against-dist.mjs`.

Result: all three phases pass (reinstall 4s, get 77s, synth 85s). The generated kubernetes bindings now import cleanly:

```
imports/kubernetes/provider/__init__.py:45
    _provider_functions_c77894a1 = _LazyImport("kubernetes.provider_functions")

$ pipenv run python -c "import imports.kubernetes.provider as p; print(p.KubernetesProvider)"
<class 'imports.kubernetes.provider.KubernetesProvider'>
```

That is the exact spot where #398's CI dies today on pacmak 1.128.0 (`from .functions import …` → `ModuleNotFoundError: imports.kubernetes.provider.functions`). Mechanism, verified in the installed `jsii-pacmak@1.140.0`: `relativeImportPath` (`lib/targets/python/type-name.js`) is still exported but has no caller anywhere in `lib/` — Python cross-submodule imports are emitted as `LazyImport` since 1.136.

So: once this merges, #398 should go green on a plain rebase, and #400's `provider-functions` → `functions` rename is no longer required for it.

### Pushed on top: 8636bbb7e

Relocks the 11 Python example `Pipfile.lock`s (jsii 1.125.0 → 1.140.0, constructs 10.4.5 → 10.8.1, transitives to current). The wheel built from this branch declares `Requires-Dist: jsii <2.0.0,>=1.140.0`, so the committed pins were stale. This is hygiene for local development, not a CI fix — `tools/build-example.mjs` runs each example's `reinstall` first and regenerates the lock on every run. Pipfiles untouched, `_meta.hash` carried over, formatting as committed.

### Local-repro notes (not defects in this PR)

- Ran `cdktn get`/`synth` without `--show-performance-info`: its `pidtree` `ps` polling hit `spawn EBADF` on macOS under `cdktn get`'s concurrent provider builds. Linux CI is unaffected (#398's CI gets past `get`).
- Needed `pnpm store prune`: pnpm caches the force-republished `cdktn@0.0.0` from the local Verdaccio indefinitely, which first produced a stale-library compile error (`markWriteOnlyAttribute` missing) unrelated to this branch.

### Sequencing

#396 branches from this branch and targets `main` (its diff is a superset of this one), so this PR merges first.

PR #383, opened 2026-08-19
fix(provider-generator): satisfy module provider configuration aliases on get

jsteinich:
Fixes #379

`cdktn get` fetches a module schema by writing a synthetic root config that calls the module and running `terraform get` in it. A module declaring `configuration_aliases` in its `required_providers` makes that root config invalid — Terraform demands the caller pass those configurations in — and `terraform get` used to downgrade the resulting error to a warning and write the module manifest anyway, which is what the fetch relied on.

Terraform 1.15 routed init and get through the graph-based install workflow ([hashicorp/terraform#38217](https://github.com/hashicorp/terraform/pull/38217)), which never reaches that downgrade (`installErrsOnly` no longer has any effect for `get`) and returns before writing the manifest. From 1.15.7 on, `cdktn get` therefore fails with "Missing required provider configuration" for any such module, and `.terraform/modules/modules.json` is never written, so the failure cannot simply be tolerated. OpenTofu still has the downgrade in both of its install paths and is unaffected.

## Approach

Read what the module actually declares rather than guessing up front or parsing the error text:

1. The first `terraform get` is treated as a fetch whose diagnostics are deferred. Terraform installs the module either way — it just skips the manifest.
2. The module's own `required_providers` block (parsed with the `hcl2json` call the harvest step already makes) says which aliased configurations are required.
3. Those are declared in the root config and passed to the module call, and a second `terraform get` produces the manifest.

Modules without `configuration_aliases` are unaffected, and a fetch that fails for any other reason still reports its original error.

Alternatives considered and rejected: parsing alias names out of the error prose (English-prose-dependent); reading them from the Terraform Registry API before fetching (`provider_dependencies` does not expose `configuration_aliases`, and it would not cover git or local sources); `terraform init -from-module` (skips validation entirely but takes no version argument, so registry modules cannot be pinned); version-gating on Terraform >= 1.15 (no benefit — the declaration-driven flow is version-agnostic and also silences the pre-1.15 warning).

`exec` gains a `logStderrAsDebug` option so the deferred fetch does not print an `[ERROR]` block that a following success contradicts. When the fetch turns out not to be salvageable, the held-back diagnostics are re-logged at error level before rethrowing, so that output is unchanged.

## Testing

- New `local-module-provider-aliases` fixture and snapshot test, a test that a genuinely unfetchable module still fails loudly, and unit tests for the new helpers.
[...]

Issue #337, opened 2026-07-18
CI: thin-slice runtime coverage for newer-protocol features on current stable Terraform AND OpenTofu


so0k:
## Why

CI's Terraform ceiling is **1.6.5** (`.terraform.versions.json` `tested: ["1.5.5","1.6.5"]`, which every workflow matrix derives from). Terraform only emits `functions` / `ephemeral_resource_schemas` / `write_only` from 1.8/1.10/1.11 — so no CI job can ever fetch a schema containing the features #296 adds. The generation codepaths run in CI against empty inputs; the generated surfaces are covered only by fixture unit tests, the edge-provider compile slice (#311), and the external demo harness ([sakul-learning/cdktn-provider-features-demo#5](https://github.com/sakul-learning/cdktn-provider-features-demo/pull/5)). That demo run caught three real bugs snapshot tests couldn't (nodenext import resolution, Node type-stripping rejecting a parameter property, variadic union binding) — proof the runtime slice has unique value.

## Scope: one integration test × two pinned jobs

A single new integration test (`test/typescript/provider-features/`) running the standard harness (`cdktn get` → tsc → synth) against **hashicorp-namespace providers only (`random` + `time`)** — both small and mirrored on `registry.opentofu.org`. Assertions:

- generated provider class exposes the `functions` getter; instance-method bindings compile and synthesize `provider::<name>::<fn>(...)` expressions;
- nullable-parameter rendering: omitted optional and explicit `null` produce the bare `null` keyword **in position** (no JS `undefined` dropping / arg shifting);
- ephemeral resource synthesizes to the top-level `ephemeral` key;
- `targetVersions` floors below the feature admit range fail synth with the family validation error.

**Matrix: exactly two additive entries, landing together** — Terraform current stable (1.15.x) and OpenTofu current stable (1.12.x) — pinned per-test, explicitly *not* by touching `.terraform.versions.json` (the existing 1.5.5/1.6.5 cross-product stays untouched).

## Implementation notes

- Preferred mechanism: an opt-in per-test `pinVersions` field in `tools/build-test-matrix.mjs` so the thin job inherits existing integration plumbing (container, caching, `ci/skip-integration` label). #312's TF>=1.10 ephemeral test should reuse the same mechanism.
[...]

Issue #301, opened 2026-07-03
test(tofu): Track OpenTofu provider signing support in matcher tests


sakul-learning:
## Summary

Track OpenTofu support for provider installation/signature verification in the `cdktn` Terraform matcher tests.

While investigating PR #300, the `toBeValidTerraform` and `toPlanSuccessfully` matcher tests failed locally when `TERRAFORM_BINARY_NAME` was set to OpenTofu 1.12.3 instead of Terraform. This is not specific to PR #300; it appears to be an OpenTofu/provider-signing compatibility issue with the older `kreuzwerker/docker` provider version selected by the test fixture's current constraint.

## Observed behavior

Environment:

- Terraform: `Terraform v1.7.5`
- OpenTofu: `OpenTofu v1.12.3`
- Failing env override: `TERRAFORM_BINARY_NAME=/home/vincent/.local/share/mise/installs/opentofu/1.12.3/tofu`
- Test fixture provider constraint: `kreuzwerker/docker`, `version = "~> 2.0"`

OpenTofu resolves the current test constraint to `kreuzwerker/docker v2.25.0`, then fails during `tofu init`:

```text
- Installing kreuzwerker/docker v2.25.0...
Error: Failed to install provider
Error while installing kreuzwerker/docker v2.25.0: the provider is not signed with a valid signing key; please contact the provider author (authentication signature from unknown issuer)
```

Terraform accepts the same provider/version:

```text
- Installing kreuzwerker/docker v2.25.0...
- Installed kreuzwerker/docker v2.25.0 (self-signed, key ID BD080C4571C6104C)
Partner and community providers are signed by their developers.
```

## Reproduction / version check

Minimal `required_providers` snippets were tested under `/data/tmp/opentofu-docker-provider-signing-test`.

### Current broad test constraint: `~> 2.0`

```hcl
terraform {
  required_providers {
    docker = {
      source  = "kreuzwerker/docker"
      version = "~> 2.0"
    }
  }
}
```

Result:

- `tofu init -input=false -no-color`: fails, resolves/install-attempts `v2.25.0`, signing-key error
- `terraform init -input=false -no-color`: succeeds, installs `v2.25.0` as self-signed

### Exact legacy version: `2.25.0`

Result:

- `tofu init -input=false -no-color`: fails with the same signing-key error
- `terraform init -input=false -no-color`: succeeds

### Latest registry version at time of test: `4.5.0`

```hcl
terraform {
  required_providers {
    docker = {
[...]

Issue #208, opened 2026-05-25
dependency-manager: migrate to opentofu registry


so0k:
### Description

Hashicorp has deprecated CDKTF support and removed all sample code from the terraform provider registry.

OpenTofu has indicated interest in CDK Terrain samples on their registry, if we put in in place all the necessary automation to help maintain the sample on the opentofu registry.

Raised during review of short urls

https://github.com/open-constructs/cdk-terrain/blob/58e3a11f8d82f3b7673f7af4c9be652f4ef3cbdc/packages/%40cdktn/cli-core/src/lib/dependencies/dependency-manager.ts#L371

### Help Wanted

- [ ] I'm interested in contributing a fix myself

### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #398, opened 2026-09-08
chore(gha): adopt the new jsii-terraform image and test Terraform 1.5.7 + 1.16.1

jsteinich:
### Related issue

Follow-up to #394. Step 2 of the sequencing agreed in #337. Rebased on #383, which unblocks the ceiling half.

### Description

Three commits, all consequences of #394 having published a new image and #383 having fixed `cdktn get` on modern Terraform.

**1. Bump the pinned image digest** — all 27 references across 11 workflow files move to `sha256:62c00a31...`, the image built from 2f29d5a8d. Confirmed to be that build: `:latest` and the immutable `0.24.0-2f29d5a8d45b047c5450eee9640e1cbaaee98c8d` tag resolve to the same digest. First image carrying OpenTofu 1.12.6, Terraform 1.5.7 and 1.16.1, the fail-fast install layers and the post-install binary verification.

**2. Move the floor leg from Terraform 1.5.5 to 1.5.7** — 1.5.7 is the newest pre-fork release and the exact floor cdktn already declares. `DEFAULT_TARGET_VERSIONS` in `packages/cdktn/src/validations/target-versions.ts` is `{ terraform: ">=1.5.7", opentofu: ">=1.6.0" }`, so every project that declares no `targetVersions` is told 1.5.7 is supported, while CI was testing 1.5.5 — a version nothing declares.

**3. Raise the ceiling from 1.6.5 to 1.16.1** — unblocked by #383 (54a8217). `cdktn get` was broken from Terraform 1.15.7 onward for any module declaring `configuration_aliases` (#379); until that fix was in main this would have landed red.

Both version changes are **renames of existing legs, not additions** — `tested` goes from `["1.5.5", "1.6.5"]` to `["1.5.7", "1.16.1"]`, so the integration cross-product keeps the same job count and the CI bill is unchanged. The two legs now span the declared floor and current stable instead of two adjacent 2023 releases, which is what #337 needed in order to exercise the newer-protocol schema sections (`functions`, `ephemeral_resource_schemas`, `write_only`) at all — the previous 1.6.5 ceiling structurally could not emit them.

`default` moves to 1.16.1 with it. That drives the examples matrix (`tools/build-example-matrix.mjs` reads `.default`) and — less obviously — every provider-integration run: `provider-integration.yml`'s `terraform_version` input is defaulted rather than passed, and none of its three callers override it. The duplicated unit-test matrices in `pr-unit.yml` and `release.yml` move with it too.

### Deliberately not included

- **Pruning 1.5.5 and 1.6.5 from `available`.** They are unreferenced now, but editing `.terraform.versions.json` only rebuilds the image on merge — the workflows stay pinned to the current digest, so the prune would not take effect until a later bump. It belongs with the next digest bump.
- **`tools/generate-function-bindings/function-availability/functions-matrix.json`** still lists 1.6.5. That is generated function-availability data covering every release, not a CI version.

### Follow-ups

- Derive the `terraform_version` matrices in `pr-unit.yml` and `release.yml` from `.terraform.versions.json` instead of duplicating `tested` in three places, so a version change is a one-file edit.
- OpenTofu coverage (#337 step 3) still needs the pin mechanism plus #208, #301, #392, #393.

### Checklist

- [ ] I have made corresponding changes to the documentation if applicable — n/a
- [ ] I have added tests that prove my fix is effective or that my feature works if applicable — n/a; this PR's own CI run is the verification, since it exercises the new image on both new version legs
- [ ] New and existing unit tests pass locally with my changes — n/a, no source changes


so0k:
blocked on https://github.com/open-constructs/cdk-terrain/pull/400

so0k (approved):

PR #406, opened 2026-09-09
chore(deps): move the remaining packages to TypeScript 6

jsteinich:
### Related issue

Completes https://github.com/open-constructs/cdk-terrain/issues/78 — #395 moved the `cdktn` library to TypeScript 6; this moves the remaining nine packages.

### Description

No source changes were needed: the codebase was already TypeScript 6 clean. This is config and dependency hygiene.

**`baseUrl` removed** from the five tsconfigs that set it, because TypeScript 6 errors on it. Only `generate-function-bindings` pairs it with `paths`, which resolves relative to the tsconfig without it.

**Explicit `types` arrays** where the compile needs them:

| package | types |
| --- | --- |
| cdktn-cli, cli-core, commons, hcl2cdk, provider-generator, provider-schema | `["node", "jest"]` |
| generate-function-bindings | `["node"]` |
| hcl-tools, hcl2json | none needed — left alone |

**`@types/node` added to `commons` and `provider-schema`.** Neither declared it; both were relying on ambient resolution. Once `types` is explicit that ends, and TypeScript reports `TS2688: Cannot find type definition file for 'node'`. This is a latent gap the migration surfaces rather than one it creates.

#### One thing worth a second look

`cdktn-cli`'s `compile-build-config` runs `tsc build.ts --outDir build-config` — a file on the command line alongside a `tsconfig.json`. TypeScript 6 reports that as:

```
TS5112: tsconfig.json is present but will not be loaded if files are specified on commandline
```

That was already true under TypeScript 5; 6 just says so. Adding `--ignoreConfig` acknowledges it, but on its own it broke the build in a subtler way: TypeScript 6's config-less defaults emit ESM, so `build-config/build.js` was reparsed as a module and `fs.copySync` came back `undefined`. `--module commonjs --target es2022` pin what the tsconfig used to supply.

#### Not included

Root `typescript` stays at `5.4.5`. It backs `tsc-files` in the pre-commit hooks and `typescript-eslint`, neither of which needs to move for the packages to build on 6. Happy to bump it separately if you'd prefer one version across the repo.

### Checklist

Locally: `pnpm build` green for all 9 projects, `pnpm lint:workspace` green for 35, knip clean, and the unit suite fully green (cdktn 43/43 suites, 525 passed).

### DIFF
--- a/packages/@cdktn/cli-core/package.json
+++ b/packages/@cdktn/cli-core/package.json
@@ -72,7 +72,7 @@
     "nock": "^14.0.16",
     "prettier": "2.8.8",
     "tsc-files": "1.1.4",
-    "typescript": "^5.0.4"
+    "typescript": "~6.0.0"
   },
   "engines": {
     "node": ">=22.19.0"
--- a/packages/@cdktn/cli-core/tsconfig.json
+++ b/packages/@cdktn/cli-core/tsconfig.json
@@ -24,6 +24,10 @@
     "stripInternal": true,
     "skipLibCheck": true,
     "target": "ES2018",
+    "types": [
+      "node",
+      "jest"
+    ],
     "incremental": false,
     "esModuleInterop": true,
     "rootDir": "./src",
--- a/packages/@cdktn/commons/package.json
+++ b/packages/@cdktn/commons/package.json
@@ -51,13 +51,14 @@
     "validator": "13.15.35"
   },
   "devDependencies": {
+    "@types/node": "22.20.1",
     "@types/cross-spawn": "6.0.6",
     "@types/follow-redirects": "1.14.4",
     "@types/fs-extra": "11.0.4",
     "@types/semver": "7.7.1",
     "@types/validator": "13.15.10",
     "tsc-files": "1.1.4",
-    "typescript": "^5.0.4"
+    "typescript": "~6.0.0"
   },
   "lint-staged": {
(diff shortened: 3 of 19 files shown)


so0k (approved):

sakul-learning:
LGTM — verified the TypeScript 6 migration end to end.

- `pnpm exec jest packages/cdktn/test/validations.test.ts --runInBand` → 41 passed
- `pnpm exec nx build cdktn` → jsii successful
- `pnpm exec nx test cdktn --runInBand` → 542 passed / 43 suites
- Uncached builds of all 9 changed packages on TS 6.0.3 (cli-core, commons, hcl-tools, hcl2cdk, hcl2json, provider-generator, provider-schema, cdktn-cli, generate-function-bindings) → all green, including the `hcl-tools`/`hcl2json` Go builds and the `cdktn-cli` `compile-build-config` path with the new `--ignoreConfig` flag.

The `baseUrl` removal, explicit `types: ["node", "jest"]`, and `@types/node` additions are the standard TS 6 migration; confirmed `--ignoreConfig` is a valid `tsc` 6.0.3 flag.

Non-blocking note: the root `package.json` still pins `typescript: 5.4.5` (exact) while every leaf package moves to `~6.0.0`. Likely deliberate (the root `tsc` backs `typescript-eslint` at lint time), just flagging in case it should bump too.

PR #407, opened 2026-09-09
fix(lib): retry the staging install without --prefer-offline outside CI

jsteinich:
### Related issue

No issue — follow-up from #395, where this cost real time during the upgrade.

### Description

`packages/cdktn/scripts/pack.mjs` installs the staging dependencies with `--prefer-offline`. When the local npm cache holds a packument older than a dependency bump, npm never refreshes it and reports:

```
npm error code ETARGET
npm error notarget No matching version found for jsii-pacmak@1.140.0.
```

The version does exist — the cached index just predates it. The message blames the pin, so the natural reaction is to distrust the version rather than the cache. This happened repeatedly while upgrading jsii, on a long-lived container with a warm cache.

**Outside CI** the install is retried once with `--prefer-online`, which refreshes the metadata and succeeds.

**In CI the error is rethrown untouched.** A release build that cannot resolve a pinned version should fail rather than quietly reach for a different resolution — the fallback exists to unstick local development, not to paper over a non-reproducible build. `process.env.CI` is set by GitHub Actions.

### Verification

The happy path is unchanged (warm cache, single install, no retry).

The gate was exercised by pinning a dependency to a version that does not exist, which produces the real ETARGET:

```
CI unset   retry fired once, then failed
CI=true    no retry, failed immediately
```

### Checklist

### DIFF
--- a/packages/cdktn/scripts/pack.mjs
+++ b/packages/cdktn/scripts/pack.mjs
@@ -127,21 +127,36 @@ function copySourceFiles(sourceFiles) {
  *   --no-package-lock:    don't pollute staging with a fresh lockfile.
  *   --prefer-offline:     use the local npm cache before the network.
  *   --ignore-scripts:     don't run install scripts of bundled deps; bundling needs the tree on disk only.
+ *
+ * A cached packument older than a dependency bump makes --prefer-offline report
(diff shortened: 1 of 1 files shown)


so0k (approved):
thanks for this

PR #378, opened 2026-08-07
fix(cli): stop the top-level error handler racing yargs

so0k:
### Related issue

Fixes #361
Fixes #360 — the trailing unhandled-rejection dump. The *duplicated Terraform error block* also reported there
is a separate defect in `@cdktn/commons/src/util.ts` and is not touched here.

> **Third in a stack.** Targets `fix/outputs-file-write-failures`, so the diff here is only this change.
> Review order: #375 → #376 → this.

### Description

`cdktn.ts` registered an `async` `.fail()` handler that yargs does not await — the code's own comment said so —
and called `yargs.exit(1, error)` as its first statement. So the error printed once from the handler and again
from Node's unhandled-rejection path, followed by `Collecting Debug Information...` and a
`PromiseRejectionHandledWarning`.

#### Measured, not inferred

Both failure classes were reproduced against a build of this branch's base before any change:

| Trigger | Before |
|---|---|
| Unexpected internal error (corrupt `cdk.tf.json` + `--skip-synth`) | message + stack printed **twice**; never reached the `Debug Information:` label — the awaited collection never completed |
| Terraform failure (`TerraformOutput` referencing an undeclared resource) | `undefined` / `undefined`, then a hard **`ERR_UNHANDLED_REJECTION`** abort |

Both are deterministic rather than timing-dependent, which is what made them testable. The second is also a
much cheaper reproduction than the CTRL+C-during-`init` race described in #361 — no signal timing, no provider,
no credentials.

The `undefined` / `undefined` signature had a second cause: `terraform-cli.ts` threw a **raw string**, so
`error.message` and `error.stack` were both undefined. It now throws an `Error`. Deliberately a plain `Error`
rather than `Errors.External` — a non-zero terraform exit is exactly what debug collection exists for, and the
`Errors` factories fire an unawaited telemetry POST that a terraform failure has no reason to send.

#### The fix

The handler moves to `bin/error-handling.ts`:

- `.fail()` is now **synchronous** and only records the failure. It still calls `yargs.exit(1, error)` for its
  load-bearing `hasOutput` side effect.
- `runCli` awaits `parseAsync()`, reports once through a single awaited path, and makes the one `process.exit()`
  call. Its `try/catch` also catches synchronous handler throws, which yargs 17 never routes to `.fail()` at all.

#### Crash reporting needed care

This is the part worth reviewing closely. **Nothing in the old handler reported to Sentry**: the `yargs.exit()`
[...]

PR #410, opened 2026-09-09
fix(cli): stop the top-level error handler racing yargs

### LINKED ISSUE #360: cli: Extra noise displayed on errored plan
### Expected Behavior

Just the specific failure is shown.

### Actual Behavior

An additional error stack trace is shown:

```
⠋ Processing
[2026-07-28T20:13:03.275] [ERROR] default - ╷
│ Error: "assume_role_policy" contains an invalid JSON policy: not a JSON object
│
│   with aws_iam_role.role,
│   on cdk.tf.json line 26, in resource.aws_iam_role.role:
│   26:         "assume_role_policy": "INVALID",
│
169  ╷
│ Error: "assume_role_policy" contains an invalid JSON policy: not a JSON object
│
│   with aws_iam_role.role (role),
│   on cdk.tf.json line 26, in resource.aws_iam_role.role (role):
│   26:         "assume_role_policy": "INVALID",
│
╵
External Error: Stack failed to plan: 169. Please check the logs for more information.
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
Error: terraform plan -input=false -out plan exited with code 1
    at ChildProcess.<anonymous> (C:\ProgramData\nvm\v22.22.2\node_modules\cdktn-cli\bundle\bin\cmds\handlers.js:74:1489)
    at Object.onceWrapper (node:events:634:26)
    at ChildProcess.emit (node:events:519:28)
    at ChildProcess.emit (node:domain:489:12)
    at e.emit (C:\ProgramData\nvm\v22.22.2\node_modules\cdktn-cli\bundle\bin\cmds\handlers.js:64:14925)
    at maybeClose (node:internal/child_process:1101:16)
    at ChildProcess._handle.onexit (node:internal/child_process:304:5)
(node:33596) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 11)
(Use `node --trace-warnings ...` to show where the warning was created)
```

### Steps to Reproduce

1. Setup a project from https://github.com/open-constructs/cdk-terrain/issues/169
2. Run `cdktn plan`
3. Observe the error

### Versions

language: typescript
cdktn-cli: 0.24.0-pre.91
node: v22.22.2
cdktn: 0.24.0-pre.91
cdktf: null
constructs: 10.6.0
jsii: null
terraform: 1.14.3
arch: x64
os: win32 10.0.26200
providers
aws@~> 6.45 (LOCAL)
        terraform provider version: 6.56.0

### Migration Context

Not sure / don't remember

### Upstream cdktf Issue

_No response_

### Providers

_No response_

### Gist

_No response_

### Possible Solutions

_No response_

### Workarounds

_No response_

### Anything Else?

_No response_

### References

- #169

### Help Wanted

### Community Note

### LINKED ISSUE #361: cli: Crash on failed deploy

so0k:
Part 2 of 6 in a stack; review order S1 → S6; base is the previous slice.

2. `fix(cli): stop the top-level error handler racing yargs` (this PR)

### Related issue

Fixes #360, fixes #361

### Description

This is the former PR #378, landed inside this stack because it rewrites the same yargs `.fail()` lines the telemetry work needs, and because the failure metric added in S3 belongs on the seam this slice creates. Rebasing one file through two separate PRs was the alternative. The commit keeps its original author and the adaptation is a separate commit.

The defect: `.fail()` was registered `async` (yargs does not await it, as the code's own comment said) and called `yargs.exit(1, error)` as its first statement. The error printed once from the handler and again from Node's unhandled-rejection path.

Reproduced against a build of the base before any change, deterministically rather than by timing:

| Trigger | Before |
|---|---|
| Unexpected internal error (corrupt `cdk.tf.json` with `--skip-synth`) | message and stack printed twice, and the `Debug Information:` label never reached because the awaited collection never completed |
| Terraform failure (`TerraformOutput` referencing an undeclared resource) | `undefined` / `undefined`, then a hard `ERR_UNHANDLED_REJECTION` abort |

The `undefined` / `undefined` signature had a second cause: `packages/@cdktn/cli-core/src/lib/models/terraform-cli.ts` threw a raw string, so `error.message` and `error.stack` were both undefined. It throws an `Error` now.

The fix: the handler moves to `packages/cdktn-cli/src/bin/error-handling.ts`, `.fail()` becomes synchronous and only records the failure (it still calls `yargs.exit(1, error)` for its load-bearing `hasOutput` side effect), and `runCli` in `packages/cdktn-cli/src/bin/cdktn.ts` awaits `parseAsync()` with `.exitProcess(false)`, reports once through one awaited `reportFailure`, and makes the single `process.exit()` call. That `try/catch` also catches synchronous handler throws, which yargs 17 never routes to `.fail()` at all. `packages/cdktn-cli/src/bin/exit-flush.ts` extracts the `beforeExit` flush so the success path is testable too.

**Crash reporting needed care and is the part worth reviewing closely.** Nothing in the old handler reported to Sentry; what did report was the orphaned rejection from the un-awaited `.argv` promise, picked up by the global unhandled-rejection integration. Handling the rejection properly removes that channel, so `reportFailure` captures explicitly before flushing, for unexpected and `External` errors but not for `Usage`, which cli-core's `beforeSend` already drops. Exit codes are unchanged, measured per failure class before and after.

Scoping note for the stack: `reportFailure` here takes deps `{log, logError, collectDebugInformation, captureException, flushTelemetry}` and the flush dep calls `Sentry.flush` directly, because the bounded `flushTelemetry()` helper in commons does not exist yet. S3 reroutes it onto that helper and adds the `cli.command.error` metric on top. There is no usage telemetry in this slice.

Two behaviour changes worth knowing, both deliberate:

- **More crash events than before.** The old handler called `Sentry.close(4000)` but never captured the error, so unexpected errors reaching the entrypoint were not reported; what did reach Sentry was the orphaned rejection, picked up by the global unhandled-rejection integration. Handling the rejection removes that channel, so `reportFailure` now calls `captureException` for both `External` and unexpected errors (never for `Usage`). It stays consent-gated: Sentry is only initialised after consent, so the capture is inert otherwise.
- **The success path flushes and exits explicitly at the end of `runCli`.** An earlier revision hooked `process.on("beforeExit")`, which cannot fire while a pending Sentry request keeps the event loop alive, so the 4000 ms bound was never armed in exactly the case it exists for. `runCli` now awaits the bounded flush after `parseAsync()` resolves without a failure and then calls `process.exit(process.exitCode ?? 0)`; `--help` and `--version` take the same path. A child-process test points the fixture at a sink that accepts the connection and never answers, and asserts the process still exits 0 within the bound. `watch` is unaffected because its handler promise resolves only on graceful shutdown. Commands that call `process.exit()` directly (the `*` default handler, the self-exiting synth paths) bypass this; nothing emits metrics on those paths in this slice, and S3 flushes explicitly before each of them.

Three consequences of the explicit exit, each found by running the built bundle and each covered by a unit test and a child-process test:

- **Stdio is drained before the exit.** `process.exit` right after a large write into a pipe with a slow reader truncated stdout at 64 KiB on macOS (`cdktn convert | less`, `cdktn output --json | jq`), something the old natural loop drain never hit. `runCli` now waits for a zero-length write on stdout and stderr to complete before exiting, and tolerates a reader that closed early (`cdktn --help | head -1` exits 0 instead of dying on EPIPE).
- **Shell completion still prints.** yargs does not await an asynchronous completion callback before `parseAsync` resolves, so the explicit exit cut `cdktn deploy <TAB>` off with no output. On a completion request `runCli` returns early and lets the loop drain naturally; no handler runs there, so there is nothing to flush.
- **`watch` stays alive until its session is aborted.** `watch()` in cli-core resolved right after kicking off the first run, which was harmless while the process lived on the event loop but would have ended the session under an explicit exit. It now stays pending until its abort signal fires, with a test.

**Start with** `packages/cdktn-cli/src/bin/error-handling.ts` and the `runCli` wiring in `packages/cdktn-cli/src/bin/cdktn.ts`; those two files are the whole change, everything else is a consequence.

Reading order for this slice:

1. `packages/cdktn-cli/src/bin/error-handling.ts`
2. `packages/cdktn-cli/src/bin/cdktn.ts` and `packages/cdktn-cli/src/bin/exit-flush.ts`
3. `packages/@cdktn/cli-core/src/lib/models/terraform-cli.ts`
4. `packages/cdktn-cli/src/bin/__tests__/error-handling.test.ts`, `error-handling.integration.test.ts` and `exit-flush.test.ts`

### Test plan

- [x] Unit tests green across `@cdktn/commons`, `@cdktn/cli-core` and `cdktn-cli` on this slice against S1
- [x] #360 / #361 reproductions: the corrupt-`cdk.tf.json` trigger reaches `Debug Information:` and prints once; the undeclared-resource trigger prints a real message and stack with no `ERR_UNHANDLED_REJECTION` and no `PromiseRejectionHandledWarning`
- [x] Regression guards: the `.fail` handler is asserted synchronous, `reportFailure` is asserted to run exactly once, and there is exactly one `process.exit` call
- [x] Child-process integration test points a real Sentry client at a local sink and asserts the crash envelope arrives before exit
- [x] Child-process test against a sink that never responds: the success path still exits 0 within the flush bound, with no unhandled-rejection output
- [x] `terraform-cli` throws an `Error`, covered by `terraform-cli.test.ts` and the `cdktf-project.test.ts` snapshot line
- [x] Exit codes measured per failure class before and after, unchanged

- `packages/cdktn-cli/src/bin/cmds/get.ts` calls `readConfigSync()` at module import, so an invalid `cdktf.json` crashes before the failure reporter runs. Pre-existing, and out of scope here.

### Checklist

### DIFF
--- a/packages/@cdktn/cli-core/src/lib/models/terraform-cli.ts
+++ b/packages/@cdktn/cli-core/src/lib/models/terraform-cli.ts
@@ -455,7 +455,12 @@ export class TerraformCli implements Terraform {
       exitCode !== 0 &&
       !snapshot.context.cancelled // don't fail if we cancelled the run
     ) {
-      throw `Invoking Terraform CLI failed with exit code ${exitCode}`;
+      // A plain Error, not Errors.External: a non-zero terraform exit is
+      // exactly what the entrypoint's debug collection exists for, and only
+      // untyped errors trigger it.
+      throw new Error(
+        `Invoking Terraform CLI failed with exit code ${exitCode}`,
+      );
     }
 
     return { cancelled: Boolean(snapshot.context.cancelled) };
--- a/packages/@cdktn/cli-core/src/lib/watch.ts
+++ b/packages/@cdktn/cli-core/src/lib/watch.ts
@@ -149,10 +149,18 @@ export async function watch(
     }
   }
 
-  abortSignal.addEventListener("abort", () => {
-    logger.debug("Abort signal received, stopping watch");
-    watcher.close();
-    changeState({ type: "stopped" });
+  // The session ends only on abort: the CLI exits as soon as its handler
+  // returns, so resolving earlier would end `cdktn watch` after the first run.
+  const stopped = new Promise<void>((resolve) => {
+    abortSignal.addEventListener("abort", async () => {
+      logger.debug("Abort signal received, stopping watch");
+      const inFlight = state.type === "running" ? state.currentRun : undefined;
+      await watcher.close();
+      changeState({ type: "stopped" });
+      // hardAbort rejects the in-flight deploy; that ends the run, not the process
+      await inFlight?.catch(() => undefined);
+      resolve();
+    });
   });
 
   const onFileChange = () => {
@@ -178,4 +186,5 @@ export async function watch(
   onFileChange();
 
   await sendTelemetry("watch", { event: "start" });
+  await stopped;
 }
--- a/packages/@cdktn/cli-core/src/test/lib/cdktf-project.test.ts
+++ b/packages/@cdktn/cli-core/src/test/lib/cdktf-project.test.ts
@@ -421,7 +421,7 @@ describeIfDistExists(__dirname)("CdktfProject", () => {
         throw new Error("This error should not be thrown");
       } catch (e) {
         expect(e).toMatchInlineSnapshot(
-          `"Invoking Terraform CLI failed with exit code 1"`
+          `[Error: Invoking Terraform CLI failed with exit code 1]`
         );
       }
 
--- a/packages/@cdktn/cli-core/src/test/lib/watch.test.ts
+++ b/packages/@cdktn/cli-core/src/test/lib/watch.test.ts
@@ -1,0 +1,81 @@
+// Copyright (c) HashiCorp, Inc
+// SPDX-License-Identifier: MPL-2.0
+import { EventEmitter } from "events";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+
+const watcher = Object.assign(new EventEmitter(), {
+  close: jest.fn().mockResolvedValue(undefined),
+});
+jest.mock("chokidar", () => ({ watch: () => watcher }));
+
+// A deploy that runs until the project is hard-aborted, like a real one
+// waiting on terraform.
+let currentDeploy: { reject: (e: Error) => void } | undefined;
+jest.mock("../../lib/cdktf-project", () => ({
+  CdktfProject: class {
+    public stacksToRun = [];
+    public hardAbort = () => {
+      const e = new Error("aborted");
+      e.name = "AbortError";
+      currentDeploy?.reject(e);
+    };
+    public deploy = () =>
+      new Promise<void>((_resolve, reject) => {
+        currentDeploy = { reject };
+      });
+  },
+}));
+
+import { watch, State } from "../../lib/watch";
+
+const settled = (p: Promise<unknown>) =>
+  Promise.race([
+    p.then(() => true),
+    new Promise<boolean>((r) => setTimeout(() => r(false), 50)),
+  ]);
+
+describe("watch", () => {
+  const originalCwd = process.cwd();
+  let tmp: string;
+
+  beforeEach(() => {
+    tmp = fs.mkdtempSync(path.join(os.tmpdir(), "cdktn-watch-"));
+    fs.writeFileSync(
+      path.join(tmp, "cdktf.json"),
+      JSON.stringify({ language: "typescript", watchPattern: ["./**/*.ts"] }),
+    );
+    process.chdir(tmp);
+    watcher.close.mockClear();
+    currentDeploy = undefined;
+  });
+
+  afterEach(() => {
+    process.chdir(originalCwd);
+    fs.rmSync(tmp, { recursive: true, force: true });
+  });
+
+  // the CLI exits as soon as the handler returns, so an early resolution
+  // would end `cdktn watch` right after its first deploy started
+  it("stays pending until aborted, then stops the watcher and settles the in-flight run", async () => {
+    const ac = new AbortController();
+    const states: State["type"][] = [];
+    const session = watch(
+      { synthCommand: "true", outDir: "out", onUpdate: () => {} },
+      { autoApprove: true },
+      ac.signal,
+      (state) => states.push(state.type),
+    );
+
+    expect(await settled(session)).toBe(false);
+    expect(states).toEqual(["running"]);
+    expect(watcher.close).not.toHaveBeenCalled();
+
+    ac.abort();
+
+    await session;
+    expect(watcher.close).toHaveBeenCalledTimes(1);
+    expect(states[states.length - 1]).toBe("stopped");
+  });
+});
--- a/packages/@cdktn/cli-core/src/test/models/terraform-cli.test.ts
+++ b/packages/@cdktn/cli-core/src/test/models/terraform-cli.test.ts
@@ -59,7 +59,7 @@ describe("terraform-cli", () => {
       // The exit code is produced as the machine's final-state output and read off the settled snapshot.
       jest.mocked(spawnInteractive).mockImplementation(silentPty(3));
 
-      await expect(makeCli().deploy({}, () => {})).rejects.toMatch(
+      await expect(makeCli().deploy({}, () => {})).rejects.toThrow(
         /exit code 3/,
       );
     });
--- a/packages/cdktn-cli/src/bin/__tests__/error-handling.integration.test.ts
+++ b/packages/cdktn-cli/src/bin/__tests__/error-handling.integration.test.ts
@@ -41,8 +41,29 @@ function fixtureSource(errorHandlingPath: string): string {
     });
   }
 
+  // cdktn.ts' completion function, minus the manifest: answers for "diff"
+  // only after a turn of the event loop
+  const customCompletion = function (_current, argv, completionsFilter, done) {
+    if (argv._.includes("diff")) {
+      completionsFilter(async (_err, defaults) => {
+        await new Promise((resolve) => setTimeout(resolve, 200));
+        done([...defaults, 'alpha:target stack "alpha"']);
+      });
+    } else {
+      completionsFilter();
+    }
+  };
+
   const cli = yargs(process.argv.slice(2).filter((a) => a !== "--with-listener"))
     .exitProcess(false)
+    .completion("completion", customCompletion)
+    .command(
+      "diff [stack]",
+      "diffs a stack",
+      (cmdYargs) =>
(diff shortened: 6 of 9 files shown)


jsteinich (commented):
Reviewed as part of the full `tele/s1`–`tele/s6` stack. No findings — this is a real fix that stands on its own merit, independent of the telemetry work, and it's worth landing early in the stack.

Things I specifically checked and was happy with:

- Keeping the `.fail` callback synchronous, with the `y.exit(1, error)` before recording the failure, is the right shape. The comment explaining *why* it must stay sync (yargs discards the return value, so an `await` races Node's unhandled-rejection reporter) is exactly the kind of note that stops someone "cleaning this up" into a regression later.
- The `if (!failure)` guard in the `catch` around `parseAsync` correctly handles both directions: async handler rejections that `.fail` already recorded, and synchronous handler throws that bypass `.fail` entirely.
- `describeError` handles the raw-string and bare-object throws that this codebase actually produces — including `terraform-cli.ts`, which this PR converts from a bare `throw "..."` to a real `Error`. The `JSON.stringify` returning `undefined` case is covered by the `?? String(e)`.
- The `.catch()` on the outer IIFE so `runCli` itself can never reject.

One thing worth being explicit about in the description, since it's a genuine behaviour change rather than a pure refactor: the old handler called `Sentry.close(4000)` but never captured the error, so unexpected errors reaching the entrypoint were *not* reported. `reportFailure` now calls `deps.captureException` for both `External` and unexpected errors. That's an improvement and it's correctly gated (Sentry is only initialised after consent, so `captureException` is inert otherwise) — it just means more crash events than before, which is worth a line in the PR body.

Minor, non-blocking: `createBeforeExitFlush` only runs on the natural-exit path. Several places still call `process.exit()` directly — the `.command("*")` default handler in `cdktn.ts` is one — and `beforeExit` doesn't fire for those. It doesn't matter yet since nothing emits metrics before those exits, but it becomes load-bearing in #411 once metrics start flowing.

so0k:
Thanks for the close read. Two changes on top of what you reviewed:

- The description now calls out the crash-reporting behaviour change explicitly: `reportFailure` captures External and unexpected errors, which the old handler never did, and it stays consent-gated.
- Your note on the success-path flush turned out to be more than minor. `beforeExit` cannot fire while a pending Sentry request keeps the event loop alive, which is precisely the unresponsive-endpoint case the bound was meant to cover, so the 4000 ms timeout was never armed there. `runCli` now awaits the bounded flush after `parseAsync()` resolves without a failure and exits explicitly with `process.exitCode ?? 0`, and a child-process test points the fixture at a sink that accepts the connection and never answers, asserting the process still exits 0 within the bound.

Making the exit explicit surfaced three things the natural loop drain had been hiding, each fixed with a unit and a child-process test: stdout truncated at 64 KiB into a slow pipe reader (`cdktn convert | less`), so stdio is drained first and a closed reader no longer raises EPIPE; shell completion printed nothing because yargs does not await the async completion callback, so that path returns early; and `watch()` in cli-core resolved right after its first run, so it now stays pending until its session is aborted. The direct `process.exit()` sites you mention emit nothing in this slice, and #411 flushes explicitly before each one that does.

sakul-learning:
## Review verdict: APPROVE

Reviewed the current stacked-layer diff from `origin/tele/s1-sentry-sdk-10` to `8131b2158f296dc147812c61ab8e0baf04152f87`.

The prior blocker in `packages/cdktn-cli/src/bin/exit-flush.ts` is resolved: that natural-`beforeExit` flush path is gone, and `runCli` in `packages/cdktn-cli/src/bin/error-handling.ts` now awaits the bounded Sentry flush on successful parsing before explicitly exiting. The child-process test with a Sentry sink that accepts a connection without responding exercises the exact previously-broken lifecycle: the process exits within the bound instead of waiting indefinitely for the transport.

I also verified that the completion early-return matches yargs’ own `get-yargs-completions` detection, so asynchronous completion output is not preempted by the new explicit-exit path. The `watch()` change now keeps the command handler alive until the abort signal closes the watcher, preventing `cdktn watch` from exiting after its first run.

Configured validation completed successfully:
- `pnpm exec jest packages/cdktn/test/validations.test.ts --runInBand` — 41/41 tests passed.
- `pnpm exec nx build cdktn` — completed with 0 errors.
- `pnpm exec nx test cdktn --runInBand` — 43 suites / 542 tests passed.

No `.github/workflows` files changed. The complexity alert for the two anonymous `cdktf-project.test.ts` callbacks is a matcher artifact; this layer changes only the expected Error snapshot there. The `watch.ts` complexity increase is the small, correctness-motivated abort-lifecycle change.

_Review conclusion by GPT-5.6 Sol (OpenAI Codex)._

PR #412, opened 2026-09-09
feat(cli): report the installed binary, target versions and platform in usage metrics

so0k:
Part 4 of 6 in a stack; review order S1 → S6; base is the previous slice.

4. `feat(cli): report the installed binary, target versions and platform in usage metrics` (this PR)

### Related issue

Part of #48

### Description

S3 emits usage metrics carrying only `command`, `ci` and `language`, which does not answer the questions the metrics exist for: which Terraform or OpenTofu binary people actually run, which versions their projects declare they target, and on what platform. This slice adds that environment context to the base attribute set, so every metric from here on carries it.

What changes:

- `packages/@cdktn/commons/src/terraform.ts`: a lazy binary probe that runs at most once per process, with a 1500 ms ceiling and a timeout mapping to `unknown`, plus a seam so tests can drive it without spawning anything. The probe result is shared across copies of commons in the bundle via `globalThis`. `packages/@cdktn/commons/src/debug.ts` moves onto the same probe.
- `packages/@cdktn/commons/src/telemetry.ts`: `getBinaryAttributes` (bounded, release-only version, and a version only for a recognised product), `getProjectTargetAttributes` and `setProjectTargetAttributes`, `os` and `arch`, and `semverRangeOrInvalid`, which canonicalizes a declared range or reduces it to the literal `invalid`. The base attribute set grows accordingly.
- `packages/@cdktn/cli-core/src/lib/error-reporting.ts`: one `setProjectTargetAttributes(getProjectTargetAttributes(projectPath))` capture at init, so the target attributes are read once from the project that is actually running.

Every value here is validated before it becomes an attribute rather than forwarded: versions are reduced to `MAJOR.MINOR.PATCH` and dropped entirely for an unrecognised product, ranges are canonical or `invalid`, and nothing carries a path, a binary location or free text.

**Start with** `getBinaryAttributes` and `getProjectTargetAttributes` in `packages/@cdktn/commons/src/telemetry.ts`, then the probe in `packages/@cdktn/commons/src/terraform.ts` and its timeout handling. The exact-attribute-set test in `telemetry.test.ts` is the thing to check the result against: it pins the sorted key set of `cli.command.invoked`, so an accidental extra attribute fails.

Reading order for this slice:

1. `packages/@cdktn/commons/src/terraform.ts` (the probe and its seam)
2. `packages/@cdktn/commons/src/telemetry.ts` (the new attribute builders and the validators)
3. `packages/@cdktn/cli-core/src/lib/error-reporting.ts` (the single capture site)
4. `packages/@cdktn/commons/src/terraform.test.ts`, `debug.test.ts` and the exact-attribute-set and stamping cases in `telemetry.test.ts`

### What is collected

Added to the base attribute set of every metric (the S3 set of `command`, `ci` and `language` is unchanged and still applies). Still gated by `isUsageTelemetryEnabled()` and a Sentry DSN in the build; the never-sent list from S3 is unchanged by this slice.

| Attribute | Value | Source |
|---|---|---|
| `os`, `arch` | `process.platform` / `process.arch` | Node |
| `binary` | `terraform`, `opentofu`, `unknown`, `missing` | one lazy process-global probe, 1500 ms ceiling, timeout maps to `unknown` |
| `binary_version` | `MAJOR.MINOR.PATCH`, only when `binary` is `terraform` or `opentofu` | the probe; prerelease and build text dropped; an unrecognised product sends no version |
| `target_terraform`, `target_opentofu` | canonical semver range (for example `>=1.9.0`), else `invalid` | `targetVersions` in `cdktf.json`, else the CLI's `DEFAULT_TARGET_VERSIONS`; prerelease or build identifiers make it `invalid` |
| `targets_declared` | boolean | whether `cdktf.json` has a `targetVersions` block |
| `validate_installed_binary` | boolean | `validateInstalledBinary === true` in `cdktf.json` |

No new metrics are added in this slice.

### Test plan

- [x] Unit tests green across `@cdktn/commons`, `@cdktn/cli-core` and `cdktn-cli` on this slice against S3
- [x] Exact-attribute-set test: the sorted attribute key set of `cli.command.invoked` matches the documented set, so an extra attribute fails the build
- [x] Probe: runs at most once per process, honours the 1500 ms ceiling, a timeout yields `binary: unknown`, a missing binary yields `missing`, an unrecognised product yields no `binary_version`
- [x] Version and range validation: prerelease and build text dropped from versions; a non-canonical or prerelease range becomes `invalid`
- [x] Stamping test: the target attributes captured at init appear on metrics emitted later in the run
- [x] `targets_declared` and `validate_installed_binary` read from `cdktf.json` through the same forgiving reader as the consent flag

- ["dropped stack payload loses useful information"](https://github.com/open-constructs/cdk-terrain/pull/62#discussion_r3408182518), the target-versions half of the thread (so0k's follow-up): the declared Terraform and OpenTofu ranges are on every command metric now, alongside the installed binary and its version. The per-stack half of that thread is answered in S5.

### Checklist

### DIFF
--- a/packages/@cdktn/cli-core/src/lib/error-reporting.ts
+++ b/packages/@cdktn/cli-core/src/lib/error-reporting.ts
@@ -6,6 +6,8 @@ import {
   getProjectId,
   getUserId,
   getUsageTelemetryConsent,
+  getProjectTargetAttributes,
+  setProjectTargetAttributes,
   setUsageTelemetryEnabled,
   startCommandTelemetry,
   collectDebugInformation,
@@ -126,6 +128,7 @@ export async function initializErrorReporting(
   // directory: some commands (convert) chdir into a temporary project
   // before sendTelemetry runs and must not consult that project's flags.
   setUsageTelemetryEnabled(usageTelemetryEnabled);
+  setProjectTargetAttributes(getProjectTargetAttributes(projectPath));
 
   if (!crashReportingEnabled && !usageTelemetryEnabled) {
     logger.debug("Error reporting and usage telemetry disabled");
--- a/packages/@cdktn/cli-core/src/test/error-reporting.test.ts
+++ b/packages/@cdktn/cli-core/src/test/error-reporting.test.ts
@@ -30,6 +30,7 @@ jest.mock("@cdktn/commons", () => {
     ...actual,
     collectDebugInformation: jest.fn().mockResolvedValue({}),
     setUsageTelemetryEnabled: jest.fn(actual.setUsageTelemetryEnabled),
+    setProjectTargetAttributes: jest.fn(actual.setProjectTargetAttributes),
   };
 });
 
@@ -39,6 +40,8 @@ import {
   Errors,
   isUsageTelemetryEnabled,
   resetCommandTelemetry,
+  seedTerraformCliProbeForTests,
+  setProjectTargetAttributes,
   setUsageTelemetryEnabled,
 } from "@cdktn/commons";
 import {
@@ -85,12 +88,16 @@ function useReportingFixture() {
     delete process.env.CHECKPOINT_DISABLE;
     process.env.SENTRY_DSN = "https://public@example.invalid/1";
     ciInfoMock.isCI = false;
+    // the start-of-command metric stamps the binary; keep it off the machine
+    seedTerraformCliProbeForTests(Promise.resolve("Terraform v1.9.0\n"));
   });
 
   afterEach(() => {
     setUsageTelemetryEnabled(undefined);
     resetCommandTelemetry();
     Errors.setScope("unknown");
+    setProjectTargetAttributes(undefined);
+    seedTerraformCliProbeForTests();
     process.chdir(originalCwd);
     fs.removeSync(workdir);
     Object.defineProperty(process.stdout, "isTTY", {
@@ -287,15 +294,22 @@ describe("consent gating (initializErrorReporting)", () => {
       expect(Sentry.init).toHaveBeenCalledTimes(1);
     },
   );
-  it("captures the usage decision while still in the user's cwd", async () => {
+  it("captures the usage decision and the project targets while still in the user's cwd", async () => {
     fs.writeJsonSync(path.join(workdir, "cdktf.json"), {
       sendCrashReports: false,
+      targetVersions: { terraform: ">=1.9.0" },
+      validateInstalledBinary: true,
     });
     setInteractive(false);
 
     await initializErrorReporting();
 
     expect(setUsageTelemetryEnabled).toHaveBeenCalledWith(true);
+    expect(setProjectTargetAttributes).toHaveBeenCalledWith({
+      targets_declared: true,
+      validate_installed_binary: true,
+      target_terraform: ">=1.9.0",
+    });
   });
   it("reads and persists against an explicit project path, not the cwd", async () => {
     // init creates the project in a destination directory and initializes
@@ -309,6 +323,7 @@ describe("consent gating (initializErrorReporting)", () => {
     fs.writeJsonSync(path.join(destination, "cdktf.json"), {
       sendCrashReports: false,
       sendUsageTelemetry: true,
+      targetVersions: { terraform: ">=1.9.0" },
     });
     setInteractive(true);
     const crashPrompt = jest.fn();
@@ -319,6 +334,9 @@ describe("consent gating (initializErrorReporting)", () => {
     expect(crashPrompt).not.toHaveBeenCalled();
     expect(usagePrompt).not.toHaveBeenCalled();
     expect(setUsageTelemetryEnabled).toHaveBeenCalledWith(true);
+    expect(setProjectTargetAttributes).toHaveBeenCalledWith(
+      expect.objectContaining({ target_terraform: ">=1.9.0" }),
+    );
     expect(isUsageTelemetryEnabled()).toBe(true);
     // usage-only consent: the client exists but drops error events
     await expect(initOptions().beforeSend({}, undefined)).resolves.toBeNull();
--- a/packages/@cdktn/commons/src/debug.test.ts
+++ b/packages/@cdktn/commons/src/debug.test.ts
@@ -4,7 +4,7 @@ import * as fs from "fs";
 import * as path from "path";
 
 jest.mock("./terraform", () => ({
-  terraformVersion: Promise.resolve("1.7.5"),
+  terraformVersion: () => Promise.resolve("1.7.5"),
 }));
 
 import { collectDebugInformation } from "./debug";
--- a/packages/@cdktn/commons/src/debug.ts
+++ b/packages/@cdktn/commons/src/debug.ts
@@ -404,7 +404,7 @@ export async function collectDebugInformation() {
     debugOutput["constructs"] = (await constructs) ?? null;
     debugOutput["jsii"] = (await jsii) ?? null;
   }
-  debugOutput["terraform"] = await terraformVersion;
+  debugOutput["terraform"] = (await terraformVersion()) ?? null;
   debugOutput["arch"] = os.arch();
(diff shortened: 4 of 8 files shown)


jsteinich (commented):
Reviewed as part of the full `tele/s1`–`tele/s6` stack. Breaking the `errors.ts` → `terraform.ts` cycle so `telemetry.ts` can be imported from `errors.ts` is a nice bit of untangling, and the process-global `Symbol.for` probe key is the right answer to the "bundle carries several copies of this module" problem.

One real issue and one smaller one.

### The 1500 ms bound doesn't bound what the docstring says it does

`telemetry.ts:149`:

```ts
 * Binary attributes from the version probe, bounded so a hung binary never
 * delays the command; a timed-out probe reports `binary: "unknown"`.
```

The `Promise.race` bounds the **attribute**, not the process. `exec` in `commons/src/util.ts` uses `spawn` with no `unref()` and no kill-on-timeout, so a hung `terraform version` leaves a live child handle on the event loop: the race resolves at 1500 ms, the metric gets `binary: "unknown"`, and then the CLI sits there until the child exits. `beforeExit` won't fire while that handle is active, so `createBeforeExitFlush` doesn't run either — the command hangs at exit rather than degrading gracefully.

Either kill the child when the timer fires, or `unref()` the probe's child so it can't hold the loop open. Worth fixing because "the version probe hangs" is exactly the scenario the timeout was written for.

### The probe now runs on commands that never needed the binary

Related but separate: `getBinaryAttributes()` is awaited unconditionally from `sendTelemetry` (`telemetry.ts:229`), so `init`, `get`, `convert` and `synth` now spawn `terraform version` purely to populate a metric attribute. It's memoised per process via the `Symbol.for` key so it's one spawn, not many, and it's correctly behind the `isUsageTelemetryEnabled()` early-return — but it's still a new subprocess on commands that previously touched no binary at all, and on the failure path in `reportFailure` it can add up to 1500 ms before the process exits.

Not necessarily wrong, but worth a deliberate decision: is `binary` / `binary_version` valuable enough on an `init` or `convert` metric to justify spawning for it? Restricting the probe to the commands that already shell out would remove both this and most of the exposure above.

### Minor: fabricated error prose as a success value

`terraform.ts:71-77`:

```ts
.catch(
  (err) =>
    `Error: Usage Error: Unknown: Error loading terraform version ${err}`,
);
```

Resolving to a string that *impersonates* the old `Errors.Usage` output is a smell — the `"Error: Usage Error: "` prefix now exists only to make `cdktn debug` print something that looks like it used to. Since the only consumer is `debug.ts:407` (which already does `?? null`), returning `undefined` and letting `debug.ts` render the failure message would be clearer and wouldn't leave a fake error string floating around for someone to pattern-match on later.

(The `Promise` → function change itself is fine — all in-repo consumers are updated.)

jsteinich (commented):
Re-reviewed at `7adca3969`. One small new point, from the `processState` helper that landed in #411 below this layer.

`terraform.ts:35` still hand-rolls the same globalThis pattern the helper now exists for:

```ts
const PROBE_KEY = Symbol.for("cdktn.terraformCli");
```

`errors.ts` was converted to `processState("cdktn.errorScope", ...)` in the same round, so this is the last remaining copy. Worth folding in for consistency while the layer is still open — it also means the "one spawn per process across bundle copies" guarantee is expressed the same way as the other two pieces of cross-copy state, rather than looking like an unrelated mechanism.

so0k:
@jsteinich on "the probe now runs on commands that never needed the binary":

The probe isn't new work for those commands. Before this layer, `@cdktn/commons/src/terraform.ts` ran `terraform version -json` at import, once per commons copy. That's 3× on `convert`, `debug`, `synth`, `list` and `provider list`, and 1× on `--version`, with or without consent. `terraformCheck()` added `terraform -v` on 11 commands. With this layer, consent off runs zero probes (only `terraformCheck`'s `-v`), and consent on runs one, shared by every commons copy. `convert` goes from 3 to 0. The hang case also improves: no orphaned processes once the probe is killed at 1500 ms, and `debug` is bounded where it previously waited indefinitely.

full probe results on binary runs before and after this PR (from agent session):
image

jsteinich (commented):
Thanks — you're right about the spawn counts and I'll withdraw that half of the finding. I'd missed that the old `terraform.ts` exported `terraformVersion` as a module-scope `exec(...)`, so it ran at *import* rather than on use. Per-copy at import is strictly worse than once-per-process on demand, and gating it behind consent means consent-off now runs zero. That's a net improvement, not a regression.

Two corrections to the record, both small.

**There is no kill.** `grep -nE "\.kill\(|unref\(|AbortController"` over `packages/@cdktn/commons/src` returns nothing, and `getBinaryAttributes`' `finally` block is `clearTimeout(timer)` — which clears the timer, not the child. `Promise.race` resolves the *attribute* at 1500 ms; the `spawn` from `exec` is untouched and still holds a handle.

**My "hangs at exit" framing is superseded — by #410, not by this layer.** A hung probe no longer delays the CLI because `runCli` now ends the success path with an explicit `process.exit(process.exitCode ?? 0)` after the flush and `drainStdio`, and `process.exit` does not wait on live handles. So the docstring's "bounded so a hung binary never delays the command" is true today, but for a reason that lives in another PR rather than anything in this function.

What's left is much milder than I first said: a genuinely hung binary leaves one orphaned child after the CLI exits, reparented and harmless in the normal case. Not worth blocking on. Either add the `.kill()` the docstring implies, or reword it to say the *attribute* is bounded and the process exit is what actually caps the wait — I'd slightly prefer the latter, since it keeps the honest version of the guarantee next to the code that relies on it.

(The `PROBE_KEY` point from my other comment still stands independently — `errors.ts` moved to `processState`, `terraform.ts:35` is the last hand-rolled copy.)

so0k:
Thanks for the re-review, and for the two corrections — both are right, and I went with the kill rather than the reworded docstring.

**The bound now reaches the child** — `fix(commons): bound the terraform version probe with exec's timeout`. `versionOutput()` passes `{ timeout: PROBE_TIMEOUT_MS }` (1500, named once) to `exec`, which forwards it to spawn, so the child is killed with SIGTERM and the probe settles on its own. The `Promise.race`, its timer and the `timeoutMs` parameter are gone from `getBinaryAttributes`, and its docstring is now just "Binary attributes from the version probe" — no guarantee stated where it isn't implemented. `terraformCli()` reports `missing` only for a spawn error with `code === "ENOENT"`; a timed-out or otherwise failed `version` reports `unknown`.

Measured against real fixtures: a binary that hangs resolves `{ name: "unknown" }` after ~1505 ms with no child left behind (`pgrep` clean), and an empty PATH resolves `{ name: "missing" }` in 2 ms.

And you're right that #410's explicit `process.exit` is what actually caps the wait today — with the kill in place, the orphan goes away too, so neither the docstring nor the behaviour leans on the other PR any more.

One known gap, small enough that I left it: if `terraform` on PATH is a shell wrapper that runs the real binary as a child rather than `exec`ing it, the kill only reaches the wrapper; its child keeps stdout open and `exec()` settles on `close`, so the probe still waits for it (measured ~11 s). Closing that means changing `exec` in `util.ts` to settle on `exit` or to kill the process group, which is wider than this layer. tfenv, asdf and mise shims all exec or symlink, so they aren't affected.

**Fabricated error string** — `fix(commons): report no terraform version in debug when the probe fails`. `terraformVersion()` resolves to `undefined` on failure, `debug.ts` keeps its `?? null`, and `cdktn debug` prints `"terraform": null`.

**PROBE_KEY** — `refactor(commons): keep the terraform probe in process state`. The probe now uses `processState("cdktn.terraformCli", ...)`, like `cdktn.errorScope`, so the hand-rolled symbol is gone. Same key, so it's still one spawn per process across the bundle copies, and `seedTerraformCliProbeForTests` seeds and clears through the same object.

Also in this push: the consent prompt now lists what this layer adds — "(command, language, timing, Terraform/OpenTofu binary and version, target versions, OS and architecture)".

PR #421, opened 2026-09-10
fix(cli): make `cdktn get` remove stale files for resources removed from a provider

### LINKED ISSUE #384: cli: cdktn get leaves stale files for resources removed from a provider
### Expected Behavior

After `cdktn get`, the generated output directory should contain files for exactly the resources present in the provider version pinned in `cdktf.json`. When a provider version change removes or renames a resource, its generated files should be removed too, so the emitted code always matches the pinned schema.

### Actual Behavior

Cleanup happens only at _provider folder_ granularity, so per-resource directories for removed resources are never reconciled against the new schema. The output directory becomes a **union of every provider version it has ever been generated against** rather than a reflection of the pinned one.

The barrel files are regenerated correctly and no longer export the removed resources, so the idiomatic import fails loudly (`TS2305: has no exported member 'streamWorkspace'`). What remains:

- Orphaned, fully-formed construct files accumulate in source control with every provider version change (70 files in our repo).
- Their doc comments and registry links reference the old provider version.
- `constraints.json` and `versions.json` update correctly to the new pin, so the metadata claims a version the tree does not match — with no warning.

### Steps to Reproduce

Reproduced from scratch on a clean, minimal npm project. Upgrading `mongodb/mongodbatlas` from 1.41.1 to 2.16.0 is a convenient case, since 2.x drops a handful of resources that 1.41.1 has.

Project setup:

```jsonc
// package.json
{ "dependencies": { "cdktn": "0.24.0", "cdktn-cli": "0.24.0", "constructs": "10.6.0" } }

// cdktf.json
{
  "language": "typescript",
  "app": "npx ts-node main.ts",
  "codeMakerOutput": "bindings",
  "terraformProviders": ["mongodb/mongodbatlas@1.41.1"],
  "terraformModules": []
}
```

1. `npx cdktn get` — generates a clean 1.41.1 tree.
2. Change the pin in `cdktf.json` to `mongodb/mongodbatlas@2.16.0`.
3. `npx cdktn get` again, with no flags.

```console
$ npx cdktn get                                   # on 1.41.1
$ ls bindings/providers/mongodbatlas | wc -l
179                     # correct for 1.41.1

# change the pin to mongodb/mongodbatlas@2.16.0
$ npx cdktn get                                   # no flags
$ ls bindings/providers/mongodbatlas | wc -l
230                     # 2.16.0 alone generates 219

$ ls bindings/providers/mongodbatlas/teams
README.md  index.ts     # 'teams' was removed in 2.x, still on disk

$ grep -c "'./teams'" bindings/providers/mongodbatlas/lazy-index.ts
0                       # orphaned: on disk, not exported

$ grep mongodbatlas bindings/constraints.json
    "mongodbatlas": "2.16.0"        # metadata says 2.16.0...

$ cat bindings/versions.json
{ "registry.terraform.io/mongodb/mongodbatlas": "2.16.0" }   # ...and so does this
```

Both metadata files are updated to the new pin while the tree still holds 1.41.1-only resources, so nothing on disk indicates the mismatch.

The 11 orphaned entries are the resources 2.16.0 dropped: `teams`, `data-mongodbatlas-teams`, `data-lake-pipeline` (+ its 4 data sources), `privatelink-endpoint-serverless`, `privatelink-endpoint-service-serverless` (+ its 2 data sources).

`npx cdktn get --force` then correctly reduces the tree to 219 directories and removes all 11.

### Versions

language: typescript
cdktn-cli: 0.24.0
node: v24.16.0
cdktn: 0.24.0
constructs: 10.6.0
terraform: v1.5.5
arch: arm64
os: darwin 25.6.0

### Migration Context

Not sure / don't remember

### Upstream cdktf Issue

_No response_

### Providers

N/A

### Gist

_No response_

### Possible Solutions

Reconcile at resource granularity in the default path: after generating a provider, remove resource subdirectories under `providers/<name>/` that are not in the freshly-read provider schema. That makes `cdktn get` converge on the pinned schema without requiring `--force`.

Failing that, two cheaper improvements:

- Have `filterAlreadyGenerated` treat any provider **version change** as requiring regeneration-with-cleanup, so a version change cannot leave a stale tree behind a matching `constraints.json`.
- Warn when `providers/<name>/` contains resource directories absent from the schema, so the drift is visible in CI rather than silent until apply.

### Where the behaviour comes from

In `packages/@cdktn/cli-core/src/lib/get.ts`, the non-`cleanDirectory` path calls `removeFoldersThatShouldNotExist(constraints)`:

```ts
if (cleanDirectory) {
  await fs.remove(constructsOptions.codeMakerOutput);
} else {
  // Remove all providers that are not in the new list
  await constructsMaker.removeFoldersThatShouldNotExist(constraints);
  ...
}
```

`removeFoldersThatShouldNotExist` (`packages/@cdktn/provider-generator/src/get/constructs-maker.ts`) only reconciles the top level of `providers/`:

```ts
const shouldExist = constraintsThatShouldExist.some(
  (constraint) => constraint.name === folder,
);
if (!shouldExist) {
  fs.removeSync(path.resolve(this.codeMakerOutdir, "providers", folder));
}
```

A provider still listed in `terraformProviders` always satisfies `shouldExist`, so its folder is kept wholesale and its per-resource subdirectories are never reconciled against the new schema. Generation then overwrites the resources that still exist and leaves the rest untouched. Nothing in the pipeline diffs "resource dirs on disk" against "resources in the schema".

A second factor makes this stickier than a one-off: `filterAlreadyGenerated` skips regeneration entirely when the recorded constraint matches and the provider folder exists —

```ts
const providerExists = constraintMatches && providerFolderExists;
return !providerExists;
```

Once the stale tree is committed with a matching `constraints.json`, later `cdktn get` runs treat the provider as up to date and no-op, so the staleness cannot self-heal. It does regenerate when the CDKTN version differs, which is how we finally noticed — during a CDKTN upgrade.

### Workarounds

`--force` maps to `cleanDirectory` (`packages/cdktn-cli/src/bin/cmds/ui/get.ts`), which does a full `fs.remove` of the output directory and does avoid the problem. Caveats:

- It is documented as "Regenerates all generated constructs", which reads as a cache-busting/performance flag rather than the correctness-preserving mode. Nothing hints that omitting it can leave orphaned files behind.
- It is opt-in, so the default is the unsafe path.
- Full removal is blunt: it discards the entire output tree, including any sibling files a workspace keeps alongside the generated code, and forces regeneration of every provider rather than reconciling only the affected one.

### Anything Else?

_No response_

### References

_No response_

### Help Wanted

### Community Note

X-Guardian:
### Related issue

Fixes #384

### Description

`cdktn get` now removes generated resource directories that a previous run produced but the current one did not, so the output tree matches the provider version pinned in `cdktf.json`. Cleanup previously reconciled only the top level of `providers/`, leaving the per-resource directories inside a still-listed provider untouched.

**How the pruning decides what to keep.** `TerraformProviderGenerator.emitProvider` already builds a `files` array of everything it emitted, and derives the barrel-file exports from it. That same array now also populates a new `emittedResourceFolders` map, which `ConstructsMaker` accumulates alongside the existing `versions` map. `ConstructsMaker.removeOrphanedResourceFolders` then lists each regenerated provider folder and removes any subdirectory absent from that map.

Deriving the keep-set from the emitted files — rather than re-deriving it from the schema — means pruning cannot disagree with the barrel files: a directory is kept exactly when it is exported. To keep that guarantee explicit, the duplicated `file.replace(...)` path-to-directory-name logic in `emitIndexFile` and `emitLazyIndexFile` was extracted into a shared `resourceFolderName` helper that all three call sites now use. This is a pure refactor — the 100 existing generator snapshots are unchanged.

**Scope of the deletion.** Three constraints keep it conservative:

- **TypeScript only.** `providers/<name>/<resource>/` maps one-to-one onto the emitted index exports. Other languages co-locate providers and modules in the same folders, which makes a dropped resource indistinguishable from a hand-written construct — the same reasoning `removeFoldersThatShouldNotExist` already applies.
- **Regenerated providers only.** `filterAlreadyGenerated` can skip a provider whose pin and folder are both unchanged. Such a provider has no `emittedResourceFolders` entry and is therefore never walked, so an up-to-date tree is left alone.
- **Directories only.** Provider-level files (`index.ts`, `lazy-index.ts`) sit alongside the resource directories and are skipped by an `isDirectory()` check.

It runs after `save()`, once CodeMaker has written the current run's files, so the walk sees the new tree and removes only what is genuinely left over. A provider folder that cannot be listed is logged at debug level and skipped rather than failing the run.

A plain `cdktn get` therefore converges on the same tree as `cdktn get --force`, without paying the cost of regenerating every provider from scratch.

### Checklist


so0k (approved):
Recommendation as posted: land #421 as-is for this issue

PR #74, opened 2026-03-26
fix(cli): Pin init template to TypesScript 5.x

jsteinich:
### Description

Currently initializing a TypeScript project using TypeScript 6 causes issues due to `ts-jest` not yet supporting it. (See https://github.com/kulshekhar/ts-jest/issues/5232).

This is a short term solution to unblock the next release. Hopefully the upstream dependency will be fixed, but if not we can look into alternative solutions.

### Checklist

- [x] I have updated the PR title to match [CDKTN's style guide](https://github.com/open-constructs/cdk-terrain/blob/main/CONTRIBUTING.md#pull-requests-1)
- [x] I have run the linter on my code locally
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the [documentation](https://github.com/open-constructs/cdk-terrain/tree/main/website) if applicable
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works if applicable
- [x] New and existing unit tests pass locally with my changes

PR #422, opened 2026-09-11
feat(cli): add typescript-pnpm and typescript-yarn init templates

X-Guardian:
### Description

This PR adds `typescript-pnpm` and `typescript-yarn` templates to `cdktn init`, so a project can be scaffolded with the package manager of choice. Until now `cdktn init --template=typescript` hardcoded npm and the generated `package.json` scripts, `help` text and post-convert `npm run get` all assumed npm.

**Selected through `--template`.** This follows the existing `python` / `python-pip` precedent of splitting a language on its packaging tool, and keeps the variants scriptable and visible in the interactive picker with no extra code — template discovery is already a `readdirSync` of the templates directory.

**Stored as overlays.** Only 3 of the 10 TypeScript template files differ per package manager; the other 7 include `jest.config.js` (6.6 KB) and `__tests__/main-test.ts`. Duplicating those three ways would mean every future edit to `tsconfig.json` or `main.ts` landing three times, so `typescript/` stays the complete base and each variant carries only its own files:

```
templates/
  typescript/            <- full base, 10 files (npm)
  typescript-pnpm/       <- help, package.json, .hooks.sscaff.js
  typescript-yarn/       <- the same three, plus .yarnrc.yml and .gitignore
```

`init()` scaffolds the base first, then overlays the variant. sscaff writes with plain `fs.writeFile`, so the second pass simply wins. The base pass receives `isOverlayBase`, which stops its `post` hook before the install step and leaves dependency installation to the overlay — so a pnpm project gets a `pnpm-lock.yaml` and nothing else.

**Two places keyed off the literal `"typescript"`** and now resolve through the template family / overlay base instead: the `--from-terraform-project` gate in `helper/init.ts`, so HCL conversion is offered for every TypeScript variant, and `distKeysByTemplate` in `cli-core/init.ts`, so `--dist` resolves each variant to its base's artifact.

**Yarn needs two things pnpm does not.** A corepack `yarn` shim resolves to the end-of-life 1.22 line until a project declares a version, so the `packageManager` pin is written *before* installing — that is what selects a modern Yarn (verified: Yarn 4.18.0 performs the install). The version is resolved at scaffold time from `@yarnpkg/cli` on the registry, so it tracks Yarn releases rather than going stale, and the lookup keeps working on Node 26+ where corepack is no longer bundled — the template declares `node >=22.0`. Where corepack is absent the pin records intent but cannot switch versions, so the hook reports which Yarn actually ran. Yarn also defaults to Plug'n'Play, so `.yarnrc.yml` sets `nodeLinker: node-modules` to keep `npx tsx main.ts` and Jest resolving, with a variant `.gitignore` for `.yarn/install-state.gz`.

**`cdktn provider add` now agrees with what `init` scaffolded.** `NodePackageManager` previously probed only for `yarn.lock`, so a pnpm project got `npm install` and ended up with a `package-lock.json` beside its `pnpm-lock.yaml`. Detection now reads the `packageManager` field first — the strongest signal, and what these templates write — then `pnpm-lock.yaml`, then `yarn.lock`, then npm. Quiet flags are emitted per manager, since pnpm has no `--no-progress` and Yarn Berry errors on unknown options, accepting neither `--silent` nor `--no-progress`.

**Four pre-existing bugs fixed in passing**, each a one-liner encountered while testing this path:

- `helper/init.ts` built an `ExecSyncOptions` (including `stdio: "ignore"` under `--silent`) and then never passed it, so `--silent` leaked output from the post-convert `run get`.
- `init()` never passed `cdktfVersion` to `providerAdd`, so `cdktn init --cdktf-version=X --providers=aws` scaffolded fine and then failed with `cannot use version 0.0.0`.
- The `--from-terraform-project` warning said "the typescript template" where several now qualify.
- `jest.config.js` annotated its config with ts-jest's deprecated `InitialOptionsTsJest`; ts-jest's own annotation says to use `JestConfigWithTsJest`.

### Checklist

### DIFF
--- a/packages/@cdktn/cli-core/src/lib/dependencies/package-manager.ts
+++ b/packages/@cdktn/cli-core/src/lib/dependencies/package-manager.ts
@@ -194,6 +194,32 @@ const pipPackageSchema = z.array(
   z.object({ name: z.string(), version: z.string() }).loose(),
 );
 
+// [
+//   {
+//     "name": "my-project",
+//     "dependencies": {
+//       "@cdktn/provider-random": { "from": "@cdktn/provider-random", "version": "3.0.11" }
+//     }
+//   }
+// ]
+const pnpmListSchema = z.array(
+  z
+    .object({
+      dependencies: z.record(
+        z.string(),
+        z.object({ version: z.string() }).loose(),
+      ),
+      devDependencies: z.record(
+        z.string(),
+        z.object({ version: z.string() }).loose(),
+      ),
+    })
+    .partial()
+    .loose(),
+);
+
+type NodePm = "npm" | "pnpm" | "yarn";
+
 /**
  * manages installing, updating, and removing dependencies
  * in the package system used by the target language of a CDKTN
@@ -243,8 +269,42 @@ export abstract class PackageManager {
 }
 
 class NodePackageManager extends PackageManager {
-  private hasYarnLockfile(): boolean {
-    return existsSync(path.join(this.workingDirectory, "yarn.lock"));
+  private detectedPackageManager?: NodePm;
+
+  /**
+   * The package manager this project uses. An explicit `packageManager` field (corepack) wins over lockfile
+   * probing, since it is a deliberate declaration; the result is cached because `provider add` resolves it per
+   * provider.
+   *
+   * @internal
+   */
+  public get packageManager(): NodePm {
+    if (this.detectedPackageManager) {
+      return this.detectedPackageManager;
+    }
+
+    try {
+      const pkg = fs.readJsonSync(
+        path.join(this.workingDirectory, "package.json"),
+      );
+      const declared = String(pkg?.packageManager ?? "").split("@")[0];
+      if (declared === "pnpm" || declared === "yarn" || declared === "npm") {
+        this.detectedPackageManager = declared;
+        return declared;
+      }
+    } catch {
+      // no or unreadable package.json - fall through to lockfile probing
+    }
+
+    if (existsSync(path.join(this.workingDirectory, "pnpm-lock.yaml"))) {
+      this.detectedPackageManager = "pnpm";
+    } else if (existsSync(path.join(this.workingDirectory, "yarn.lock"))) {
+      this.detectedPackageManager = "yarn";
+    } else {
+      this.detectedPackageManager = "npm";
+    }
+
+    return this.detectedPackageManager;
   }
 
   public async addPackage(
@@ -254,21 +314,21 @@ class NodePackageManager extends PackageManager {
   ): Promise<void> {
     console.log(`Adding package ${packageName} @ ${packageVersion}`);
 
-    // probe for package-lock.json or yarn.lock
-    let command = "npm";
-    let args = ["install"];
+    const command = this.packageManager;
+    const args = [command === "npm" ? "install" : "add"];
 
-    if (this.hasYarnLockfile()) {
-      command = "yarn";
-      args = ["add"];
-    }
     args.push(
       packageVersion ? packageName + "@" + packageVersion : packageName,
     );
 
+    // Quiet flags differ per manager: pnpm has no --no-progress, and Yarn Berry errors on unknown options - it has
+    // neither --silent nor --no-progress (Yarn 1 accepted both).
     if (silent) {
-      args.push("--silent");
-      args.push("--no-progress");
+      if (command === "npm") {
+        args.push("--silent", "--no-progress");
+      } else if (command === "pnpm") {
+        args.push("--silent");
+      }
     }
 
     // Install exact version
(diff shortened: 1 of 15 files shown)


sakul-learning:
Reviewed `648dc2acbbc3`. Builds and the configured test targets pass; one path introduced by the new yarn template is broken.

**`typescript-yarn`: provider listing goes through a Yarn 1-only command**

`templates/typescript-yarn/.hooks.sscaff.js` pins `packageManager: yarn@<current 4.x>` into the generated `package.json`, and the new detection getter in `packages/@cdktn/cli-core/src/lib/dependencies/package-manager.ts:281` prefers that field, so `listProviderPackages()` (`:444`) routes yarn projects to `listYarnPackages()` (`:356`), which runs `yarn list --json`. Yarn 2+ (Berry) removed `list`:

```
$ yarn list --json
Usage Error: Couldn't find a script named "list".
```

`exec` in `@cdktn/commons` rejects on a non-zero exit, so `listYarnPackages` throws, `getCurrentlyInstalledVersion` rethrows it as `Failed to list packages:` (`dependency-manager.ts:193`), and `allProviders` (`:503`) propagates. In a project scaffolded with `--template=typescript-yarn`, `cdktn provider list`, `cdktn provider upgrade` and `cdktn debug` all fail. `cdktn provider add` is unaffected — it never lists, and `yarn add -E` does work on Berry.

It only fails in the configuration the template is aiming for: without corepack the pin cannot switch versions, Yarn 1.22 performs the install, and `yarn list --json` works.

Berry does have an equivalent — `yarn info --name-only --all --json` returns `"@cdktn/provider-random@npm:15.0.0"`, one JSON string per line — so this needs a Berry branch plus an update to `yarnListSchema` (`:168`), which currently expects `{data:{trees:[…]}}`.

**Two smaller items**

1. `templates/typescript/.hooks.sscaff.js:9` adds `exports.pre = () => requirePackageManager("npm")`, and sscaff runs `pre` on every pass, so `--template=typescript-pnpm` and `--template=typescript-yarn` still require `npm` on `PATH` even though they install with pnpm/yarn. The `isOverlayBase` guard only exists in `post`; sscaff passes the variables to `pre`, so the check can move there.
2. The added tests cover detection only. The new per-manager quiet-flag branch in `NodePackageManager.addPackage` (`:324`) — npm pushes `--silent --no-progress`, pnpm `--silent`, yarn neither — and the pnpm listing path have no coverage. Both are pure argument construction and cheap to test.

so0k:
My main concern is that this built-in template list will keep growing and it creates confusion while the CDK ecosystems mainly leverages https://projen.io (afaik).

I can understand the "batteries included" approach the original CDKTF team took here, but I would rather replace it all leveraging projen instead so it sets people up into that ecosystem.

> NOTE: That said, the AWS PDK team just started breaking the projen heavy set up down into smaller re-usable components such as https://aws.amazon.com/about-aws/whats-new/2026/09/nx-plugin-for-aws/

## Extrapolating this PR

The code-level findings from this PR are in the bot's comment above. 

The PR's own argument for overlays is that the shared files live in one place: only 3 of the 10 TypeScript template files differ per package manager, so the file cost of a variant is small. That's true, and it's a good reason to keep the overlay. 

But every variant also adds surface in core code that has to know about the toolchain:

- detection in `NodePackageManager.packageManager` (`packages/@cdktn/cli-core/src/lib/dependencies/package-manager.ts:281`)
- the install/add `argv` and its per-manager quiet flags (:324)
- `listProviderPackages` plus a listing implementation per manager (:356, :410`)
- `isTypescriptTemplate` and `packageManagerFor` in `packages/cdktn-cli/src/bin/cmds/helper/init.ts`
- `distKeysByTemplate` and `TEMPLATE_OVERLAY_BASE` in `packages/@cdktn/cli-core/src/lib/init.ts`
- the `.gitignore` carve-outs, the `help` text, and the `.hooks.sscaff.js` install list

That is six or so places per variant, and this PR is the demonstration: the yarn variant ships `provider list`, `provider upgrade` and `debug` broken because the listing path assumed Yarn 1 while the template pins Yarn 4. The overlay made the files cheap but the affected code area is much larger.

## Should we consider customizations in test runners next?

Package manager is the affordable axis — `help`, `package.json`, `.hooks.sscaff.js` — and the differences are mostly textual.

Test runner is not. jest drags in `jest.config.js` (6.6 KB), `setup.js`, `__tests__/main-test.ts` with jest globals, the `tsconfig.json` types, and the `.gitignore` carve-outs `!jest.config.js` / `!setup.js`. 

Similarly to how people are moving from ts-node to tsx (as we just did), a lot of people prefer vitest (a lot less TS hurdles compared to Jest).

There's also `bun test` and `node:test` each change roughly half of the base template, plus the `test` and `test:watch` scripts and the help text. "7 of 10 files shared" becomes "5 of 10", against the same six core touchpoints. 

## Alternative approaches

1. **Make remote templates first-class.** `--template=<url>` is accepted, but the toolchain is inferred from the template *name*: `isTypescriptTemplate` and `packageManagerFor` in `helper/init.ts` hardcode npm/pnpm/yarn and fall back to `npm run get`. A remote bun template would scaffold correctly and then run `npm run get`. Deriving the install/run command from the scaffolded project — which this PR already gives us, via the `packageManager` field and lockfile detection — fixes remote templates and removes the duplication with `TEMPLATE_OVERLAY_BASE` in the same change.
2. **Remove init, replace with ..** Nx Plugin / TurboRepo (plop) generator / projen type (already exists) and direct people to each of these instead.
3. **Keep init, but rewrite init to leverage projen.** If we really want to make `init` more flexible, leverage Projen and every package manager / runtime / test runner / ... it already supports
4. **If we want the matrix, use a toolchain descriptor.** One table of install/add/list/run commands, quiet flags and the dev flag, consumed by both the sscaff hook and `PackageManager`, instead of one directory per variant plus N name-keyed branches. That is the real difference from projen: projen has the component model, so its growth stays configurable. Here it would be structural, and each new variant would add another place to forget — which is exactly what happened to `yarn list` in this PR.

## Price of admission for variant nr.3 onward

An end-to-end test that scaffolds the variant and asserts the result. `packages/cdktn-cli/src/test/cmds/init.test.ts` is gated behind `describeIfDistExists` (it needs `pnpm package` plus Maven), so today no template is scaffolded in a normal test run. That is why the yarn listing failure had to be found by hand, and it is the cheapest thing to fix given the failure mode.

## Where I land

deprecate init system

@jsteinich I'd value your read on whether the variant count is a decision to make now, or after remote templates are first-class.

X-Guardian:
I'll leave this PR here for you to do with what you want then.

My personal view is that currently `cdktn init` is broken for the package manager it uses itself, `pnpm` which is a bit embarrassing. This is just two variants on the typescript template, following the python/python-pip pattern already in place. You are talking as if there have been lots of templates added, which is not correct.

so0k:
> I'll leave this PR here for you to do with what you want then.
> 
> My personal view is that currently `cdktn init` is broken for the package manager it uses itself, `pnpm` which is a bit embarrassing. This is just two variants on the typescript template, following the python/python-pip pattern already in place. You are talking as if there have been lots of templates added, which is not correct.

sure, I'd remove yarn and keep the pnpm fixes. I am reluctant to touch `cdktn init` is all - I highlighted this in past PRs that work on the init templates

EDIT:
- https://github.com/open-constructs/cdk-terrain/issues/79
- https://github.com/open-constructs/cdk-terrain/pull/74

but I'll wait for Jon's opinion if we keep Yarn (and fix the commands) or just stick to pnpm (as pointed out we use pnpm for this project).

jsteinich:
* I don't think we should add a yarn template - it doesn't seem popular enough to warrant its inclusion
* I can see the argument for adding pnpm as we do use it with the project
* I'd really like to see the addition of new templates being driven by actual community requests for it. @X-Guardian perhaps this does serve as your request, but it's not clear that you'd actually be using the template from your comments.
* In the past I would use init for reproducing reported bugs; however, these days AI can set up a full reproduction projection much easier. For real projects, I just copy an existing one (yes, I know projen was designed to handle that flow). I'm not sure what other's experiences are, but it seems like we should find out before decided to invest time into init expansions / improvements.
* If we do add a new template, we should add integration tests to ensure that it actually functions correctly.

PR #425, opened 2026-09-11
fix(tests): bound and retry the dist publish to the local registry

X-Guardian:
### Description

`run-against-dist.mjs` publishes every tarball in `dist/js` to a throwaway Verdaccio registry before the integration tests run. That `npm publish` had no timeout and no retry, so a registry that stops responding mid-upload leaves npm waiting on the socket indefinitely — the script never returns and the job burns its full 60-minute budget before GitHub cancels it. Each publish is now bounded at two minutes and retried up to three times.

**What this looked like in CI.** On [run 34601028750](https://github.com/open-constructs/cdk-terrain/actions/runs/34601028750/job/103270765900), `linux_provider (datadog)` ran for 1h0m46s and reported `cancelled` rather than `failure` — the conclusion GitHub records when a job hits `timeout-minutes`. The step had in fact stalled three seconds in, while publishing `cdktn@0.0.0.jsii.tgz`:

```
13:00:40.4696312  /__w/.../dist/js/cdktn@0.0.0.jsii.tgz
13:00:40.5213662  npm warn using --force Recommended protections disabled.
14:00:33.3804396  context canceled
```

The provider tests never started, so the matrix leg in the job name is incidental — the hang is upstream of them, in the shared publish loop. `linux_provider (kubernetes)` in the same run shows the healthy path: tarball contents follow that same `npm warn` line about 0.6s later. The `cdktn**` block in `test/verdaccio.yaml` sets no `proxy`, so this was not an npmjs uplink stall; the likely cause is the in-process Verdaccio server ceasing to service the request, and `cdktn@0.0.0.jsii.tgz` is the largest tarball in the set at ~870kB of `.jsii` alone.

**Why a timeout plus a retry.** The CLI *install* immediately below this loop already rides out exactly this class of transient Verdaccio failure with a three-attempt retry; the *publish* above it had neither guard. The two now match in shape, with a shorter backoff (5s/10s) because a publish that has already hit its timeout has waited two minutes. A healthy publish to a local registry takes about a second, so the two-minute bound separates a stall from slowness with a wide margin and will not fire on an ordinarily slow runner.

`--force` is already passed on every attempt, so a retry overwrites whatever a killed attempt left half-written in Verdaccio's storage — the republish is idempotent and needs no cleanup between tries. The failure message distinguishes a timeout from an ordinary publish error, so a recurrence is diagnosable from the log alone rather than presenting as a bare `cancelled`.

Worst case is now roughly six minutes ending in a clear error, instead of an hour of a held runner ending in an opaque `cancelled`. That matters more than the wall-clock: the org's runner concurrency cap means one wedged job blocks the queue for everything behind it.

### Checklist

### DIFF
--- a/test/run-against-dist.mjs
+++ b/test/run-against-dist.mjs
@@ -36,6 +36,11 @@ const REGISTRY_URL = "http://localhost:4873/";
 // inline as npm's per-registry config key.
 const NPM_AUTH_ARG = `--${REGISTRY_URL.replace(/^https?:/, "")}:_authToken=dummy`;
 
+// A healthy publish to the local registry takes about a second; anything past
+// this is a stall, not slowness.
+const PUBLISH_TIMEOUT_MS = 2 * 60 * 1000;
+const PUBLISH_ATTEMPTS = 3;
+
 process.env.CDKTF_DIST = cdktnDist;
 
 /** @type {import("node:http").Server | null} */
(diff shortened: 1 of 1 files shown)


so0k (approved):

sakul-learning:
### Two non-blocking notes on the publish retry

**1. A retry is not idempotent if the first attempt's publish already reached Verdaccio's storage.** The description says `--force` on every attempt makes a retry "overwrite whatever a killed attempt left half-written", so the republish is idempotent and needs no cleanup. That holds when the killed attempt never committed, but not when Verdaccio stored the tarball and the response was lost — which is exactly a case a client-side timeout cannot distinguish.

Reproduced against Verdaccio 6.7.1 (the version in `test/package.json`), publishing the same tarball twice with the exact flags this script passes:

```
npm publish --registry=http://localhost:4873/ --//localhost:4873/:_authToken=dummy --force cdktn-0.0.0.tgz

attempt 1: + cdktn@0.0.0                                                                                          exit 0
attempt 2: npm error code E409
           npm error 409 Conflict - PUT http://localhost:4873/cdktn - this package is already present            exit 1
```

The result is identical without `--force`: the flag only disables npm-side "recommended protections" and does not change the server's answer. Verdaccio rejects the republish before any storage write, in `checkPackageLocal` (`verdaccio/build/lib/storage-utils.js`), which raises `API_ERROR.PACKAGE_EXIST` ("this package is already present") whenever metadata for the package already exists. So if the timed-out attempt had in fact been committed, attempt 2 and attempt 3 both return 409 and the run ends with `npm publish failed for <pkg> after 3 attempts: ... E409` even though the tarball is published. That is still better than the hour-long hang this PR fixes, so it is not a regression — but if the aim is a retry that tolerates a partially-successful attempt, treating "already present" as success (e.g. `break` out of the loop on that error) would get there, whereas retrying into a permanent conflict cannot.

**2. The same hang class is still reachable just below.** The CLI install loop right after this one (lines 196-213, `pnpm add cdktn-cli --registry=http://localhost:4873/`) installs from the same in-process Verdaccio and its `execa` call sets no `timeout`; its three-attempt retry only engages once the call fails, which a stalled socket never does. Bounding it the same way would close the phase that can still hold the job until the 60-minute job timeout.

For what it's worth, the bound itself is sound and holds under the case I expected to break it: because the third failed attempt calls `fail()` and exits, a registry that stalls on the very first tarball aborts after ~6.25 minutes, so the loop never walks the whole tarball set and "roughly six minutes" is accurate. `err.timedOut` is set by execa v5's own timeout path (`lib/kill.js`), the final attempt's "retrying" log and sleep are correctly unreachable, and `prettier --check test/run-against-dist.mjs` passes (it is the only CI gate over this file — eslint ignores `**/*.mjs` and jest's `testMatch` is `**/*test.ts`).

PR #431, opened 2026-09-12
chore(tests): run the function-binding scripts with tsx

jsteinich:
### Related issue

Loose end from #406. Noticed while auditing the remaining `ts-node` references after the TypeScript 6 work.

### Description

`tools/generate-function-bindings` still ran its four codegen scripts through ts-node after #406 moved the package to `typescript ~6.0.0`:

```
fetch-metadata                  ts-node scripts/fetch-metadata.ts
generate                        ts-node scripts/generate.ts
update-function-matrix          ts-node scripts/update-function-matrix.ts
generate-function-availability  ts-node scripts/generate-function-availability.ts
```

That is the pairing that broke `cdktn synth` for the init template in #396: ts-node merges a bundled `@tsconfig/nodeXX` default supplying a deprecated `moduleResolution: node10`, which TypeScript 6 rejects with `TS5107`. Nothing in CI runs these scripts — they are invoked manually via the root `generate-function-bindings` script — so the risk was silent and would have surfaced the next time someone regenerated the bindings.

### On certainty

Whether ts-node actually fails here is environment-sensitive, and I could not settle it:

| environment | result |
| --- | --- |
| CI container (`jsii-terraform`) | reproduced reliably — `tsNodeModuleResolution: Node10` → `TS5107` |
| Windows | never reproduced; ts-node resolves no `moduleResolution` at all |
| plain `node:22` container | not reproduced — **and the template's own known-failing tsconfig passes there too**, so this check cannot clear anything |

Rather than keep chasing why the injection is environment-dependent, this moves to tsx: the same runner the init template, the integration fixtures and the examples already use. That removes the exposure regardless.

### Verification

`tsc --noEmit` is clean under TypeScript 6, and three of the four scripts were executed through tsx locally:

- `generate-function-availability` — ran, wrote `function-availability.generated.ts`
- `generate` — ran clean
- `update-function-matrix` — reached its own `Unsupported platform win32/x64` guard, a pre-existing Windows limitation in the script rather than a runner problem, so tsx had loaded and executed it

Critically, `git status` showed **no changes to any tracked file** afterwards: the generated output is byte-identical to what is committed, so tsx produces the same result ts-node did.

`fetch-metadata` was not run — it performs network fetches — but tsx's ability to load this package's module graph is established by the other three.

### Checklist

### Not included

`packages/@cdktn/cli-core/src/lib/synth-stack.ts:118` still has a comment reading "increase memory to allow ts-node (when using TypeScript)", which is now stale since the template uses tsx. Left out to keep this scoped; happy to fold in or do separately.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1556,9 +1556,9 @@ importers:
       '@types/node':
         specifier: 22.20.1
         version: 22.20.1
-      ts-node:
-        specifier: ^10.9.1
-        version: 10.9.2(@swc/core@1.16.2(@swc/helpers@0.5.23))(@types/node@22.20.1)(typescript@6.0.3)
+      tsx:
+        specifier: 4.23.13
+        version: 4.23.13
       typescript:
         specifier: ~6.0.0
         version: 6.0.3
@@ -9060,6 +9060,7 @@ snapshots:
   '@cspotcode/source-map-support@0.8.1':
     dependencies:
       '@jridgewell/trace-mapping': 0.3.9
+    optional: true
 
   '@cypress/request@3.0.10':
     dependencies:
@@ -9822,6 +9823,7 @@ snapshots:
     dependencies:
       '@jridgewell/resolve-uri': 3.1.2
       '@jridgewell/sourcemap-codec': 1.5.0
+    optional: true
 
   '@jsii/check-node@1.140.0':
     dependencies:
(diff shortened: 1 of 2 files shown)


eduardomourar (commented):
May I suggest just ditching any extra dependency altogether. With Node.js support for TypeScript natively (more details [here](https://nodejs.org/learn/typescript/run-natively)), you can just run `node script.ts` after v22.18 (and with `--experimental-strip-types` before).

so0k:
> May I suggest just ditching any extra dependency altogether. With Node.js support for TypeScript natively (more details [here](https://nodejs.org/learn/typescript/run-natively)), you can just run `node script.ts` after v22.18 (and with `--experimental-strip-types` before).

I think this can be a follow up - ts-node replacement with tsx has been spread across 3 PRs so far, so need to find all the places and update if we go with that

so0k (approved):

sakul-learning:
Local verification on the pinned Node v22.22.3 suggests `node script.ts` is not a drop-in replacement for the `tools/generate-function-bindings` scripts: their extensionless `./constants` / `./matrix` imports fail with `ERR_MODULE_NOT_FOUND`, and importing `matrix.ts` explicitly reparses it as ESM and then fails because it uses `__dirname`. Would native Node therefore require a coordinated import-extension/module migration across all affected scripts rather than just changing these package commands? Deferring that broader cleanup to a follow-up seems safer.

PR #433, opened 2026-09-13
feat(lib)!: add asset staging core with no bundling

eduardomourar:
### Related issue

Spin-off from https://github.com/open-constructs/cdk-terrain/pull/339

### Description

Second slice of the asset pipeline (foundation types landed in #371). Gives `AssetHashType.OUTPUT`, `exclude`, and `extraHash` real behavior in `TerraformAsset`, backed by a new `AssetStaging` construct — no Docker/bundling code, which stays scoped to #339.

- `AssetStaging` — resolves `SOURCE`/`CUSTOM`/`OUTPUT` hashing (`OUTPUT` no longer throws: without a bundler, an asset's "output" is its source verbatim) with `exclude`/`ignoreStrategy`/`extraHash`, and a `stage()` method that packs content honoring the same exclusions used for hashing. Hashing happens eagerly in the constructor; the filesystem side effect is deferred to `stage()`, called from the owning construct's `onSynthesize`.
- `TerraformAsset` gains `exclude`/`extraHash` config options. Routes through `AssetStaging` only when either is set, so existing simple usage is unchanged — no new construct child, no behavior change. Also fixes a latent gap: packing now receives the same ignore strategy used for hashing, so excluded files are actually absent from the staged/zipped output, not just the hash.
- `AssetStaging` never changes packaging based on `exclude`/`extraHash` — a directory asset stays a directory unless the caller explicitly asks for `ARCHIVE`. This is a deliberate departure from the original #339 spike, which silently zipped directories once any advanced option was set.
- Adds a per-`App` content-keyed cache for the `SOURCE`/`OUTPUT` hash walk (same asset referenced from multiple resources/stacks hashes once), and `ASSET_HASH_SALT_CONTEXT_KEY` for app-wide cache-busting alongside the existing per-asset `extraHash`. Both ported from #339, minus its Docker-bundling coupling.

Held back for #339: `IAssetBundler`, Docker bundling, and the `ILocalBundling`/bind-mount/volume-copy execution.

Not in scope here: the publisher layer (`IAssetPublisher`/`AssetReference`) and fixing `TerraformAsset`'s staging directory to live inside the stack directory (#380's concern #1) — both pre-existing, separate concerns.

### Checklist

BREAKING CHANGE: IAssetPackaging gains a required member omitsDirectoryEntries; external implementations must add it. Custom TerraformAsset assetHash values are now validated against `/^[A-Za-z0-9_.-]+$/` and throw if they contain other characters (e.g. "feature/my-branch").


jsteinich (commented):
`AssetStaging` itself looks good — hashing eager, `stage(targetPath)` taking the path from the caller, and the cache scoped per construct-tree root rather than module-global. Two notes on how `TerraformAsset` consumes it.

packages/cdktn/src/terraform-asset.ts
```diff
@@ -111,7 +133,20 @@ export class TerraformAsset extends Construct implements IAsset {
     const stat = fs.statSync(this.sourcePath);
     const inferredType = stat.isFile() ? AssetType.FILE : AssetType.DIRECTORY;
     this.type = config.type ?? inferredType;
-    this.assetHash = this.resolveAssetHash(id, config);
+
+    if (config.exclude?.length || config.extraHash) {
```
jsteinich: Routing through `AssetStaging` only when `exclude`/`extraHash` are set leaves two hashing implementations that have to be kept in agreement. They agree today — `type === ARCHIVE` maps onto `packaging === AssetPackaging.ZIP`, both read the same context key, and a no-op ignore strategy is hash-neutral — but that is three invariants held by convention.

One has already diverged. `AssetStaging` validates a custom hash, this path returns it verbatim:

```ts
new TerraformAsset(s, "a", { path, assetHash: "../../escape" })                  // accepted
new TerraformAsset(s, "a", { path, assetHash: "../../escape", extraHash: "x" })  // throws
```

`assetHash` is used as a path segment in `TerraformAsset.path`, so the accepted form escapes the assets directory at synth. The permissive behaviour predates this PR, but the asymmetry is new — and `"a resolved assetHash with unsafe characters throws"` sits under the exclude/extraHash describe, so only one path is covered.

Suggest dropping the branch and always routing through `AssetStaging`. With no exclude/extraHash it produces the same hash and the same packing, so the fallback buys nothing, and agreement becomes structural instead of conventional. It closes the traversal case too, which is a behaviour change worth calling out.

If the branch stays, a test pinning the two paths to the same hash (e.g. `exclude: ["does-not-exist.txt"]` vs no exclude) would lock the invariant down.

packages/cdktn/src/asset-staging.ts
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    switch (assetHashType) {
+      case AssetHashType.CUSTOM:
+        throw assetHashTypeCustomRequiresHash(id);
+      case AssetHashType.SOURCE:
+      case AssetHashType.OUTPUT:
+      case undefined: {
+        const canonical = !!this.node.tryGetContext(CANONICAL_ASSET_HASHES);
+        const archive = this.packaging === AssetPackaging.ZIP;
```
jsteinich: An identity check against the built-in singleton means any custom `IAssetPackaging` always gets `archive: false`.

The predicate that matters is whether the emitted artifact preserves directory entries — a property of the packaging, not of which instance it is. That belongs on `IAssetPackaging`. `false` is the safe default (spurious diffs rather than missed ones), but a custom zip-equivalent silently gets different framing than the built-in.

Worth adding the member now if it is wanted: `IAssetPackaging` only shipped on the `next` channel days ago, and adding to a behavioural interface is breaking for implementers once it is out.

packages/cdktn/src/terraform-asset.ts
eduardomourar: I have dropped the branch and now is routing through `AssetStaging`.

packages/cdktn/src/asset-staging.ts
eduardomourar: Done

jsteinich (commented):
Both points addressed — and the custom-packaging test pins exactly the right thing.

Two resulting behaviour changes need the breaking marker, same as #371 got:

1. `IAssetPackaging.omitsDirectoryEntries` is a new **required** member on an interface that has been published since #371 (`v0.25.0-pre.21` onward). jsii behavioural interfaces require implementers to supply every member, so an external implementation stops compiling.

2. Custom `assetHash` values are now validated for every `TerraformAsset`. They were returned verbatim before; now `SAFE_ASSET_HASH` rejects anything outside `[A-Za-z0-9_.-]`, so `assetHash: "feature/my-branch"` throws where it used to work. Right change — it just needs to reach the changelog.

No version cost to marking it: the manifest is at 0.24.0 with `bump-minor-pre-major`, and #371 already landed as `feat(lib)!:`, so this cycle targets 0.25.0 either way. Release-please applies the highest bump once rather than compounding per commit.

jsteinich (approved):

PR #434, opened 2026-09-14
test(lib): ensure TerraformAsset stage within stack own directory

eduardomourar:
### Related issue

 Follow-up from #380.

### Description

#380 claimed asset staging lands as a sibling of `stacks/`, breaking once uploaded to a hosted runner. Traced it: `TerraformAsset` already stages inside `stacks/<stack-id>/`, and has since before #371. The bug only ever existed in the old, never-merged spike's `AssetStaging`, which #433 didn't carry forward.

No code fix needed. Adds a regression test pinning this contract so it can't regress silently later (e.g. in #339).

### Checklist


packages/cdktn/test/canonical-asset-hash.test.ts
```diff
@@ -514,6 +514,41 @@ describe("TerraformAsset artifact layout derives from the packaging", () => {
   });
 });
 
+describe("TerraformAsset stages inside the stack's own directory (#380)", () => {
```
jsteinich, line 517: This guards `TerraformAsset`, which was not the thing at risk. The #380 concern was `AssetStaging.determineAssetOutdir()` resolving to `<outdir>/assets`, a sibling of `stacks/`.

In #433 that method does not exist — `stage(targetPath)` takes the path from the caller — so the risk is gone by construction rather than by assertion. Still a reasonable guard on the invariant, but the title points at a class that never had the bug.

jsteinich (commented):
Cheap and worth having. One framing note below, plus: this touches `canonical-asset-hash.test.ts`, as does #433, so the two will likely conflict — worth picking a merge order.

packages/cdktn/test/canonical-asset-hash.test.ts
eduardomourar: Changed the PR title to be more clear

eduardomourar:
I would like to merge this PR first as it is smaller.

jsteinich (approved):

PR #436, opened 2026-09-14
test(tests): move the testing-matchers fixtures to Docker provider 3.x

jsteinich:
### Related issue

Follow-up to #429 / #301, which moved the `cdktn` unit matcher fixture off Docker provider 2.x. The five language integration matchers were left out of that PR; this does them.

### Description

`test/{typescript,go,java,csharp,python}/testing-matchers/cdktf.json` pinned `kreuzwerker/docker@~> 2.0`, which resolves to 2.25.0 and fails under OpenTofu:

```
Error while installing kreuzwerker/docker v2.25.0: the provider is not signed with a
valid signing key; please contact the provider author (authentication signature from
unknown issuer)
```

`~> 3.0` floats to the latest 3.x, which is signed, and v3 is what the rest of the suite already uses — `@cdktn/provider-schema` pins `=3.0.2`, `@cdktn/hcl2cdk` pins `=3.0.1`, and #429 moved the unit fixture to `~> 3.0`.

Only Go needed a code change: `docker_image.latest` was removed in v3, and Go was the only fixture using it. `dockerImage.Latest()` becomes `dockerImage.Name()`, which is what the Python, Java and C# fixtures already pass for the container image. TypeScript passes a literal string and is unchanged. (The `"nginx:latest"` strings elsewhere are image tags, not that attribute.)

Confirmed against the real v3 schema rather than release notes:

```
docker_image attributes (v3): force_remove, id, image_id, keep_locally, name,
                              platform, pull_triggers, repo_digest, triggers
has latest? false
```

There are no snapshots under `test/*/testing-matchers/`, and the harness asserts only that the language-side run reports `PASS`, so there is no expected-output churn.

### Testing

Not run locally — these need `pnpm package` plus the go, dotnet, java and python toolchains and a live Docker daemon. CI is the verification.

Verified locally: the v3 schema above, that Go was the only fixture using the removed attribute, and that no `docker@~> 2` pin remains under `test/*/testing-matchers/`.

### Out of scope

`test/python/providers/cdktf.json` pins `kreuzwerker/docker@= 2.16.0` and has the same problem — separate PR.

`@cdktn/hcl2cdk`'s `~>2.15.0` and inline `2.14.0` HCL in `convertProject.test.ts` stay as they are: those are v2-era HCL **inputs**, and converting legacy HCL is the point of that test.

### Correction: the signing boundary is mid-3.x

While doing the same bump for `test/python/providers` (#440) it turned out that "v3 is the first signed major" — which #429 also states — is wrong. Against OpenTofu 1.12.6: 3.0.1, 3.0.2, 3.2.0, 3.3.0, 3.5.0 and 3.6.0 are all rejected as unsigned; 3.7.0, 3.8.0 and 3.9.0 install signed with key `0DCE698927DAF8EC`.

This PR is still correct, but for a narrower reason than stated: `~> 3.0` resolves to the newest 3.x, which is signed. A pin below 3.7.0 would not have worked.


sakul-learning:
Reviewed at `00d16fcb` against `origin/main`. **Approving** — the migration is correct and there is nothing here I would change before merge.

Scope check: all five fixtures move `kreuzwerker/docker` from `~> 2.0` to `~> 3.0`, and the two Go call sites in `test/go/testing-matchers/main.go` replace the removed `dockerImage.Latest()` with `dockerImage.Name()`. That matches the Python, Java and C# fixtures, which already pass the image's `name` into the container's `image` argument. No workflow, credential, or dependency-source changes.

I checked that the five targets really do exercise these fixtures in CI rather than only generating bindings, since that determines whether the Go rename is actually validated. It is: the Go job runs the full `main_test.go`, all twelve tests pass, and `TestToBeValidTerraformPass` runs `terraform init` and `validate` against a stack built from the v3-generated `docker` bindings — so `Name()` is confirmed to exist and type-check against the regenerated provider, not merely to compile. The other four report their expected assertion counts (TypeScript 2/2, Python `11 passed`, Java `Task :test` + `BUILD SUCCESSFUL`, C# resolving without throwing). All ten `linux_integration` jobs covering these five languages pass, at Terraform 1.5.7 and 1.16.1.

The PR body's correction also holds up: `~> 3.0` is correct because it floats to the newest 3.x, not because the major itself is signed. `test/python/providers/cdktf.json` on current `main` now records Docker 3.9.0, and `~> 3.0` permits it.

Two notes below, neither blocking.

## 1. `cdktn:test` reports as a flaky task, and the runner-side cause looks like memory pressure

Running the suite with `--runInBand` takes **337s** for 45 suites / 581 tests / 299 snapshots. Nx has now self-reported the task as flaky:

```
 NX   Nx detected a flaky task

  cdktn:test

Flaky tasks can disrupt your CI pipeline. Automatically retry them with Nx Cloud.
```

Nx records a task as flaky when successive runs disagree, and the earlier disagreement was real: two suites failed with `Cannot find module '../lib'` in `test/asset-hash.test.ts` and `test/assets-types.test.ts`. That run began immediately after Nx reported `Successfully ran target build for project cdktn` / `Nx read the output from the cache instead of running the command` — the build was served from cache rather than executed, and the test process then could not resolve the package's own compiled entry point. An identical rerun passed 45/45. In Nx's output this surfaces as a warning attached to an otherwise green target:

```
Warning: command "jest --runInBand" exited with non-zero status code

 NX   Successfully ran target test for project cdktn and 1 task it depends on
```

Nothing in this PR touches `packages/cdktn`, so **this is not caused by these six fixture edits.** I am flagging it because a task that is both 337s and sensitive to available memory will eventually land red on an unlucky runner, and because the green-with-warning line makes the previous disagreement easy to miss.

Two caveats on how much this implies for CI, since I only have one environment to read from: my runs were on 4 vCPU / 8GB with swap already partly in use, and I invoked the suite alongside two other Jest/nx processes, so my conditions were more contended than a single CI job. I also cannot rule out that Nx's flaky counter here is only reflecting my own reruns rather than anything that has happened on a CI runner.

The concrete things I would want pinned down, in order:

1. Whether `cdktn:test` is memory-bound. 45 suites in one worker with `--runInBand` means one process holds the whole suite's peak footprint.
2. Whether the build outputs are reliably present when the test target starts, given that a cache-served build was followed by a failure to resolve `../lib`.

If a CI runner is already regularly retried for this, then it is known noise and this note is just corroboration. The other possibility worth checking is whether flaky-task detection is enabled for this repo at all — the message above is Nx suggesting it, which implies it is not yet configured.

## 2. The fix is not exercised under the runtime it exists to fix

The five fixtures this PR changes are integration targets. `tools/build-test-matrix.mjs` resolves every target without a `pinnedRuntimes` entry to `.terraform.versions.json`'s `tested` list — currently Terraform `1.5.7` and `1.16.1` — with `product: "terraform"`. Only `typescript/provider-features/test.ts` is pinned to OpenTofu `1.12.6` (and Terraform `1.16.1`). So these five fixtures are verified against the Terraform CLI in CI and never against `tofu`, even though OpenTofu rejecting the old provider signature is precisely what they were changed to fix.

I initially raised this as an open gap. Having read the surrounding context, it is more accurately **known and pending**, and what it means depends on which axis you are talking about:

- The **unit** axis closes first. #429 fixed `packages/cdktn/src/testing/__tests__/matchers.test.ts` and closed #301. #439 adds one OpenTofu `1.12.6` entry to the `affected_terraform` matrix, scoped by the `unit-test:terraform` tag to `cdktn`, `@cdktn/cli-core`, `@cdktn/provider-generator` and `@cdktn/provider-schema`. None of the `test/*/testing-matchers` targets live in those packages.
- The **integration** axis is where these fixtures live, and #439 explicitly leaves it out, pricing a full OpenTofu leg on the integration matrix at +77 jobs and roughly +115 runner-minutes. #337 is the open issue covering that axis.

So the unit fixtures and these integration fixtures need separate Docker pins and fail for the same signing reason, but only one of the two halves gets a `tofu` leg from #439. I am not proposing that this PR carry that work — the cost is real and it is being handled deliberately elsewhere. The decision I would want recorded, against #337 rather than here, is whether the integration matrix eventually carries an OpenTofu entry covering these five targets, or whether OpenTofu coverage stays on the unit axis alone and these fixtures are accepted as Terraform-verified only. Either answer is defensible; they give different confidence that the pins stay valid under `tofu`.

Related and worth folding into the same follow-up: `@cdktn/provider-schema` pins `=3.0.2` and `@cdktn/hcl2cdk` pins `=3.0.1`. Per #440 both are below the 3.7.0 signing boundary. #440 notes neither currently breaks the #439 unit leg, so this is latent rather than active — but they are the same class of pin as the ones being fixed here.

so0k (approved):

so0k:
i think nothing really relevant in the bot review comment - just its initial findings were lacking context and caused by its own runtime restrictions

feel free to ignore it

PR #438, opened 2026-09-14
chore(gha): select the CLI by binary name instead of concatenating a version

jsteinich:
### Related issue

Prerequisite for adding an OpenTofu axis to any workflow other than `integration.yml`. Follows #405, which introduced the `binary` matrix field.

### Description

`integration.yml` sets `TERRAFORM_BINARY_NAME` from a `binary` matrix field. Every other workflow still builds the name by concatenating:

```yaml
TERRAFORM_BINARY_NAME: "terraform${{ matrix.terraform }}"
```

That makes a non-Terraform runtime impossible to express — there is no version string that turns `"terraform${...}"` into `tofu1.12.6`. This removes the concatenation everywhere so the runtime is nameable.

**No behaviour change.** Every binary resolves to exactly what it did before; this is plumbing only.

| workflow | change |
| --- | --- |
| `unit.yml` | takes `terraform_binary` **in place of** `terraform_version` — the version was never used as a version there, only to build the binary name and key the concurrency group and schema cache |
| `provider-integration.yml` | keeps `terraform_version` (also exported as `TERRAFORM_VERSION`, and drives `setup-terraform` on the windows path) and gains `terraform_binary` beside it |
| `pr-unit.yml` | matrix carries the binary |
| `release.yml` | both `unit.yml` fan-outs pass a binary |
| `examples.yml` | takes it from the matrix |
| `tools/build-example-matrix.mjs` | emits `binary` alongside `terraform` |

Provider and schema caches are now keyed by binary rather than version, matching what `integration.yml` already does — Terraform and OpenTofu resolve providers from different registries into the same `TF_PLUGIN_CACHE_DIR`, so they must not share a cache entry.

### Testing

`build-example-matrix.mjs` output:

```
terraform: ["1.16.1"]  binary: ["terraform1.16.1"]
targets: 58  include: 5
sample include: {"target":"@examples/csharp-aws","terraform":"1.16.1",
                 "binary":"terraform1.16.1","hclOutput":true}
```

`prettier --check` clean across the workflows and the matrix builder. No `"terraform${{ … }}"` concatenation remains in `.github/workflows/`.

The real verification is that this PR's own CI run is unchanged from `main` — same jobs, same binaries.

### Follow-up

This unblocks a scoped OpenTofu entry on the unit matrix, which is where #301 actually lived: the matcher fixture is a unit test, and nothing in CI exercises it under `tofu` today, so that fix can silently regress.


so0k (approved):

sakul-learning:
Re-checked the cache-key side of this change at `ae48762`.

First, confirmation on the matrix fix: `.github/workflows/release.yml:246` and `.github/workflows/release.yml:273` now declare `terraform_binary: [terraform1.16.1, terraform1.5.7]` as a real axis, so both release channels go back to 8 packages × 2 binaries. The `include`-only form that silently dropped `terraform1.16.1` is gone.

**One item left behind: `pr-unit.yml` still carries `terraform_version`, and its only remaining reader is the provider-schema cache key.**

`.github/workflows/pr-unit.yml:104-108` declares both fields on every matrix entry:

```yaml
matrix:
  include:
    - terraform_version: "1.16.1"
      terraform_binary: terraform1.16.1
    - terraform_version: "1.5.7"
      terraform_binary: terraform1.5.7
```

In that file `terraform_version` is not exported as `TERRAFORM_VERSION`, is not part of the concurrency group, and does not select the CLI — `TERRAFORM_BINARY_NAME` comes from `matrix.terraform_binary` at `.github/workflows/pr-unit.yml:177`. The only remaining reference anywhere is the actions/cache key at `.github/workflows/pr-unit.yml:147`:

```yaml
key: providerSchema-${{ runner.os }}-${{ matrix.terraform_version }}-unit-affected
```

Every other provider/schema/plugin cache in this change partitions on the binary name instead: `.github/workflows/unit.yml:62` (`inputs.terraform_binary`), `.github/workflows/provider-integration.yml:135` and `:188`, and `.github/workflows/examples.yml:126`. `pr-unit.yml` is the one site still keyed by version.

**Impact.** Two entries that name different runtimes but share a version string would restore and save the same `providerSchema` actions/cache entry. The blast radius is narrower than for the plugin cache: the directory's own contents are already segmented per CLI product and minor version — `packages/@cdktn/provider-schema/src/read.ts:35-52` derives a suffix such as `tofu-1.12`, applied in `packages/@cdktn/provider-schema/src/cache.ts:17-22` — so a shared entry cannot serve the wrong runtime's schema. What is lost is the cache entry itself: both jobs restore the same key, and the first save wins, so the other job's newly fetched schemas are not persisted. Nothing changes today, because `.terraform.versions.json` currently lists Terraform 1.5.5/1.5.7/1.6.5/1.15.8/1.16.1 and OpenTofu 1.12.6 only — no two runtimes share a version string.

**Suggested change**, if runtime-scoped cache partitioning is the intent here as everywhere else:

```yaml
key: providerSchema-${{ runner.os }}-${{ matrix.terraform_binary }}-unit-affected
```

The binary name already embeds the version, so no granularity is lost, and `terraform_version` would then have no reader left in the file, so the two `terraform_version:` lines can be dropped with it.

**Is the version key deliberate?** I could not find a functional reason for keeping it. The key dates from 78652380c ("chore: migrate jest to @nx/jest plugin with swc-jest (#160)"), when `terraform_version` was the matrix axis and also fed the concatenated binary name. This change moved the binary name to `terraform_binary` and left the cache key reading the version. If retaining the version key is intentional — for example to keep in-flight PR caches valid under the old key — say so and I'll drop this.

Not a blocker: this head's CI is green (259 checks passed, one intentionally skipped Windows integration job), and the entries in this matrix remain Terraform-only.

<details>
<summary>Verification</summary>

- zizmor baseline, merge-base `fec67bf379ae` vs `ae48762ef99c`: 5 changed workflows, 119 baseline findings, 52 head findings in those workflows, 0 new or touched findings. Pass.
- `pnpm exec jest packages/cdktn/test/validations.test.ts --runInBand`: 1/1 suites, 41/41 tests, 11/11 snapshots.
- `pnpm exec nx build cdktn`: 0 errors, 0 warnings.

</details>

sakul-learning:
**The cache-key change suggested above is already written — it is the stacked PR #439.**

`14a828cad` in #439 is the only commit in that branch which is not already in this one (its parent is this PR's head, `ae48762ef`). It touches `.github/workflows/pr-unit.yml` alone, +3/−1, and makes two changes:

```diff
@@ -106,6 +106,8 @@ jobs:
             terraform_binary: terraform1.16.1
           - terraform_version: "1.5.7"
             terraform_binary: terraform1.5.7
+          - terraform_version: "1.12.6"
+            terraform_binary: tofu1.12.6
@@ -144,7 +146,7 @@ jobs:
           path: ${{ steps.global-cache-dir-path.outputs.providerSchema }}
-          key: providerSchema-${{ runner.os }}-${{ matrix.terraform_version }}-unit-affected
+          key: providerSchema-${{ runner.os }}-${{ matrix.terraform_binary }}-unit-affected
```

The two changes belong together. The OpenTofu matrix entry is the first case where the cache key's version and the CLI's binary name can disagree, so the provider-schema cache key moves to the binary name in the same commit that introduces that entry.

**So this PR does not need the extra line.** Merging it as it stands leaves `providerSchema-${{ runner.os }}-${{ matrix.terraform_version }}-unit-affected` on `main` until #439 lands, which is equivalent today: no runtime pair shares a version string in `.terraform.versions.json` (Terraform 1.5.5/1.5.7/1.6.5/1.15.8/1.16.1, OpenTofu 1.12.6). After #439, nothing in the repository reads `matrix.terraform_version` — the two `terraform_version:` lines in the `include:` entries become labels, which is why they stay: the matrix values are appended to the rendered check name regardless of the explicit `name:` field, so the jobs still display as `Affected Unit Tests (terraform) (1.16.1, terraform1.16.1)` and `(1.5.7, terraform1.5.7)`.

Two things to weigh if the suggestion is applied here instead:

- #439's commit would then only contribute the OpenTofu entry, and the cache-key hunk would already be present when it rebases onto `main`.
- Dropping the `terraform_version:` lines along with the key changes the rendered check names to `(terraform1.16.1)` / `(terraform1.5.7)`, which matters only if a required status check matches the full name. And a new commit on this branch needs a fresh approval.

Verification at `ae48762ef99c8`: `git grep -n 'matrix\.terraform_version'` across the repository returns exactly one hit, `.github/workflows/pr-unit.yml:147`; the only other provider-schema cache key in the workflows is `.github/workflows/unit.yml:62`, which is already keyed on `inputs.terraform_binary`.

PR #439, opened 2026-09-14
feat(gha): run the affected unit tests against OpenTofu

jsteinich:
### Related issue

Closes the CI gap left by #429 / #301. **Stacked on #438** — merge that first; this branch contains its commit.

### Description

#301 was a unit-level failure: the matcher fixture pinned a Docker provider version OpenTofu rejects as unsigned. #429 fixed it — but nothing in CI runs those tests under `tofu`, so it can regress silently and we would not know.

One matrix entry on `affected_terraform`:

```yaml
- terraform_version: "1.12.6"
  terraform_binary: tofu1.12.6
```

**That is already scoped, which is why it costs one job rather than a second copy of the unit suite.** The job runs:

```
nx affected -t test --exclude='*,!tag:unit-test:terraform'
```

so it only touches the four projects tagged terraform-dependent — `cdktn`, `@cdktn/cli-core`, `@cdktn/provider-generator`, `@cdktn/provider-schema`. Those are exactly the packages whose tests shell out to a CLI; `cdktn-cli`, `hcl2cdk` and `hcl2json` invoke none and are already excluded.

The provider schema cache moves from being keyed by version to by binary — tofu and terraform fetch schemas from different registries and must not share an entry.

### Cost

One extra job on `depot-ubuntu-24.04-16`, only when a terraform-tagged project is affected. For comparison, a full OpenTofu leg on the integration matrix would be +77 jobs and roughly +115 runner-minutes.

### Deliberately not extended to `release.yml`

Its two `unit.yml` fan-outs are a package × version cross-product with **no tag filter**, so an OpenTofu entry there would also run `hcl2cdk`, `hcl2json` and `cdktn-cli` under tofu — none of which touch a CLI. Scoping that properly is a larger change, and worth doing only once this entry has proven itself on PRs.

### Testing

CI on this PR is the verification, and it is the interesting part: this is the first time the `cdktn` unit tests — including the #429 matcher fixture — run against OpenTofu in CI. If #429 were wrong, this job is what would say so.


sakul-learning (changes requested):
`release.yml`'s `unit_test` and `unit_test_next` matrices replace the existing `terraform_version` axis with two sequential `matrix.include` entries for `terraform_binary`. Because `terraform_binary` is not an original matrix key, both include entries apply to every original `package` combination, and the second (`terraform1.5.7`) overwrites the first (`terraform1.16.1`). Each release matrix therefore produces only the eight package jobs using Terraform 1.5.7 instead of the intended 8 package × 2 binary jobs, silently dropping Terraform 1.16.1 coverage from both release channels.

Please make `terraform_binary` a regular matrix axis, for example:

```yaml
terraform_binary: [terraform1.16.1, terraform1.5.7]
```

Workflow security baseline: 5 changed workflows; 119 findings at the merge-base baseline; 52 head findings in the changed workflows; 0 new or touched findings. Zizmor result: pass.

so0k (approved):

PR #441, opened 2026-09-16
feat(provider-generator): disable Go runtime type checking in jsii-pacmak

jsteinich:
Passes `--no-runtime-type-checking` to `jsii-pacmak` when Go is the generation target, so the emitted Go omits the per-type `__checks.go` / `__no_checks.go` validator files.

Relates to #387 (partially — see *Scope* below).

## Why

Go bindings for large providers have outgrown the Go module proxy's 500 MiB (524,288,000 B) source-tree cap. Measured on the published `cdktn-provider-aws-go` at `aws/v25.5.0` (provider 6.64.0), using `git ls-tree -r -l` rather than a working-tree walk:

| | bytes | % of module |
|---|---:|---:|
| **module tree (`aws/`)** | **568,251,211** | **108.4% of cap** |
| `*OutputReference.go` (9,983) | 149,756,268 | 26.4% |
| other `.go` (19,642) | 141,593,245 | 24.9% |
| `__checks.go` (18,721) | 126,438,833 | 22.3% |
| `__no_checks.go` (18,721) | 58,506,543 | 10.3% |
| `*List.go` (6,331) | 45,697,473 | 8.0% |
| embedded jsii tarball | 45,717,554 | 8.0% |

jsii-pacmak emits that validator pair for every type because Go has no implementation inheritance, so each proxy type re-declares every inherited member. Together they are **184,945,376 B — 32.6% of the module**.

Dropping them puts the same bindings at **383,305,835 B, 73.1% of the cap**.

## Why this is not an API break

Everything removed is unexported: `validateXxxParameters` methods on unexported `jsiiProxy_*` receivers. No exported name, signature, or import path changes, in Go or any other language.

pacmak guards the validator *definitions* (`lib/targets/go/package.js` `emitValidators`) and their *call sites* (`runtime/property-access.js`, `runtime/method-call.js`, `runtime/class-constructor.js`) behind the same flag, so the emitted Go stays self-consistent and compiles.

## What it costs

Error quality at the call site. Today:

```go
func (j *jsiiProxy_S3BucketCorsRuleOutputReference) SetAllowedHeaders(val *[]*string) {
	if err := j.validateSetAllowedHeadersParameters(val); err != nil {
		panic(err)
	}
	_jsii_.Set(j, "allowedHeaders", val)
}
```

A nil gives `panic: parameter val is required, but nil was provided` at the exact call. Without the validators the nil crosses the jsii boundary and fails later — a synth-time error, or a silently omitted attribute.

That loss is lopsided. Classifying all 583,728 validator functions against the jsii assembly by whether the member is *declared on* the type or *inherited*:

| | bytes | funcs |
|---|---:|---:|
| inherited (`SetInternalValue`, `SetTerraformResource`, `Resolve`, `GetStringAttribute`, …) | 132,739,855 | 462,670 |
| declared on the type (schema-specific setters) | 43,075,489 | 121,058 |

**71.8% of the removed bytes guard members callers never invoke directly.** Only the 121,058 declared-member validators cover mistakes a user can realistically make.

Consumers who want the checks back have no opt-in after this change — previously they could build with `-tags no_runtime_type_checking` to opt *out*. Worth weighing; it is why this is a draft.

Keeping only the declared-member validators ("targeted A") was measured as an alternative: it saves 132,742,293 B (83.1% of cap) versus 184,945,376 B (73.1%), i.e. 71.8% of the benefit, and would require a change to jsii-pacmak's Go emitter — the inherited validators cannot simply be inherited, because `internal.Type__cdktnComplexObject` is a type alias to the *interface*, so unexported methods are not promoted.

## Scope

The flag is global to a pacmak run, so it is applied only when Go is the sole target. That always holds for `cdktn get`, where `isGoTarget` and its siblings derive from a single `targetLanguage`; the guard is defensive for direct `generateJsiiLanguage` callers.

**This covers locally generated bindings only.** Prebuilt providers package Go through their own `jsii-pacmak --target go` invocation, generated by `cdktn-io/cdktn-provider-project` into each provider repo's `.projen/tasks.json`. Closing #387 needs the same flag applied there; that is a separate PR.

This also does not change the growth trajectory. The module has grown 525,519,743 → 568,251,211 B between `aws/v24.6.0` (2026-06-11) and `aws/v25.5.0` (2026-09-10) — **0.469 MB/day, very close to linear**. At that rate the headroom this buys is roughly **10 months**.

## Testing

- `pnpm nx test @cdktn/provider-generator` — 3 new unit tests covering Go-only, non-Go, and mixed-target argument construction.
- `tsc --noEmit` and `eslint` clean on the changed files.
- Not yet validated end-to-end by generating and compiling Go bindings for a real provider; that is the main thing this draft still needs.


packages/@cdktn/provider-generator/src/get/constructs-maker.ts
```diff
@@ -95,6 +95,23 @@ export interface ExportDefinition {
...
+ * Go bindings are emitted without runtime type checking: pacmak's validators are
+ * unexported, so dropping them changes no public API, and they account for ~33% of
+ * the generated Go by size. See the PR description for measurements.
+ */
+export function pacmakArgs(opts: GenerateJSIIOptions): string[] {
+  const args = ["--code-only"];
+  // The flag is global to the pacmak run, so only apply it when Go is the sole target.
+  const goOnly = !!opts.golang && !opts.python && !opts.java && !opts.csharp;
+  if (goOnly) {
+    args.push("--no-runtime-type-checking");
```
so0k, line 110: should this also be configurable through CLI flags or ENV Variables (only set in large provider release flow?)

i know this is draft so maybe that's already the plan?

jsteinich: With some of the further investigation done, I'm considering not even merging this.
If we do decide to merge, your suggestion makes a lot of sense though.

PR #442, opened 2026-09-16
feat(lib)!: add asset bundling extension interface

eduardomourar:
### Related issue

Spin-off from #339, per the core split in #380.

### Description

Adds `IAssetBundler`, the extension point for asset bundling. No implementation ships; bundlers live in their own packages and pass via a `bundler` option.

- `IAssetBundler` — `bundle(options): string`, plus optional `bundlerKey` folded into the asset hash so a changed build changes identity.
- `BundleOptions` — `source` and caller-owned `outputDir`.
- `IAssetPackaging.acceptsDirectorySource`.
- `bundler?` on `AssetStaging` and `TerraformAsset`.

### Checklist

BREAKING CHANGE: `IAssetPackaging` gains `acceptsDirectorySource` as a required member. Custom implementations of the interface must add it.


packages/cdktn/src/asset-staging.ts
```diff
@@ -91,6 +101,23 @@ export interface AssetStagingOptions extends AssetOptions {
...
+   * With a bundler, the hash is taken over the bundler's inputs — the source
+   * tree, plus any `extraHash` describing the build — not the built output.
+   * `AssetHashType.OUTPUT` is therefore rejected when a bundler is set, since
+   * hashing the output would require building eagerly at construction time;
+   * fold build inputs into `extraHash`, or supply a `CUSTOM` `assetHash`,
+   * instead.
+   *
+   * @default - the source is staged verbatim, with no build step
+   */
+  readonly bundler?: IAssetBundler;
```
jsteinich, line 146: With a bundler set, identity is a function of the source tree only — the bundler itself never reaches the hash. Swapping `node:18` for `node:20`, or flipping `minify`, produces the same hash, the same staged path, and a byte-identical config. Reproduction in the review body.

`extraHash` is the documented defence, but it is manual and silent when forgotten, and both early planned consumers (local and Docker bundling) hit this on first real use.

There is a precedent in the codebase that fits better: `IIgnoreStrategy.cacheKey`. The same shape on `IAssetBundler` would let `AssetStaging` fold build identity in automatically, the way it already folds `extraHash`:

```ts
export interface IAssetBundler {
  bundle(options: BundleOptions): string;
  /** Folded into the asset hash so a changed build changes identity. */
  readonly bundlerKey?: string;
}
```

A Docker bundler returns `docker:<image>:<command>`, an esbuild bundler its options digest. Compiled this against jsii 5.9 — it projects as optional, and an implementation may omit it entirely:

```
IAssetBundler      interface  [('bundlerKey', 'optional')]
FakeDockerBundler  class      [('bundlerKey', 'optional')]
KeylessBundler     class      []
```

Cheap now; breaking for implementers once the interface ships.

packages/cdktn/src/asset-staging.ts
```diff
@@ -214,13 +252,41 @@ export class AssetStaging extends Construct implements IAsset {
...
+      const produced = this.bundler.bundle({
+        source: this.sourcePath,
+        outputDir: scratch,
+      });
+      // The bundler's output is packaged exactly as an unbundled source would
+      // be. The ignore strategy still applies to the built output, so a
+      // bundler that copies over ignored files does not smuggle them into the
+      // artifact.
+      this.packaging.pack({
+        source: produced,
```
jsteinich: Applying the ignore strategy to the bundler's output is defended as stopping a bundler from smuggling ignored files in, but the more common pairing is the opposite:

```ts
new TerraformAsset(s, "fn", {
  path: "./src",
  exclude: ["node_modules"],   // don't hash my local install
  bundler: installBundler,     // ...the bundler produces its own
});
```

Worth splitting by bundler style: esbuild-style bundlers inline their dependencies, so there is no `node_modules` in the output and this is harmless. Install-style bundlers — `npm install --production`, `pip install -t` — write exactly the directory the user excluded from source, and `pack()` then strips it, shipping an artifact without its dependencies. A Python function construct would hit this.

`exclude` reads as source filtering; applying it to output deletes legitimately-built content.

packages/cdktn/src/assets.ts
```diff
@@ -225,6 +225,65 @@ export class AssetPackaging {
...
+   *
+   * The returned directory is then packaged (copied or archived) into the
+   * staged location by the caller's {@link IAssetPackaging}, exactly as an
+   * unbundled source directory would be. Implementations should write into
+   * `options.outputDir` and return it (or a subdirectory of it), never write
+   * back into `options.source`.
+   *
+   * @param options - see {@link BundleOptions}
+   */
+  bundle(options: BundleOptions): string;
```
jsteinich, line 301: The contract here is that `bundle` returns a directory, but `FilePackaging.pack` is `fs.copyFileSync(options.source, options.target)`. With a bundler, `options.source` is that directory:

```
copyFileSync(<dir>, target)  ->  EPERM on Windows, EISDIR on POSIX
```

So `TerraformAsset` with `type: AssetType.FILE` plus any bundler dies at synth with an opaque filesystem error. DIRECTORY and ARCHIVE are fine — both read the directory.

"Bundle my TS into one `index.js` and stage it as a file" is a reasonable thing to want. Either validate at construction and reject the combination with a clear error, or let `bundle` return a file path and have packaging handle both.

Related: `fileName` for a no-extension packaging is `path.basename(this.sourcePath)`, so a bundled asset is still named after its source. Harmless for ARCHIVE, but source naming leaking into bundled output.

jsteinich (commented):
The interface shape matches what #380 proposed — `bundle(options): string` returning the output directory, build deferred into the synth window, bundlers outside core. Worth noting one divergence: #380 said *"`OUTPUT` hashing stays supported, documented as forcing an eager build"*; this rejects it outright when a bundler is set. Same direction, but a stricter resolution that deserves explicit sign-off rather than reading as already-settled.

Three notes below. The first I verified end to end rather than reading off the code — two bundlers writing different output, consumed by a real resource referencing `asset.path`:

```
hashA / hashB        9FD43BDFF8DE0DE43D08F9C369E89522   identical
pathA / pathB        assets/asset/9FD43BDF.../archive.zip  identical
emitted cdk.tf.json  byte-identical
staged artifact      bytes differ
```

`_onSynthesize` removes the previous version folder before writing, so the old artifact is actively deleted and the new one written at the same path — it is not a stale file lingering, it is the artifact being replaced underneath a plan that shows nothing.

Test gaps while in here: nothing covers FILE packaging with a bundler (see below), nothing proves the scratch dir is actually cleaned, nothing covers a bundler that throws, and nothing exercises the documented "or a subdirectory of it" return.

packages/cdktn/src/asset-staging.ts
eduardomourar: Added it to IAssetBundler, folded into the hash alongside extraHash/salt. Went with your IIgnoreStrategy.cacheKey shape exactly: optional, and a bundler can omit it. So a Docker bundler returns docker:<image>:<command>, esbuild an options digest, and swapping node:18→node:20 changes identity automatically instead of relying on someone remembering extraHash.

packages/cdktn/src/asset-staging.ts
eduardomourar: You're right, applying it to the output was backwards. `exclude` now filters the source only. The install-style case (exclude: ["node_modules"] + a bundler that produces its own) keeps the built node_modules now; test added proving it survives.

packages/cdktn/src/assets.ts
eduardomourar: Rejecting at construction with a clear error rather than letting copyFileSync blow up on a directory. Kept the interface contract (bundle returns a directory) instead of overloading the return type. That also closes the fileName source-name leak you flagged, since only DIRECTORY/ARCHIVE are reachable with a bundler and neither uses the source name.

eduardomourar:
@jsteinich, you were right to want sign-off here, and after thinking about it I went the other way from my first pass: rather than reject OUTPUT, I built the eager-build machinery so it actually works. OUTPUT now builds once in the constructor, hashes the real output, and stage() reuses that build. So it matches #380 (OUTPUT supported, eager build) and lands the internals the local/Docker bundler PRs need — they just implement IAssetBundler and pick a hash type. SOURCE stays the deferred, skippable default; OUTPUT trades skippability for hashing real bytes, which is the point of it.

Test gaps closed too: FILE+bundler, scratch cleanup, a throwing bundler, subdirectory return, and the OUTPUT build-once/reuse path.

packages/cdktn/src/asset-staging.ts
```diff
@@ -211,16 +271,130 @@ export class AssetStaging extends Construct implements IAsset {
...
+   * @param canonical - use the canonical hashing scheme, see `hashPath`
+   * @param archive - hash under an archive frame (no directory entries)
+   */
+  private hashOutput(canonical: boolean, archive: boolean): string {
+    const scratch = fs.mkdtempSync(path.join(os.tmpdir(), "cdktn-bundle-"));
+    const produced = this.bundler!.bundle({
+      source: this.sourcePath,
+      outputDir: scratch,
+    });
+    this.eagerBuild = { scratch, produced };
```
jsteinich, line 311: The scratch directory created here is only ever cleaned up inside `stage()` — both `rmSync` calls live there, and there is no exit hook. So when `stage()` is never called, it survives the process:

```ts
// OUTPUT + bundler, in a stack that is not being synthesized
new AssetStaging(...)   // builds eagerly into os.tmpdir()/cdktn-bundle-XXXX
// stage() never runs -> scratch stranded
```

That is exactly the case this path already documents as wasteful ("the stack is bundled whether or not it is being synthesized"), so the leak compounds it rather than being an unrelated edge. A `cdktn diff` across a multi-stack app strands one scratch per OUTPUT-bundled asset in the untouched stacks, and bundler output is not small — `node_modules`, build caches. `%TEMP%` is not swept automatically on Windows.

`"cleans up the eagerly-built scratch directory after stage()"` covers the happy path; the never-staged path is the one that leaks.

A process-level registry with an `exit` hook doing sync cleanup would cover it, or the constructor could hash and discard and let `stage()` rebuild — at the cost of the "built once" property this path exists to get.

packages/cdktn/src/asset-staging.ts
```diff
@@ -129,13 +179,28 @@ export class AssetStaging extends Construct implements IAsset {
...
     this.ignoreStrategy =
       props.ignoreStrategy ?? new ExcludeIgnoreStrategy(props.exclude ?? []);
 
+    // A bundler always produces a directory of output. Single-file packaging
+    // copies its source with `copyFileSync`, which throws (EISDIR/EPERM) when
+    // handed a directory - so the combination is rejected here rather than
+    // failing opaquely at synth. Packaging that reads a directory (a tree
+    // copy, or an archive) is fine; the incompatible case is the one that
+    // neither produces nor archives a directory.
+    if (
```
jsteinich: The comment states the real rule — "the incompatible case is the one that neither produces nor archives a directory" — but `omitsDirectoryEntries` does not mean "archives a directory". It means the emitted artifact carries no directory records, which is a hash-framing property.

A custom `tar` / `tar.gz` packaging preserves directory entries (`omitsDirectoryEntries: false`) and emits a single file (`producesDirectory: false`), so this rejects it — even though tar handles a directory source perfectly well.

It is the same conflation as the `=== AssetPackaging.ZIP` identity check that got fixed in #433: a flag that answers a different question standing in for the one actually being asked. The property needed here is orthogonal to both existing ones — *does `pack()` accept a directory as its source?* — so it wants its own member rather than a proxy.

Cheap now; `IAssetPackaging` gets more expensive to extend with every consumer added.

jsteinich (commented):
All three addressed, and implementing `OUTPUT` rather than rejecting it puts this back in line with what #380 landed on. The `bundlerKey` docs and the `packBundlerOutput` rationale both capture the reasoning well, and the new tests cover the gaps — including the throwing-bundler cleanup and the subdirectory return.

Two things the eager-build path introduces.

packages/cdktn/src/asset-staging.ts
eduardomourar: Fixed. Went with the process-exit registry so the "built once" property stays intact.

`hashOutput` now registers each eager scratch dir in a module-level set, and a lazily-installed `process.once("exit")` handler `rmSync`s any that survive. `stage()` still cleans up and unregisters on the happy path, so the exit hook only ever sees the strays — the never-staged case you described (OUTPUT + bundler in an unsynthesized stack during `cdktn diff`).

Verified with a subprocess test that constructs an OUTPUT-bundled asset, never calls `stage()`, and asserts the scratch dir is gone after the process exits.

packages/cdktn/src/asset-staging.ts
eduardomourar: Fixed. Added a dedicated `IAssetPackaging.acceptsDirectorySource` and keyed the guard on it, rather than proxying through `omitsDirectoryEntries`.

Same resolution as the `=== AssetPackaging.ZIP` fix in #433: a flag answering the actual question — does `pack()` take a directory source? — instead of one that answers a different one. `FILE` is `false`; `DIRECTORY`/`ZIP` are `true`. A `tar.gz`-style packaging (single-file output, directory entries preserved) is now correctly accepted, covered by a new test.

jsteinich (commented):
Both addressed well. The exit sweep drops entries in `cleanupScratch` as well as removing the directory, so the set does not grow across repeated synths, and only the eager build registers — the SOURCE-path scratch stays inside its own `try/finally`. `acceptsDirectorySource` is the right property and retires the proxy. Moving the type validation ahead of `AssetStaging` is a good catch on top; the collapsed `stat.isFile() !== (this.type === AssetType.FILE)` is equivalent across all four cases.

One thing left: `IAssetPackaging` gains `acceptsDirectorySource` as a **required** member. #433 merged on the 16th, so the interface is already published with `omitsDirectoryEntries` — this is a second breaking change to a live behavioural interface, and neither commit here is marked.

No version cost, same as last time: the manifest is at 0.24.0 and #371 and #433 are both already in the cycle as `feat(lib)!:`, so the pending release is 0.25.0 either way.

Unrelated to this PR, for the record — the failing unit job is the new OpenTofu matrix cell from #439, which landed on main at 20:43 today and arrived here via the rebase:

```
tofu1.12.6 init exited with code 1
│ Error: Failed to install provider
│ Error while installing kreuzwerker/docker v2.16.0: the provider is not ...
-  "registry.terraform.io/hashicorp/null"
+  "registry.opentofu.org/hashicorp/null"
```

The failures are in `@cdktn/provider-schema` and `@cdktn/provider-generator`; `packages/cdktn` passed 621/621 in the same job. #433 and #434 merged before that job existed and main does not run the workflow, so this PR looks like the first to exercise it — worth chasing on #439 rather than here.

jsteinich (approved):

jsteinich (commented):
Nine real bundlers (TerraConstructs local/esbuild/bind-mount/volume-copy/chain, buildkit, Rolldown) were ported onto this interface and run against real toolchains in [sakul-learning/cdktn-bundler-harness](https://github.com/sakul-learning/cdktn-bundler-harness) — 108/108 scenarios pass, so this is not a "does not work" list. The [DX-REPORT](https://github.com/sakul-learning/cdktn-bundler-harness/blob/main/DX-REPORT.md) covers where an implementation still needed a workaround. It targets head `98b39e49`, which is content-identical to the current head.

I checked the following against the code here rather than taking them from the report.

**Contract mismatches — docs currently state something the code does not do.** Cheapest to close, and the most likely to burn an implementer:

1. `BundleOptions.source` is documented absolute, but `terraform-asset.ts:134` stores `path.relative(process.cwd(), absolutePath)` for a relative `path`, and `AssetStaging` passes it through unresolved. That is the common case, and a bundler spawning a tool with its own cwd (docker `-w`, esbuild) resolves elsewhere. Either resolve in `AssetStaging` or drop the guarantee.
2. `exclude` is documented as filtering what the bundler reads; the bundler is handed the raw source. It does reach identity via `hashSource`, so the two disagree in both directions — an excluded file that changes moves the artifact and not the hash, and copy-everything bundlers ship bytes the identity does not account for. Either materialise the filtered input or delete the claim.

**Correctness:**

3. A second `stage()` rebuilds: `eagerBuild` is consumed, then the call falls through to the deferred path. Under `OUTPUT` the hash came from the first build, so a non-deterministic bundler stages bytes that do not match its own hash. Asserting single-use or retaining the artifact both close it.
4. The exit sweep only registers `process.once("exit")`, which Node does not run on SIGINT/SIGTERM — so Ctrl-C during a synth strands the scratch, which is exactly when the largest trees are left behind.

**Smoothing:**

5. `bundle()`'s return is not validated, so a file-returning bundler silently bypasses the directory contract.
6. The FILE-packaging guard throws with `AssetStaging`'s id, so the message reads "TerraformAsset **Staging**" — the internal child rather than the user's construct.
7. `bundlerKey` is the only config channel and it is optional and unvalidated; forgetting it silently reuses stale artifacts. Worth stating in the docs that it must serialise everything that can move the output, and shipping a helper so the convention has one implementation rather than one per bundler.

**Probably follow-ups rather than this slice** — both are design additions and both have documented workarounds: an archive-shaped artifact (`AssetType.FILE` + bundler is a hard reject, so archive-producing bundlers wrap in a directory and get a zip containing `archive.zip`), and a decline protocol so a third-party local bundler and a third-party Docker bundler can be composed instead of reimplemented together.

